PR #9, opened 2017-11-25
[bugfix] - `filter` method fails if a host doesn't have one of the filtering keys

ogenstad:
The current filter function requires that the filters being used are defined on all of the hosts in the inventory.

### DIFF
--- a/brigade/core/inventory.py
+++ b/brigade/core/inventory.py
@@ -1,3 +1,4 @@
+import uuid
 from multiprocessing import Manager
 
 from brigade.core import helpers
@@ -106,6 +107,17 @@ class Host(object):
     def __repr__(self):
         return "{}: {}".format(self.__class__.__name__, self.name)
 
+    def _getitem_or_no_match(self, item, no_match):
+        try:
+            return self.data[item]
+        except KeyError:
+            try:
+                if self.group:
+                    return self.group[item]
+            except KeyError:
+                pass
+            return no_match
+
     def items(self):
         """
         Returns all the data accessible from a device, including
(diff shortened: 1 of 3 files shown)


ogenstad:
Perhaps `.getitem()` or `.getdata()` would be better than just `.get()`

brigade/core/inventory.py
```diff
@@ -106,6 +107,25 @@ def __str__(self):
...
+    def get(self, item, default=None):
+        """
+        Returns the value ``item`` from the host or hosts group variables.
+
+        Arguments:
+            item(``str``): The variable to get
+            default(``any``): Return value if item not found
+
+        """
+        try:
```
dbarrosop: I think this would be cleaner:

```
try:
    return self.__item__(item)
except KeyError:
    return default
```

Basically you are duplicating the code in `self.__item__` so calling it should be enough

brigade/core/inventory.py
```diff
@@ -189,6 +209,7 @@ def filter(self, filter_func=None, **kwargs):
             filtered = {n: h for n, h in self.hosts.items()
                         if filter_func(h, **kwargs)}
         else:
+            no_match = uuid.uuid4()
             filtered = {n: h for n, h in self.hosts.items()
-                        if all(h[k] == v for k, v in kwargs.items())}
+                        if all(h.get(k, no_match) == v for k, v in kwargs.items())}
```
dbarrosop: Shouldn't we assume `None`? So, basically, in the inventory:

```
---
my_host1:
    my_var: null
my_host2: {}
```

Both would pass a filter where I say `inv.filter(my_var=None)`

ogenstad: I was going back and forth on that one. I can't really think of a good scenario for this. Just wanted to avoid to match more hosts than the user actually indented. Though it could go either way.

Do you want me to remove it?

ogenstad:
Fixed

PR #11, opened 2017-11-25
Added  `netmiko_ssh` task to run any netmiko method

ktbyers:
Note, I didn't include netmiko as a requirement to the brigade package.


ogenstad:
I think the _ssh in `netmiko_ssh` feels a bit redundant. Could it perhaps be named netmiko_commands, netmiko_run, netmiko_exec or something?

ktbyers:
@ogenstad Sounds reasonable...I will rename to `netmiko_run` in a future PR.

dbarrosop:
I can do it myself as I write tests for it

PR #12, opened 2017-11-25
Add `num_workers` argument to `run()`

ogenstad:
Some tasks might have to run in serial, this change allowes a user to
change the number of workers for a specific task without changing the
value of Brigade.num_workers

### DIFF
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -126,8 +126,7 @@ class Brigade(object):
         Returns:
             :obj:`brigade.core.task.AggregatedResult`: results of each execution
         """
-        if not num_workers:
-            num_workers = self.num_workers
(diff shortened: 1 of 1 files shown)


brigade/core/__init__.py
```diff
@@ -125,10 +126,13 @@ def run(self, task, **kwargs):
         Returns:
             :obj:`brigade.core.task.AggregatedResult`: results of each execution
         """
-        if self.num_workers == 1:
+        if not num_workers:
+            num_workers = self.num_workers
```
dbarrosop: `num_workers = num_workers or self.num_workers` : )

ogenstad:
Sorry, bad habits.. Fixed now.

PR #13, opened 2017-11-26
add host properties to gather them consistently across tasks

dbarrosop:
WIP, still want to add way more tests, maybe discuss a few things and update documentation.

Basically trying to add properties to the `Host` class so gathering basic info is easier and consistent across tasks.

Also, because I am doing changes that might break stuff I am adding tests for all the tasks.


brigade/plugins/tasks/networking/napalm_configure.py
```diff
@@ -25,13 +25,15 @@ def napalm_configure(task, configuration, replace=False, hostname=None, username
...
+        "hostname": hostname or task.host,
+        "username": username or task.host.username,
+        "password": password or task.host.password,
         "timeout": timeout,
-        "optional_args": optional_args or {"port": task.host["napalm_port"]},
+        "optional_args": optional_args or {},
     }
-    network_driver = get_network_driver(driver or task.host["nos"])
+    if "port" not in parameters["optional_args"] and task.host.network_api_port:
+        parameters["optional_args"] = task.host.network_api_port
```
ogenstad: This should be `parameters["optional_args"]["port"] = task.host.network_api_port` for all napalm tasks. However it might also be confusing for some people when there's also an `ssh_port` property on the host. `network_api_port` is more correct in this case though.

brigade/plugins/tasks/networking/netmiko_run.py
```diff
@@ -32,21 +32,13 @@ def netmiko_ssh(task, method, ip=None, host=None, username=None, password=None,
           * result (``dict``): dictionary with the result of the getter
     """
     parameters = {
-        "username": username or task.host["brigade_username"],
-        "password": password or task.host["brigade_password"],
+        "ip": ip or host or task.host.host,
+        "username": username or task.host.username,
+        "password": password or task.host.password,
+        "port": task.host.network_api_port,
```
ogenstad: Here it would make more sense to use `"port": task.host. ssh_port`

tests/inventory_data/groups.yaml
```diff
@@ -0,0 +1,16 @@
+---
+all:
+    my_var: comes_from_all
+    brigade_host: 127.0.0.1
+    brigade_username: root
+    brigade_password: docker
+    brgiade_os: linux
```
ogenstad: Typo: brgiade_os

PR #15, opened 2017-11-27
Add tcp_ping and more tests

ogenstad:
Perhaps the new Exception should be called BrigadeFileError?

At the moment it only triggers on file not found errors and yaml validation errors. Should also look at permission issues and provide more help about what goes wrong with the yaml validation if an incorrect file is submitted.


brigade/core/helpers/jinja_helper.py
```diff
@@ -1,17 +1,29 @@
...
     env = Environment(loader=FileSystemLoader(path),
                       undefined=StrictUndefined,
                       trim_blocks=True)
-    #  env.filters.update(jinja_filters.filters())
-    template = env.get_template(template)
+
+    try:
+        template = env.get_template(template)
+    except TemplateSyntaxError as e:
+        raise TemplateError(e.message, e.lineno,
```
dbarrosop: I don't think we should hide the original exception. The `BrigadeExecutionError` is useful because, if you use the subprocess library to execute commands, commands that fail don't return an exception of their own. However, if something can throw a meaningful exception of its own I think that's better than trying to define our own exceptions instead. Thoughts?

ktbyers: I would rather have the original exception.

ogenstad: I think the above example with the Jinja exception is a quite bad example as I don't really add anything. If we instead look at the load_json task it basically works like this:

```python
>>> import json
>>> with open('broken.json', 'r') as f:
...     data = json.loads(f.read())
...
  File "<stdin>", line 2, in <module>
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/2.7.14/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Invalid control character at: line 6 column 21 (char 90)
>>>
```

Catching the original exception would mean that a ValueError is raised if the file broken.json wasn't a proper json file. (or `json.decoder.JSONDecodeError: Invalid control character at: line 6 column 21 (char 90)` in Python 3). In this example I think it makes more sense to create another exception which also highlights the file which actually contained the error as that would save time for people to track down the error. I would imagine that this also could generate less support questions. Could also make it easier for people to be able to see what kind of errors to normally expect.

However, mostly I just wanted to create a few test cases to move this forward. Can revert to just relying on the original exceptions. If we want to have custom exceptions later it's easy to change once the tests are in place.

So just let me know which parts to remove, or if you have any other comments on the PR.

dbarrosop: So the way brigade works is that it should log the error. So an alternative mechanism to solve this would be to have `logger.debug("opening file {}".format(filename))` right before parsing the file.

ogenstad: Not sure that I like the idea of having to enable debug logging in order to figure out what's going on if there's an issue with some of the hosts. 

For now I'll just remove the custom exceptions and we can revisit this in the future if needed.

Did you have any other comments on this PR?

dbarrosop: Well, we could get into a debate here of traceability vs error handling :)

ogenstad: We'll just call it repo-owners-democracy and move on. ;)

PR #18, opened 2017-12-02
Add Netmiko send_command with ntc-templates textfsm support

ktbyers:
I need to review this: https://github.com/napalm-automation/brigade/tree/dbarrosop/host_properties


ktbyers:
Also need to get agreement on one task per Python module or not.

dbarrosop:
@ktbyers could you look at this PR? https://github.com/napalm-automation/brigade/pull/25/files#diff-3809cfaf6a7bf2bf113debc68df5541cR174

Would be nice to make it in a way where netmiko tasks can reuse the same connection.

ktbyers:
@dbarrosop Yes, let me review...I was thinking the same i.e. that it would be good to allow a way for SSH connection reuse.

ktbyers:
Will resubmit separately.

PR #24, opened 2017-12-05
Change structure of documentation

ogenstad:
@dbarrosop, take a look at this and see what you think. With this setup I don't think any additional changes to the structure will be needed.

At least for a while. :)

I will start to write some parts of the intro tutorial as well.


dbarrosop:
I agree, looks really good. Merging :)

PR #25, opened 2017-12-05
major improvements to ssh related tasks

dbarrosop:
* Removed `scp` task in favor of `sftp` task.
* Migrated from ssh/scp local commands to paramiko
* Added a reusable `ssh_connection` property that can be used across tasks.


brigade/core/inventory.py
```diff
@@ -168,6 +171,46 @@ def nos(self):
...
+        if ssh_config_file is None:
+            ssh_config_file = os.path.join(os.path.expanduser("~"), ".ssh", "config")
+
+        client = paramiko.SSHClient()
+        client._policy = paramiko.WarningPolicy()
+        client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+
+        ssh_config = paramiko.SSHConfig()
+        if os.path.exists(ssh_config_file):
+            with open(ssh_config_file) as f:
```
ogenstad: Perhaps this should be handled in the same way as self._ssh_connection. Just to avoid reading the same file multiple times for multiple hosts, or have it auto loaded with the config.

ogenstad:
Does the connections need to be closed at the end, or will it just send a tcp reset/fin?

Also should there be a way to close connections that are no longer needed?

dbarrosop:
No, it's actually not needed as paramiko has a `__del__` magic method closing the connection. I did a tcpdump to test I can confirm it:

```
11:48:27.906391 IP localhost.54869 > localhost.65001: Flags [F.], seq 3417, ack 3823, win 12639, options [nop,nop,TS val 2122312404 ecr 2122312403], length 0
11:48:27.906417 IP localhost.54869 > localhost.65001: Flags [F.], seq 3417, ack 3823, win 12639, options [nop,nop,TS val 2122312404 ecr 2122312403], length 0
11:48:27.906464 IP localhost.65001 > localhost.54869: Flags [.], ack 3418, win 12652, options [nop,nop,TS val 2122312404 ecr 2122312404], length 0
11:48:27.906472 IP localhost.65001 > localhost.54869: Flags [.], ack 3418, win 12652, options [nop,nop,TS val 2122312404 ecr 2122312404], length 0
```

brigade/core/inventory.py
dbarrosop: true

dbarrosop: actually, it is not that easy to solve here right now but it should be fixable once we have the `config` object you are working on

dbarrosop:
There are still minor improvements to do that require ogenstad's work on the `Config` class and we still have to discuss how to make sure this can be leveraged by `netmiko` and maybe `napalm` but as this is already better than what we had I am going to merge it and release a new version.

brigade/core/inventory.py
ogenstad: I was just thinking about some function which would only load the file if it hadn't been previously loaded. But yes the config object would be a good start for this. Should need quite a few hosts before this actually is a problem, it's more that it's unnecessary.

dbarrosop: yeah, I think we should pass around the `Config` to all the hosts and solve it there so no need to bother for now. It's a minor optimization anyway.

PR #26, opened 2017-12-06
For discussion - Add config class

ogenstad:
Something like this could work regarding `ssh_connection` in #25.

Could also ship a SimpleYaml config plugin similar to the SimpleInventory one.

Currently one would access the variables with `task.brigade.config.ssh_config_file` which seems a bit long, so perhaps some shortcut is needed.

### DIFF
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -43,25 +43,20 @@ class Brigade(object):
     Arguments:
         inventory (:obj:`brigade.core.inventory.Inventory`): Inventory to work with
         dry_run(``bool``): Whether if we are testing the changes or not
-        num_workers(``int``): How many hosts run in parallel
-        raise_on_error (``bool``): If set to ``True``, :meth:`run` method of will
-          raise an exception if at least a host failed.
+        config (:obj:`brigade.core.configuration.Config`): Configuration object
+        config_file (``str``): Path to Yaml configuration file
 
     Attributes:
         inventory (:obj:`brigade.core.inventory.Inventory`): Inventory to work with
         dry_run(``bool``): Whether if we are testing the changes or not
-        num_workers(``int``): How many hosts run in parallel
-        raise_on_error (``bool``): If set to ``True``, :meth:`run` method of will
-          raise an exception if at least a host failed.
+        config (:obj:`brigade.core.configuration.Config`): Configuration parameters
     """
 
-    def __init__(self, inventory, dry_run, num_workers=20, raise_on_error=True,
+    def __init__(self, inventory, dry_run,
                  config=None, config_file=None):
         self.inventory = inventory
 
         self.dry_run = dry_run
-        self.num_workers = num_workers
-        self.raise_on_error = raise_on_error
         if config_file:
             self.config = Config(config_file=config_file)
         else:
@@ -129,19 +124,19 @@ class Brigade(object):
 
         Raises:
             :obj:`brigade.core.exceptions.BrigadeExceptionError`: if at least a task fails
-              and self.raise_on_error is set to ``True``
+              and self.config.raise_on_error is set to ``True``
 
         Returns:
             :obj:`brigade.core.task.AggregatedResult`: results of each execution
         """
-        num_workers = num_workers or self.num_workers
+        num_workers = num_workers or self.config.num_workers
 
         if num_workers == 1:
             result = self._run_serial(task, **kwargs)
         else:
             result = self._run_parallel(task, num_workers, **kwargs)
 
-        if self.raise_on_error:
+        if self.config.raise_on_error:
             result.raise_on_error()
         return result
 
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -4,6 +4,7 @@ import traceback
 from multiprocessing.dummy import Pool
 
 from brigade.core.task import AggregatedResult, Task
+from brigade.core.configuration import Config
 
 
 if sys.version_info.major == 2:
@@ -54,12 +55,14 @@ class Brigade(object):
           raise an exception if at least a host failed.
     """
 
-    def __init__(self, inventory, dry_run, num_workers=20, raise_on_error=True):
+    def __init__(self, inventory, dry_run, num_workers=20, raise_on_error=True,
+                 config=None):
         self.inventory = inventory
 
         self.dry_run = dry_run
         self.num_workers = num_workers
         self.raise_on_error = raise_on_error
+        self.config = config or Config()
 
         format = "\033[31m%(asctime)s - %(name)s - %(levelname)s"
         format += " - %(funcName)20s() - %(message)s\033[0m"
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -3,8 +3,8 @@ import sys
 import traceback
 from multiprocessing.dummy import Pool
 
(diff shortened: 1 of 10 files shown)


brigade/core/__init__.py
```diff
@@ -54,12 +55,14 @@ class Brigade(object):
           raise an exception if at least a host failed.
     """
 
-    def __init__(self, inventory, dry_run, num_workers=20, raise_on_error=True):
+    def __init__(self, inventory, dry_run, num_workers=20, raise_on_error=True,
+                 config=None):
         self.inventory = inventory
 
         self.dry_run = dry_run
         self.num_workers = num_workers
```
dbarrosop: I guess `num_workers` and `raise_on_error` should go in the `Config` as well

brigade/core/configuration.py
```diff
@@ -0,0 +1,16 @@
...
+}
+
+
+class Config:
+    """
+    Documentation
+
+    """
+
+    def __init__(self):
```
dbarrosop: I'd add also the option to pass a configuration file and load the config parameters from there.

dbarrosop: By the way, I guess order should be:

1. env variable
2. configuration file
3. default

dbarrosop:
> Could also ship a SimpleYaml config plugin similar to the SimpleInventory one.

I think it should be provided by default tbh by the `config` class itself, it's simple enough that using "plugins" for that seems a bit too much. You can always provide your own configuration class if you have special needs.

ogenstad:
Makes sense to me.

Regarding order I think I prefer environment variables before config file.

If we add yaml support to the default config class we need to send in the file as a parameter. Based on that I would change it so that the Brigade __init__ looks like this instead:
```python
    def __init__(self, inventory, dry_run, config=None):
        self.config = Config(config)
```

In that case I don't think there's an immediate need to use a custom config class. If needed it could perhaps be changed to something like:
```python
    def __init__(self, inventory, dry_run, config=None):
        if isinstance(config, str):
            self.config = Config(config)
        else:
            self.config = config or Config()
```

The alternative would be to have both `config` and `config_file` as arguments. Or to always load brigade.yaml from the current directory or something like /etc/brigade/config.yaml, which I don't really like. Thoughts?

Thinking of storing defaults in a dictionary or a yaml file we include with Brigade. Have to figure out a good way to document the defaults. Probably just have Sphinx read that dict or yaml file and create some table.

dbarrosop:
> I don't think there's an immediate need to use a custom config class

I agree, I am happy providing `load_json` and `load_yaml` and `load_toml` and `load_whatever` functions there. As long as the data structure in those file is consistent it should be pretty straightforward.

> The alternative would be to have both config and config_file as arguments

I personally prefer this one.

ogenstad:
I was thinking of something like this. After `_assign_properties()` is run we can trigger something to set all of the ssh parameters.

Considering of the CONF variable should be moved to an external .py or perhaps .yaml file.

brigade/core/configuration.py
```diff
@@ -0,0 +1,61 @@
+import ast
+import os
+
+
+import yaml
+
+
+CONF = {
```
dbarrosop: I'd suggest using `marshmallow` or somehting like that:

https://marshmallow.readthedocs.io/en/latest/

dbarrosop: maybe it's too heavy...

brigade/core/configuration.py
```diff
@@ -0,0 +1,61 @@
...
+    'raise_on_error': {
+        'description': "If set to ``True``, :meth:`run` method of will raise an exception if at "
+                       "least a host failed.",
+        'env': 'BRIGADE_RAISE_ON_ERROR',
+        'type': bool,
+        'default': True,
+    },
+    'ssh_config_file': {
+        'description': 'User ssh_config_file',
+        'env': 'BRIGADE_SSH_CONFIG_FILE',
```
dbarrosop: should we assume always that is `"BRIGADE_{}".format(key.upper())`?

ogenstad: Not really sure, perhaps not for things like http proxy settings and such?

brigade/core/configuration.py
ogenstad: I think it would be a shame to add a requirement just for this. We don't need any advanced features. It will only be a matter of inputting data in that section and have some tests to verify.

ogenstad: Also if desired that part is really easy to change in the future. Might depend on how many parameters that will need to be defined.

brigade/core/configuration.py
dbarrosop: maybe we can default it to be as I described but have the possibility to override with this method

ogenstad: Yes that makes sense, also makes the dict easier to read. Pushed a change with this.

ogenstad:
I think this is ready for review now. Not really happy about the documentation of the parameters of the Config class, it will work from the website but not as much from the built in help of Python, perhaps we can just provide a few examples there.

The next step would be to have the Config load things such as ssh_config files, don't know if that should be done by default though or if it should be done once the first ssh connection is done.

Also we can thing of what other parameters should be added like `ssh_system_config_file`.

ogenstad:
One thing I realised with this is that the Host object wouldn't currently have access to Brigade.config from the current `ssh_connection()`. That would also have to be addressed in some way.

dbarrosop:
Yeah, I think we will have to add a `brigade` reference to the host so the host can access it. Right now tasks can but it's probably a good idea to do the same in the host as well.

brigade/core/configuration.py
```diff
@@ -0,0 +1,69 @@
...
+            if CONF[p]['type'] == 'bool':
+                if os.environ.get(env) is not None:
+                    v = os.environ.get(env)
+                elif c.get(p) is not None:
+                    v = c.get(p)
+                else:
+                    v = CONF[p]['default']
+                v = ast.literal_eval(str(v).title())
+            else:
+                v = os.environ.get(env) or c.get(p) or CONF[p]['default']
```
dbarrosop: as discussed, this problably be need to be rewritten as:

```
v = os.environ.get(env) or c.get(p)
v = v if v is not None else CONF[p]['default']
```

dbarrosop: Actually, the whole block can be rewritten as:

```
env = CONF[p].get('env') or 'BRIGADE_' + p.upper()
v = os.environ.get(env) or c.get(p)
v = v if v is not None else CONF[p]['default']
if CONF[p]['type'] == 'bool':
    v = ast.literal_eval(str(v).title())
else:
    v = types[CONF[p]['type']](v)
setattr(self, p, v)
```

ogenstad:
Fixed the assignments and added a BASEPATH to the Sphinx conf since the tests failed when it couldn't find the config jinja template.

brigade/core/configuration.py
```diff
@@ -0,0 +1,64 @@
...
+        'type': 'int',
+        'default': 20,
+    },
+    'raise_on_error': {
+        'description': "If set to ``True``, (:obj:`brigade.core.Brigade.run`) method of will raise "
+                       "an exception if at least a host failed.",
+        'type': 'bool',
+        'default': True,
+    },
+    'ssh_config_file': {
```
dbarrosop: can you change paramiko connection to use this?

PR #35, opened 2017-12-08
Decouple SSH connection getter and setter operation

ktbyers:
This will allow a Netmiko SSH connection be assigned to the host object.

This is also a more clear pattern (instead of overloading the `getter` with a `setter` pattern). In other words, I think it much clearer to call some method to create the SSH connection then pass that to `test_host.ssh_connection` through assignment.

We also might want to rename the attribute to attribute to be more generic like `connection` so it could be used in non-ssh contexts like NETCONF, APIs, serial connections.

### DIFF
--- a/brigade/core/inventory.py
+++ b/brigade/core/inventory.py
@@ -177,42 +177,51 @@ class Host(object):
         if hasattr(self, "_ssh_connection"):
             return self._ssh_connection
 
-        # TODO configurable
-        ssh_config_file = os.path.join(os.path.expanduser("~"), ".ssh", "config")
+    @ssh_connection.setter
+    def ssh_connection(self, connection=None):
+        """Set the ssh connection. Defaults to creating a new Paramiko SSH connection.
 
-        client = paramiko.SSHClient()
-        client._policy = paramiko.WarningPolicy()
-        client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        Arguments:
+            connection(``Connection Object``): Connection object to be used in future tasks.
+        """
+        if connection is not None:
+            self._ssh_connection = connection
+        else:
+            # TODO configurable
+            ssh_config_file = os.path.join(os.path.expanduser("~"), ".ssh", "config")
 
-        ssh_config = paramiko.SSHConfig()
-        if os.path.exists(ssh_config_file):
-            with open(ssh_config_file) as f:
-                ssh_config.parse(f)
+            client = paramiko.SSHClient()
+            client._policy = paramiko.WarningPolicy()
+            client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
 
-        parameters = {
-            "hostname": self.host,
-            "username": self.username,
-            "password": self.password,
-            "port": self.ssh_port,
-        }
+            ssh_config = paramiko.SSHConfig()
+            if os.path.exists(ssh_config_file):
+                with open(ssh_config_file) as f:
+                    ssh_config.parse(f)
 
-        user_config = ssh_config.lookup(self.host)
-        for k in ('hostname', 'username', 'port'):
-            if k in user_config:
-                parameters[k] = user_config[k]
+            parameters = {
+                "hostname": self.host,
+                "username": self.username,
+                "password": self.password,
+                "port": self.ssh_port,
+            }
 
-        if 'proxycommand' in user_config:
-            parameters['sock'] = paramiko.ProxyCommand(user_config['proxycommand'])
+            user_config = ssh_config.lookup(self.host)
+            for k in ('hostname', 'username', 'port'):
+                if k in user_config:
+                    parameters[k] = user_config[k]
 
-        # TODO configurable
-        #  if ssh_key_file:
-        #      parameters['key_filename'] = ssh_key_file
-        if 'identityfile' in user_config:
-            parameters['key_filename'] = user_config['identityfile']
+            if 'proxycommand' in user_config:
+                parameters['sock'] = paramiko.ProxyCommand(user_config['proxycommand'])
 
-        client.connect(**parameters)
-        self._ssh_connection = client
-        return client
+            # TODO configurable
+            #  if ssh_key_file:
+            #      parameters['key_filename'] = ssh_key_file
+            if 'identityfile' in user_config:
+                parameters['key_filename'] = user_config['identityfile']
+
+            client.connect(**parameters)
+            self._ssh_connection = client
 
 
 class Group(Host):
--- a/brigade/plugins/tasks/commands/remote_command.py
+++ b/brigade/plugins/tasks/commands/remote_command.py
@@ -22,6 +22,7 @@ def remote_command(task, command):
     Raises:
         :obj:`brigade.core.exceptions.CommandError`: when there is a command error
     """
+    task.host.ssh_connection = None
     client = task.host.ssh_connection
 
     chan = client.get_transport().open_session()
--- a/brigade/plugins/tasks/files/sftp.py
+++ b/brigade/plugins/tasks/files/sftp.py
@@ -130,6 +130,7 @@ def sftp(task, src, dst, action):
         "put": put,
         "get": get,
     }
+    task.host.ssh_connection = None
     client = task.host.ssh_connection
     scp_client = SCPClient(client.get_transport())
     sftp_client = paramiko.SFTPClient.from_transport(client.get_transport())


ktbyers:
Here is a POC using this:

```
brigade = Brigade(inventory=SimpleInventory("hosts.yml", "groups.yml"), dry_run=True)
a = brigade.filter(role='lab1')

net_connect = ConnectHandler(ip='cisco1.twb-tech.com', username='pyclass', password=getpass(),
                             device_type='cisco_ios')

test_host = a.inventory.hosts['pynet-rtr1']

# Assign the SSH connection to a host
test_host.ssh_connection = net_connect
print(test_host.ssh_connection)
print(test_host.ssh_connection.find_prompt())
```

I still need to update the Netmiko task code to check for the connection, use the connection, and assign the connection.

I am not sure how you want to do the brigade generated connection. I think it probably makes sense to just have that outside of the `setter` and then you could just pass that object in as such:

```
brigade_host.ssh_connection = new_connection
```

Currently, it will try establish an SSH connection using your SSH code if you do the following (which isn't very logical)

```
brigade_host.ssh_connection = None
```

ktbyers:
Note, the changes to the other two files were not how I think it should be, but were to get the unit tests to work.

dbarrosop:
> This is also a more clear pattern (instead of overloading the getter with a setter pattern)

This is a very common pattern: https://stackoverflow.com/questions/4037481/caching-attributes-of-classes-in-python

Overloading the setter as you did is actually quite strange because if you try to set it to `None` for some reason you are actually doing something completely different.

> I think it much clearer to call some method to create the SSH

I am fine with that but in that case the property code would be something like:

```
def connect_ssh(self):
    # code goes here
    self._ssh_connection = ...

@property
def ssh_connection(self):
    if not hasattr(self, "_ssh_connection"):
        self.connect_ssh()
    return self._ssh_connection
```

That way you could handle the connection yourself if you want to or just let brigade handle it for you. 

> We also might want to rename the attribute to attribute to be more generic like connection

What if the platform supports multiple mechanisms to manage the device? I manage my network equipment both using ssh against the underlying linux system and via its REST API. Other platforms are starting to support grpc for some things while still using netconf/whatevs for other. I personally think we should have named connections `ssh_connection`, `napalm_connection`, `grpc_connection` and let the tasks know which one is needed.

Looking at your poc I think we should have two ways of doing this:

1. A task that can handle creating the connecting in a custom way and ends up setting `self._ssh_connection`.
2. The property as we have today for people that don't care and is happy with the default method that just uses the inventory data. Although maybe using the `connect_ssh` function as you proposed in case there is some room for reusability in (1)

@ogenstad thoughts?

ogenstad:
I think this works as a PoC but I wouldn't want to have to do it like that. It should be easier to use.

```python
# Assign the SSH connection to a host
test_host.ssh_connection = net_connect
```

I also see a value of allowing multiple types of connections. At the same time I don't think we should end up with something like this:

* ssh_connection
* netmiko_connection
* napalm_connection
* grpc_connection
* winrm_connection
* socket_connection
* http_connection

I.e we have multiple properties on each host where most of them are never needed. Perhaps something like this could be used:

```python
client = task.host.get_connection('ssh')
client = task.host.get_connection('napalm')
client = task.host.get_connection('netmiko', **kwargs)
```
Then just have those return the correct connection.

dbarrosop:
an alternative is just provide a napalm_connect, ssh_connect, etc task that just sets self.whatever_connextion and get rid of the property completely.

ktbyers:
Yes, I think it is a very unintuitive that accessing an attribute causes assignment to occur. Especially for something significant like creating an SSH connection.


Yeah, I wouldn't do that None in practice I would just do something like this and get rid of the hidden creation of the ssh connection.

```
brigade_host.ssh_connection = ssh_connection()
```


On this code:

```
def connect_ssh(self):
    # code goes here
    self._ssh_connection = ...

@property
def ssh_connection(self):
    if not hasattr(self, "_ssh_connection"):
        self.connect_ssh()
    return self._ssh_connection
```

In the above scenario are you willing to at add a setter so that I could assign to the `ssh_connection` attribute directly? Otherwise you would be making the private attribute the path to assign `ssh_connection`. 

```
@ssh_connection.setter
def ssh_connection(self, connection):
    self._ssh_connection = connection
```


@dbarrosop Can you expand on this a little bit? I am not sure I understood this.

```
A task that can handle creating the connecting in a custom way and ends up setting self._ssh_connection.
```


@ogenstad Do you have something in mind here?

```
I think this works as a PoC but I wouldn't want to have to do it 
like that. It should be easier to use.

# Assign the SSH connection to a host
test_host.ssh_connection = net_connect
```

I think this would assignment would automatically happen in the task. For example, I would have the netmiko tasks check whether this attribute exists and use it if it does (and create a connection, assign it if it doesn't exit).

dbarrosop:
actually, the more I think about it the more sense I think it makes dealing with this with a task. That way you run it at the very beginning and if some devices fail to connect you can decide to abort or not. With the current pattern you can’t. Also, having a task means we can offload all the details/customizations there and let the rest of the tasks focus on just doing what they are supposed to be doing

dbarrosop:
@ktbyers I think our last few comments just cross paths without seeing each other :) can you take a look to my last one? I think we should just get rid of properties and functions and just a have task that creates the attribute that the upcoming tasks will expect.

dbarrosop:
another advantage of the task approach is that anyone can create connection types. Not just people working on the core of brigade.

ogenstad:
>another advantage of the task approach is that anyone can create connection types. Not just people working on the core of brigade.

Well that could be done regardless. Even if it's on the Host object we could make it so that people can send in their own Host object to the inventory.

I have people coming over for dinner and have to be brief. But I think a downside to having it in a separate task is for all of the scenarios where a persistent connection isn't needed. In those instances it wouldn't make sense to run another task just to open the connection. What do you guys think of the get_connection() concept?

>I think this would assignment would automatically happen in the task. For example, I would have the netmiko tasks check whether this attribute exists and use it if it does (and create a connection, assign it if it doesn't exit).

This was basically what I meant, that Brigade should handle everything so that you don't have to do this outside of Brigade.

dbarrosop:
Well, if you don’t need a persistent connection you can close it right afterwards or just let it there until it closes itself automatically. The idea is to have a simple pattern that is easy to use instead of 40 different, specially if they are not obvious because of A or B.

The get_connection is just a layer of indirection, unless I am missing something I don’t think it solves anything.

ktbyers:
@dbarrosop Yes, maybe we just have NAPALM, Netmiko, whatever automatically do the following (in the task):

```
check does brigade_host.napalm_connect exist
if not, create a brigade_host.napalm_connect
if it does, use the existing brigade_host.napalm_connect
```

dbarrosop:
> I think this would assignment would automatically happen in the task. For example, I would have the netmiko tasks check whether this attribute exists and use it if it does (and create a connection, assign it if it doesn't exit).

I’d rather avoid that. That means we have to add a lot of complexity to every task to check 20 paramaters (did it pass a host, ip, username, ssh_file, etc or should I use the ones from the inventory?) It’s much cleaner if the netmiko_sendconmand expects an already connected session and the only thing it has to bother with is sending a command and gathering the output.

dbarrosop:
@ktbyers what you suggest is exactly how it works right now for the paramiko tasks and it has the drawback that if we want to allow customizing the connections we have to duplicate a lot of complexity on EVERY task. I still think having tasks like napalm_connect, paramiko_connect, etc it’s the most flexible and simple way of dealing with this. You can put the customizations there and just have the rest of the tasks enjoy an already established connection without having to do checks or deal with anything.

ktbyers:
@dbarrosop and I talked via Slack. He is going to submit a counter-proposal and then we can discuss that.

ktbyers:
@dbarrosop I am totally fine with the task approach...I think that is the logical place to put it. I was previously planning on implementing this reuse of connections inside the tasks in Netmiko.

ogenstad:
I'm not very fond of having a separate task just to setup the connection. In that case I would prefer the current setup instead. My idea with `get_connection()` would work exactly like the current solution, the only difference is that it wouldn't expose several public properties. It would also solve @ktbyers's issue of not having a property you can set to something strange. So `get_connection('paramiko')` would just return ._connection_paramiko once a connection is established.

If a user only wants to run `napalm_get` I think it just complicates things for users that they first have to have a task for `napalm_connect` and then after that `napalm_get`. Or if I want to transfer a file to a host, why should I have to use two tasks when the current implementation only uses one?

ogenstad:
> I think it just complicates things for users that they first have to have a task for napalm_connect and then after that napalm_get. 

I guess this could be solved by having napalm_get always run napalm_connect so that the user doesn't have to run two tasks.

dbarrosop:
I opened #38 borrowing on the ideas discussed here. Including the `get_connections` proposal which I have mixed feelings about but I see how it can be convenient.

PR #38, opened 2017-12-10
connections as tasks

dbarrosop:
This PR moves the logic to connect to the device to a task removing the `ssh_connection` property and introducing a `get_connection` function. The `get_connection` function will try to find an already established connection or call the relevant task to establish it.

What I like about this:

1. We keep the core generic. The logic to connect to devices via different mechanisms is kept outside as tasks.
2. It also means that if we want to customize the connection (by sending specific parameters like a non-default ssh_config/key or whatever) we can do it in the connection task and keep the rest of the tasks that uses that connection cleaner, avoiding duplicating code leading to inconsistencies.
3. If you are happy with the defaults and/or with the data coming from the inventory you don't have to call the task yourself as the first time `get_connection` is called will establish the session automatically.

Alternative to the discussion held here #35 

@ktbyers @ogenstad thoughts?


ogenstad:
This is pretty much what I was thinking of yesterday with the added benefit of not having all of the code within the core inventory. So I'm very happy with this solution.

Another thing that also could be valuable is to have some config key so that the connections are validated on the following tasks. This could be such as if a long time as passed, or if you would upgrade a device during a run and then want to reconnect, or just having something which does some is_alive() and reconnects if the connection isn't actually working. But that could be something to revisit later.

I'm happy with how this turned out. So +1 from me.

ktbyers:
@dbarrosop @ogenstad What if we modified this slightly?

```
+    client.connect(**parameters)
 +    host.paramiko_connection = client
```

Such that we made assignment happen to:

```
brigade_host.connection_map['paramiko_ssh'] = connection
```

This would give a uniform place to look for existing connections and make it easier to figure out which connections exist (i.e. some iterable we can just loop over to find all connections).

But either way...the solution sounds good.

dbarrosop:
Yeah, I think using the dict makes sense, will fix in a bit.

ktbyers:
+1

ktbyers:
@dbarrosop Are you willing to do something like this?

This would let me use this for brigade inventory and for non-brigade inventory. I need to test this, but was starting to work on this today:

```
173     def get_connection(self, connection, *args, **kwargs):
174         """
175         This function will try to find an already established connection
176         or call the task that establishes the connection if none is found.
177 
178         Arguments:
179             connection_name (str): Name of the connection, for instance, netmiko, paramiko,
180                 napalm...
181         """
182         if connection not in self.connections:
183             task_name = "{}_connection".format(connection)
184             if not kwargs.get("host"):
185                 kwargs['host'] = self
186             getattr(connections, task_name)(*args, **kwargs)
187         return self.connections[connection]
```

PR #40, opened 2017-12-11
Add tests for Sphinx

ogenstad:


ogenstad:
Ok, that didn't work with Travis. Will test some more.

ogenstad:
Added the tests of Sphinx to all environments instead.

dbarrosop:
Great!

PR #42, opened 2017-12-11
Netmiko connection reuse and netmiko textfsm

ktbyers:
.


ktbyers:
Note, I added the **kwargs to get_connection to allow for arguments to be passed in outside of Brigade inventory.

In other words, I wanted to less tightly couple the connection management from the inventory so that we could use other inventory plugins (or not have an inventory).

Though maybe connection management and inventory is still too tightly coupled since it is a method of a Host object.

Do we want to consider totally decoupling connections from Brigade the Host class? Something like a Connection class. The Brigade Host could then assign this connection class as an attribute of the host.

This would allow other Brigade inventory plugins or no-inventory situation to still re-use/track the connections in Brigade (in a more uniform way without being coupled to a Brigade Host object).

ktbyers:
Need to add a way to gracefully cleanup a connection and cleanup all connections.

brigade/plugins/tasks/networking/netmiko_tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    if cmd_args is None:
+        cmd_args = ()
+    if cmd_kwargs is None:
+        cmd_kwargs = {}
+    result = netmiko_method(*cmd_args, **cmd_kwargs)
+    return Result(host=task.host, result=result)
+
+
+@netmiko_get_connection
+def netmiko_send_command(task, ip=None, host=None, username=None, password=None,
```
dbarrosop: What's the use case for duplicating all the connection parameter on all the tasks? Why can't the user call `netmiko_connection` if some customization is needed and then just call `netmiko_send_command`? I think we should handle connection only in the connection task to avoid duplicating parameters in many tasks (that could lead to inconsistencies if we don't update all the tasks when something changes) and to avoid bloating the documentation with lots of parameters.

brigade/core/inventory.py
```diff
@@ -192,12 +192,11 @@ def get_connection(self, connection):
         """
         if connection not in self.connections:
             task_name = "{}_connection".format(connection)
-            try:
-                task = getattr(connections, task_name)
-            except AttributeError:
```
dbarrosop: why did you remove this?

brigade/core/inventory.py
```diff
@@ -192,12 +192,11 @@ def get_connection(self, connection):
...
-                raise AttributeError("not sure how to establish a connection for {}".format(
-                    connection))
-            task(host=self)
+            if kwargs.get("host") is None:
```
dbarrosop: I don't like this `if else` here. The core of brigade shouldn't care about specifics of tasks so whatever error handling a particular task needs should be handled in the task, not in the core.

brigade/core/inventory.py
ktbyers: That might have been a merge-timing issue...I might have missed it in the merge conflict. I will re-add.

brigade/core/inventory.py
ktbyers: That is fine, will remove. 

I debated whether it made sense to raise an error here and thought it did as `host` is a pretty generic argument and both Netmiko and Brigade use `host` arguments.

dbarrosop:
> Need to add a way to gracefully cleanup a connection and cleanup all connections.

We can add "close_$connection_name" tasks that users can call at any time if they want to. But I don't think we need to handle those automatically ourselves. Garbage collector should call `__del__` and close the connections automatically:

https://github.com/napalm-automation/brigade/pull/25#issuecomment-349945205

brigade/plugins/tasks/networking/netmiko_tasks.py
dbarrosop: I think our philosophy here basically should be to have lots of small and simple tasks rather than few and very complex. I'd prefer to have smaller and simpler workflows than end up writing this monster again:

https://github.com/napalm-automation/napalm-ansible/blob/develop/napalm_ansible/napalm_install_config.py

IMHO, because we are dealing with python code directly we are in a good position to make things simpler to write, maintain and understand.

ktbyers: The use was to have the task automatically setup the connection if it didn't exist...so end-users didn't have the extra step of the connection setup call.

I would say this should be a prerogative of the task...i.e. Brigade shouldn't care.

That all being said...I will change it as the argument management is getting ugly.

ktbyers:
@dbarrosop Agreed we shouldn't try to do anything to close the connection ourself, but should just have `close_$connection_name` so that the end-users can gracefully close the connection (freeing up the VTY port on SSH) and so the entry in the `connection_map` gets cleanly removed.

dbarrosop:
> This would allow other Brigade inventory plugins

This works already, I am not using the default inventory at $DAYJOB. I should probably document how to create a custom inventory, it's extremely easy though but an example is always welcome. For reference:

```
from brigade.core.inventory import Inventory

import requests


class CustomInventory(Inventory):

    def __init__(self, params=None, **kwargs):
        # code here to retrieve hosts and group information from my REST backend
        # I use params to filter the data in the backend directly
        # and avoid having to filter unnecessarily thousands of devices
        super().__init__(hosts, groups, **kwargs)
```

> no-inventory situation to still re-use/track the connections in Brigade (in a more uniform way without being coupled to a Brigade Host object).

That'd be tricky but maybe can be solved by doing something like:

```
def my_task(param1, param2, **kwargs):
      # param1 and param2 are parameters that are required by the task
      brigade_params = kwarge.pop("brigade_params")
      host = brigade_params["host"]
      brigade = brigade_params["brigade"]
      dry_run = brigade_params["dry_run"]
```

However, that could lead to some tasks being reusable while some others aren't. A simpler and maybe nicer alternative would be to do:

```
from brigade.plugins.functions import something

def my_task(task, param1, param2):
    connection = task.get_connection("netmiko")
    something.my_task_as_a_function(connection, param1, param2, task.dry_run)
```

That way it is very clear that `something.my_task_as_a_function` is reusable while the tasks aren't.

ktbyers:
Will re-submit with proposed changes...might be a few days though.

brigade/plugins/tasks/networking/netmiko_tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    if cmd_kwargs is None:
+        cmd_kwargs = {}
+
+    if len(cmd_args) >= 1:
+        command = cmd_args[0]
+    else:
+        command = cmd_kwargs['command_string']
+
+    # Netmiko connection automatically created by 'netmiko_get_connection' decorator (if needed)
+    result = netmiko_conn.send_command(*cmd_args, **cmd_kwargs)
```
ogenstad: This part still feels very strange to me. 99% of the time I just want to send in a command using send_command. I think there should be one very simple task that just does that.

tests/tasks/networking/test_netmiko_run.py
```diff
@@ -4,9 +4,10 @@
 class Test(object):
 
     def test_netmiko_run(self, brigade):
+        cmd_kwargs = {'command_string': 'hostname'}
         result = brigade.filter(name="dev4.group_2").run(networking.netmiko_run,
                                                          method="send_command",
-                                                         command_string="hostname")
+                                                         cmd_kwargs=cmd_kwargs)
```
ogenstad: To reiterate from the last comment. I think what's in the current tests for this and the current implementation is much cleaner.

brigade/plugins/tasks/networking/netmiko_tasks.py
```diff
@@ -0,0 +1,180 @@
...
+    if not os.path.isdir(template_dir) or not os.path.isfile(index):
+        msg = """
+Valid ntc-templates not found, please install https://github.com/networktocode/ntc-templates
+and then set the NET_TEXTFSM environment variable to point to the ./ntc-templates/templates
+directory."""
+        raise ValueError(msg)
+    return template_dir
+
+
+def get_structured_data(raw_output, platform, command):
```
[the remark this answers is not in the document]
dbarrosop: I agree with this. We could even provide the `collect_data` task but I like the idea of having lots of small tasks to build workflows and potentially bigger tasks using those small tasks to create bigger workflows.

tests/tasks/networking/test_netmiko_run.py
dbarrosop: Yeah, I agree, I did that change while writing the tests and trying to figure out some weird error that said something like `e` is not a valid something (I send the command without being enclosed in an iterable). I don't have much experience with netmiko so my error was due to not knowing the library but we should assume users using this task don't know the library either. Otherwise they could just do `host.get_connection('netmiko').send_command(...)` themselves :P

brigade/plugins/tasks/connections/netmiko_connection.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import unicode_literals
+
+from brigade.plugins.tasks.networking.netmiko_tasks import netmiko_args
```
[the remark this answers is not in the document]
dbarrosop: I think we agreed on handling the connection only in the connection task and avoid repeating all the arguments on each task so I guess this change is pending already :)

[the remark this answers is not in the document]
ktbyers: Yeah, that is probably fine as I was using it in `netmiko_tasks` as well and that is going to go away.

brigade/plugins/tasks/networking/netmiko_tasks.py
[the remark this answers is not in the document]
ktbyers: I am probably going to push all of that textfsm-glue code into Netmiko as I have another place I use it and it is generally valuable ('get_template_dir', 'get_structured_data', 'clitable_to_dict').

I will probably push the `use_textfsm` argument into Netmiko's send_command and send_command_timing as well. If not, I will leave it as an argument on the task.

brigade/plugins/tasks/networking/netmiko_tasks.py
ktbyers: I will probably fix for `netmiko_send_command` as given most of the arguments are going to go away, the reason I wanted to do it is no longer there. Note, this doesn't apply to `netmiko_run_command` which I still think makes sense to do it as I originally did. Which I will expand on below.

tests/tasks/networking/test_netmiko_run.py
ktbyers: As per above, I will probably fix this in `netmiko_send_command` as the connection arguments are going away so it doesn't make sense to do it there.

It still does make sense to use the original solution with netmiko_run, however, as netmiko_run can call any Netmiko method and there are a lot of them. 

In other words, since any Netmiko method can be called, it makes sense to have clear separation between the arguments that are for the task and the arguments that will be passed to the method. It will also make it clearer on the documentation. It will also make it less likely to introduce end-user code problems due to changes in the task definition.

dbarrosop: Note that users can always do:

```
def something(task):
    task.host.get_connection(“netmiko”).netmiko_method(arg1, arg2)

brigade.run(something)
```

So basically our tasks have to provide value over the simplicity of just those 2 lines.

PR #44, opened 2017-12-13
added nsot's inventory plugin and transform_function to inventory

dbarrosop:
So I have added a `NSOT` inventory and I will probably use it to document how to write a custom inventory. If you want to try it out you can do:

```
docker run -p 8990:8990 -d --name=nsot nsot/nsot start --noinput
bash nsot.sh  # after nsot is up and running, which takes a few seconds.
```

You will have to configure `~/.pynsotrc` with:

```
[pynsot]
url = http://localhost:8990/api
auth_method = auth_header
auth_header = X-NSoT-Email
```

But basically, what you should get is (note that nsot doesn't have the concept of groups):

``` python
In [1]: from brigade.plugins.inventory import nsot

In [2]: n = nsot.NSOTInventory()

In [3]: n.hosts                                                                                                                                                  
Out[3]:
{u'rtr00-site1': Host: rtr00-site1,
 u'rtr00-site2': Host: rtr00-site2,
 u'rtr01-site1': Host: rtr01-site1,
 u'rtr01-site2': Host: rtr01-site2}

In [4]: n.filter(site='site1').hosts
Out[4]: {u'rtr00-site1': Host: rtr00-site1, u'rtr01-site1': Host: rtr01-site1}

In [5]: n.filter(site='site2').hosts                                                                                                                             
Out[5]: {u'rtr00-site2': Host: rtr00-site2, u'rtr01-site2': Host: rtr01-site2}

In [6]: n.filter(os='junos').hosts                                                                                                                               
Out[6]: {u'rtr01-site1': Host: rtr01-site1, u'rtr01-site2': Host: rtr01-site2}

In [7]: n.filter(os='junos', site='site1').hosts
Out[7]: {u'rtr01-site1': Host: rtr01-site1}

In [8]: n.filter(os='junos', site='site1').hosts['rtr01-site1'].data
Out[8]:
{u'asn': u'65002',
 'group': None,
 u'host': u'127.0.0.1',
 u'hostname': u'rtr01-site1',
 u'id': 2,
 'interfaces': {u'ge-0/0/1': {u'addresses': [u'2001:db8:caf3::1/128'],
   u'attributes': {u'connects_to_device': u'rtr00',
    u'connects_to_iface': u'et1',
    u'link_type': u'fabric'},
   u'description': u'',
   u'device': 2,
   u'device_hostname': u'rtr01-site1',
   u'id': 5,
   u'mac_address': None,
   u'name': u'ge-0/0/1',
   u'name_slug': u'rtr01-site1:ge-0/0/1',
   u'networks': [u'2001:db8:caf3::/127'],
   u'parent': None,
   u'parent_id': None,
   u'speed': 1000,
   u'type': 6},
  u'ge-0/0/2': {u'addresses': [u'2001:db8:caf3::3/128'],
   u'attributes': {u'connects_to_device': u'rtr00',
    u'connects_to_iface': u'et2',
    u'link_type': u'fabric'},
   u'description': u'',
   u'device': 2,
   u'device_hostname': u'rtr01-site1',
   u'id': 6,
   u'mac_address': None,
   u'name': u'ge-0/0/2',
   u'name_slug': u'rtr01-site1:ge-0/0/2',
   u'networks': [u'2001:db8:caf3::2/127'],
   u'parent': None,
   u'parent_id': None,
   u'speed': 1000,
   u'type': 6},
  u'lo0': {u'addresses': [u'2001:db8:b33f::101/128'],
   u'attributes': {u'connects_to_device': u'loopback',
    u'connects_to_iface': u'lo0',
    u'link_type': u'loopback'},
   u'description': u'',
   u'device': 2,
   u'device_hostname': u'rtr01-site1',
   u'id': 2,
   u'mac_address': None,
   u'name': u'lo0',
   u'name_slug': u'rtr01-site1:lo0',
   u'networks': [u'2001:db8:b33f::/64'],
   u'parent': None,
   u'parent_id': None,
   u'speed': 1000,
   u'type': 6}},
 'name': u'rtr01-site1',
 u'os': u'junos',
 u'password': u'',
 u'port': u'12203',
 u'router_id': u'10.1.1.2',
 'site': u'site1',
 u'site_id': 1,
 u'user': u'vagrant'}
```


ogenstad:
Looks nice! A description around this should go in the howto section within the docs.

I'm guessing this is mostly an example for now, but the defaults in the inventory are perhaps a bit strange. Should only use the environment variables or a combination of those and arguments to the class.

Also the placement of nsot.sh should be moved. Perhaps howto/nsot-inventory.

>nsot devices update --site-id 2 --id 3 -a os=eos -a host=127.0.0.1 -a user=vagrant -a password=vagrant -a port=12443

This currently wouldn't work with the Napalm tasks as the inventory will look for "brigade_nos" and "brigade_username" instead of "os" and "user". The same with port and password.

However I don't think people should need to build their inventory to match what Brigade defaults to. A solution could be to also send in a custom Host object. I also thought about being able to set the defaults in the config for what value the host properties used. I.e the network_api_port property defaults to brigade_network_api_port, but if someone wanted to they could change that default. I think you already mentioned that you did something custom for the property lookups?

dbarrosop:
> I'm guessing this is mostly an example for now

yeah, just dumped what I had yesterday, will create tests and move the nsot.sh somewhere to populate a docker container we can use to test this task.

> This currently wouldn't work with the Napalm tasks as the inventory will look for "brigade_nos" and "brigade_username" instead of "os" and "user". The same with port and password.

I know, just copied the script from some other stuff I had. Can change but if it's only for testing the task it doesn't really matter.

> However I don't think people should need to build their inventory to match what Brigade defaults to. 

Regarding that, didn't do it because it was late but I was thinking the same and I figured easiest would to add a `xfrm` function. So we can do something like:

```
def xfrm_do_changes(host):
      host.data["brigade_user"] = host.data["user"]
      host.data["brigade_password"] = host.data["password"]
      # maybe some other changes as well

inventory = NSOTInventory(xfrm=xfrm_do_changes)
```

Then, later on, in the base `Inventory` we could do:

```
if xfrm:
    for h in self.host.values():
        xfrm(h)
```

That way any inventory should be able to do any tweaks needed.

brigade/plugins/inventory/nsot.py
```diff
@@ -0,0 +1,36 @@
...
+            d['site'] = sites[d['site_id'] - 1]['name']
+            d['interfaces'] = {}
+
+            # We assign attributes to the root
+            for k, v in d.pop('attributes').items():
+                d[k] = v
+
+        # We assign the interfaces to the hosts
+        for i in interfaces:
+            devices[i['device'] - 1]['interfaces'][i['name']] = i
```
dbarrosop: I realize this is not going to work if we remove a device. So I have to fix it.

ogenstad:
```python
def xfrm_do_changes(host):
      host.data["brigade_user"] = host.data["user"]
      host.data["brigade_password"] = host.data["password"]
      # maybe some other changes as well

inventory = NSOTInventory(xfrm=xfrm_do_changes)
```

Yes, this would be nice. Perhaps named in a different way though.

brigade/core/__init__.py
```diff
@@ -123,8 +123,8 @@ def run(self, task, num_workers=None, **kwargs):
             **kwargs: additional argument to pass to ``task`` when calling it
 
         Raises:
-            :obj:`brigade.core.exceptions.BrigadeExceptionError`: if at least a task fails
-              and self.config.raise_on_error is set to ``True``
+            :obj:`brigade.core.exceptions.BrigadeExecutionError`: if at least a task fails
+              and self.raise_on_error is set to ``True``
```
ogenstad: You replaced `self.config.raise_on_error` here with `self.raise_on_error`. Was that a mistake?

ogenstad:
+1

PR #47, opened 2017-12-13
Netmiko connection 

ktbyers:


ktbyers:
See here for additional reference;

https://github.com/napalm-automation/brigade/pull/45

ktbyers:
@dbarrosop Where do you want it documented?

Here is the error you will get when this happens (though should be rare):

```
AttributeError: 'str' object has no attribute 'host'
```

Basically, you will pass in a host to netmiko_connect and then you will get an error because that will be a string and not a Brigade `host object`.

ktbyers:
Okay, updated and added a message to more clearly indicate the issue (if someone tries to incorrectly pass in a Netmiko 'host' argument).

brigade/plugins/tasks/connections/netmiko_connection.py
```diff
@@ -0,0 +1,46 @@
...
+    'nxos': 'cisco_nxos',
+    'eos': 'arista_eos',
+    'junos': 'juniper_junos',
+    'iosxr': 'cisco_iosxr'
+}
+
+
+def netmiko_connection(task=None, host=None, **netmiko_args):
+    """Connect to the host using Netmiko and set the relevant connection in the connection map.
+
```
dbarrosop: Should we add a note about `netmiko_host` mapping to `host` in netmiko's ConnectHandler?

dbarrosop:
Yeah, that's smart. The only comment I have is if it's worth adding a note to the docstring regarding that so it shows in the documentation.

dbarrosop:
other than the previous comment +1

PR #48, opened 2017-12-14
removing host from connection tasks as it's no longer needed

dbarrosop:
@ogenstad brought up that after moving the connection to a task `host` is no longer needed, which is true, so I am removing it. This also simplifies netmiko's connection as there is no longer a collision.

@ktbyers @ogenstad

### DIFF
--- a/brigade/plugins/tasks/connections/netmiko_connection.py
+++ b/brigade/plugins/tasks/connections/netmiko_connection.py
@@ -9,26 +9,13 @@ napalm_to_netmiko_map = {
 }
 
 
-def netmiko_connection(task=None, host=None, **netmiko_args):
+def netmiko_connection(task=None, **netmiko_args):
     """Connect to the host using Netmiko and set the relevant connection in the connection map.
 
     Arguments:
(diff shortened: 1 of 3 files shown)


brigade/plugins/tasks/connections/netmiko_connection.py
```diff
@@ -9,26 +9,13 @@
...
-
-    try:
-        host.host
-    except AttributeError:
-        msg = "Both Netmiko and Brigade have a host argument. Use the 'netmiko_host' argument " \
-              "or the 'ip' argument to specify a device to connect to (if not specifying in " \
-              "Brigade's inventory)."
-        raise AttributeError(msg)
-
+    host = task.host
```
ktbyers: This section should get removed also:

```
     # Both netmiko and brigade use host, allow passing of an alternately named host argument
     if netmiko_args.get("netmiko_host"):
         netmiko_host = netmiko_args.pop("netmiko_host")
         netmiko_args['host'] = netmiko_host
```

ktbyers:
+1

PR #50, opened 2017-12-15
connect automatically

dbarrosop:
So I added the autoconnect back, however:

1. In `napalm_connection` we will try to get `napalm_options` and pass them as `optional_args` (we can do the same for the rest.
2. Instead of having connections work with either a `task` or a `host` I call the connection task from the host. For that I needed to have access to the parent brigade object from the host so tasks that are configurable could read it from there. For instance: https://github.com/napalm-automation/brigade/blob/develop/brigade/plugins/tasks/connections/paramiko_connection.py#L18
3. I also added `available_connections` so the connections that are available for the tasks can be customized


brigade/core/__init__.py
```diff
@@ -68,6 +70,10 @@ def __init__(self, inventory, dry_run,
             level=logging.ERROR,
             format=format,
         )
+        if available_connections is not None:
+            self.available_connections = available_connections
+        else:
+            self.available_connections = connections.available_connections
```
ogenstad:
```python
self.available_connection = available_connections or connections.available_connections
```
?

Or perhaps:

```python
self.available_connection  = connections.available_connections
if available_connections:
    self.available_connections.update(available_connections)
```

The default available_connections dict could be named brigade_connections, or core_connections.

ogenstad:
Aside from my small comment above I think it looks good.

The available_connections will need some documentation. So perhaps a few lines about that in the class.

brigade/core/__init__.py
dbarrosop: No, I don't want to force people to use the ones we provide. So if they want to customize it they can do the update themselves. 

I didn't want to do `available_connections or connections.available_connections` in case someone passed `{}` deliberately but I am not sure about the use case so maybe I should do that...

ogenstad: But for the second option if someone wanted to overwrite the default `self.available_connections['napalm']` they could just pass in `napalm` as a key in their dict of connections while at the same time preserving the shipped version of the `paramiko` connection.

dbarrosop: Yeah, but imagine you want to ensure only connections provided by you are used. Because security or compliance or whatever. Easiest would be to just pass a full dict with your modules. Otherwise you'd have to check the ones provided by the `core` and pass them in your own dict with a dummy function or something that triggers an error.

dbarrosop:
Yeah, there are a few things that need documentation here. I was mostly waiting for agreement before doing that :P

brigade/core/__init__.py
ogenstad: I would assume that it would be more common that someone wants to overwrite one of the connection types. If I want to overwrite the paramiko one but keep the napalm one I would need to send in my own paramiko connection but also send in the core napalm connection. For security reasons you can also reach into self.available_connections and delete the other ones.

I think the most common scenario will be that someone has one extra connection type. Like if you had some rest-api which takes a username and password and after that you authenticate with a token. If I brought that to Brigade I might still want to use napalm, paramiko and netmiko, but now I have to spend extra time to recreate them.

I don't think any of these will be super common scenarios though. :)

ogenstad:
:) I heard that Documentation Driven Development is the new black.

brigade/core/__init__.py
dbarrosop: Yeah, but what you are suggesting can be done as:

```
brigade.plugins.tasks.connection.available_connections.update({"napalm": custom_napalm_connection})
```
I don't even need to actually pass it to brigade as it will be read from there by default anyway :P

While the case while I want to remove connections would be with your suggestion:

```
my_connections = {"napalm": my_custom_napalm}
for k in brigade.plugins.tasks.connection.available_connections.keys():
    my_connections[k] = dummy_connection_that_fails

brigade = Brigade(available_connections=my_connections)
```

dbarrosop: And I agree this is going to be strange (specially disabling connections), if anything the second case and I think it's equally easy to accommodate with both yours and my solution while mine makes less assumptions of how things should work :P

ogenstad:
I didn’t say it would be a pretty way to remove connections :) Just meant that it wouldn’t be hard. 

My main point is that I think my scenario is more probable to happen than the one you describe. 

It could be solved with a key in Config like external_connections_policy which takes “merge” or “replace”. 

Don’t know if this just makes something simple overly complex.

dbarrosop:
> It could be solved with a key in Config like external_connections_policy which takes “merge” or “replace”.

I'd rather not have keys to change the behavior of things if we can help it. As your proposal and mine result in almost the same convenience for the user I'd rather stick with something that is generic and does't make assumptions on how users are supposed to use things.

Also, each key we add increases the chances users might get something wrong :P

dbarrosop:
In any case, no matter what we decide we need the `available_connections` argument because of how the `filter` works so I'd suggest not getting stuck with this conversation as what you are suggesting can be an improvement later on.

@ktbyers do you have any opinion about this PR? if you are good with it let me know what you want me to add to `netmiko` (or push a commit yourself directly to this branch)

ogenstad:
+1 works nicely (even if the default behaviour for available_connections is wrong :) )

brigade/core/__init__.py
```diff
@@ -52,9 +53,10 @@ class Brigade(object):
         config (:obj:`brigade.core.configuration.Config`): Configuration parameters
     """
 
-    def __init__(self, inventory, dry_run,
-                 config=None, config_file=None):
+    def __init__(self, inventory, dry_run, config=None, config_file=None,
+                 available_connections=None):
         self.inventory = inventory
+        self.inventory.brigade = self
```
ktbyers: Shouldn't this be `self.inventory._brigade = self`?

brigade/core/inventory.py
```diff
@@ -189,11 +198,12 @@ def get_connection(self, connection):
...
-                "Couldn't find an established connection for '{c}'. "
-                "Did you call '{c}_connection'?"
-            ).format(c=connection)
-            raise AttributeError(msg)
+            try:
+                conn_task = self.brigade.available_connections[connection]
+            except KeyError:
+                raise AttributeError("not sure how to establish a connection for {}".format(
+                    connection))
+            self.brigade.filter(name=self.name).run(conn_task, num_workers=1)
```
ktbyers: What is the purpose of the `filter` here?

You might want to but a  comment in the code here (since it isn't super clear from the code that conn_task and adds the connection and thus `get_connection` returns the newly created connection in the return statement. 

```
# Establishes the connection, conn_task must add connection to self.connection['connection_name']
self.brigade.filter(name=self.name).run(conn_task, num_workers=1)
```

dbarrosop: > What is the purpose of the filter here?

Will add a comment but the idea is that I wanted to call the connection task in the same way as a user would call it explicitly so the filter is basically saying "run this only for myself". The reason to call it like this is for two reasons:

1. To avoid having `task=None, host=None` and then `host = host or task.host` as we had before (remember that `netmiko_host` issue we had :P)
2. To avoid surprises. If we use the API in the documented way there are less chances of surprises.

> You might want to but a comment in the code here (since it isn't super clear from the code that conn_task and adds the connection and thus get_connection returns the newly created connection in the return statement.

Maybe I misinterpreted this but I think this requires two distinct documentation points:

1. We need to make very clear that `get_connection` has this behavior of establishing the connection if none exists. I will do that in the docstring of this method so users reading the docs see it. Probably a quick "how to" document would be interesting as well.
2. We need to make also very clear that if you write a connection task, the task has to set `host.connections[connetion_name]`. For that I will write a "how to write a connection task" similar to this: http://brigade.readthedocs.io/en/develop/howto/writing_a_custom_inventory.html

brigade/core/__init__.py
dbarrosop: I think we should document it and make it official. It's useful to access other hosts' data and the configuration.

ktbyers:
Everything else looks good to me.

brigade/core/inventory.py
ktbyers: Sounds good on your two points.

I also think we should just add a comment here:

```
# Establishes the connection, conn_task also adds connection to
# self.connection['connection_name']
self.brigade.filter(name=self.name).run(conn_task, num_workers=1)
```

I have looked at this code twice now (at least) and each time I have initially made the error of thinking it is not returning the connection (as it is not obvious at this point that the 'conn_task' call automatically does that).

brigade/core/__init__.py
ktbyers: @dbarrosop Yes, but this line 59 needs to be fixed correct and not be `self.inventory.brigade` and instead be `self.inventory._brigade` (i.e it is missing the underscore)?

dbarrosop: why? if we want to make it official we shouldn’t mark it as private.

brigade/core/inventory.py
dbarrosop: ok, will do

brigade/core/__init__.py
ktbyers: My error, what you have here...should be good.

dbarrosop: cool, thanks. Let me know if you want me to fix netmiko’s connection task. Otherwise I will fix napalm’s and the documentation and merge later today.

dbarrosop:
As people seems to agree overall I am going to merge. We can iterate if there is something to be changed.

PR #52, opened 2017-12-17
moar docs and fixes for issues I found while working with the examples

dbarrosop:
Ok, while working on the examples and adding more docs I fixed

1. A mechanism to to "skip" hosts in case of `raise_on_error=False`. You can see the fixes:
 * https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-db62ea4f5faa3283214fcfd457f3f64e
 * https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-3809cfaf6a7bf2bf113debc68df5541c
 * https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-118676f40dc730bc62bf5d116683f4abR86
 * https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-118676f40dc730bc62bf5d116683f4abR36

2. I also did a slight change to `Config` class so it can be used programmatically.

3. Added this method: https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-118676f40dc730bc62bf5d116683f4abR43. Which basically allows us to do the following: https://github.com/napalm-automation/brigade/compare/moar_docs?expand=1#diff-e608b5f82f50e5a951b67eb57de24dacL23. Basically we can thanks to this run tasks and subtasks in the same exact way from a user perspective

4. Tiny bugfix to `napalm_get`.

5. Fixed `examples/get_facts/get_facts.py`
6. Added docs

TBD: Fix `examples/configure/` and add a few more (maybe these can wait?)


dbarrosop:
sending smaller PRs

PR #55, opened 2017-12-17
making Config configurable programmatically

dbarrosop:
The idea is that I want to be able to do:

```
    brigade = Brigade(
        inventory=SimpleInventory("../hosts.yaml", "../groups.yaml"),
        dry_run=True,
        config=Config(raise_on_error=False),
    ) 
```

### DIFF
--- a/brigade/core/configuration.py
+++ b/brigade/core/configuration.py
@@ -41,7 +41,7 @@ class Config:
         config_file(``str``): Yaml configuration file.
     """
 
-    def __init__(self, config_file=None):
+    def __init__(self, config_file=None, **kwargs):
 
         if config_file:
             with open(config_file, 'r') as f:
@@ -50,6 +50,9 @@ class Config:
             c = {}
 
         self._assign_properties(c)
+
+        for k, v in kwargs.items():
+            setattr(self, k, v)
 
     def _assign_properties(self, c):
 


ogenstad:
I like the idea of this. However, one potential issue I see is that people can set their own config keys in this way. That might cause a problem if we introduce that key as a "real" config value in a later release.

This could be handled by only allowing keys that exist in CONF.

Or perhaps allowing users to set their own config keys in this way by using a prefix like `_` or similar. That way we could also load "private" keys if someone sent in a yaml files containing those to the config.

dbarrosop:
> However, one potential issue I see is that people can set their own config keys in this way. That might cause a problem if we introduce that key as a "real" config value in a later release.

Well, that's true for any class. They could do the same in any other class.

> Or perhaps allowing users to set their own config keys in this way by using a prefix like _ or similar

I have been thinking about this and struggling with a decent solution. I think we shouldn't support what you suggest. We can provide "functions" to help configure an application (maybe even generalize the `Config` class so it can be reused) but what we are talking about is the object we pass to `brigade` so we shouldn't allow misuse. I'd suggest merging this as is and then discuss on a different PR a generic config class for users and a set of functions to let them configure their in a similar way as `brigade` can be configured.

In summary, I am merging this one and open an issue to track this new feature :)

PR #56, opened 2017-12-17
added task.run to help executing subtasks

dbarrosop:
The idea is to be able to run subtasks in the same way we run tasks. See the test for an example.


ogenstad:
+1

PR #57, opened 2017-12-17
if raise_on_error=False skip following tasks for hosts that failed

dbarrosop:
With the current behavior a host that failed a task will run the rest of the tasks if `raise_on_error=False`. This changes the behavior to skip the execution on subsequent tasks.

Requires #56

### DIFF
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -36,6 +36,19 @@ if sys.version_info.major == 2:
 logger = logging.getLogger("brigade")
 
 
+class Data(object):
+    """
+    This class is just a placeholder to share data amongsts different
+    versions of Brigade  after running ``filter`` multiple times.
+
+    Attributes:
+        failed_hosts (list): Hosts that have failed to run a task properly
+    """
+
+    def __init__(self):
+        self.failed_hosts = set()
+
+
 class Brigade(object):
     """
     This is the main object to work with. It contains the inventory and it serves
@@ -43,6 +56,7 @@ class Brigade(object):
 
     Arguments:
         inventory (:obj:`brigade.core.inventory.Inventory`): Inventory to work with
+        data(:obj:`brigade.core.Data`): shared data amongst different iterations of brigade
         dry_run(``bool``): Whether if we are testing the changes or not
         config (:obj:`brigade.core.configuration.Config`): Configuration object
         config_file (``str``): Path to Yaml configuration file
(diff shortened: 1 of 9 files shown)


brigade/core/__init__.py
```diff
@@ -36,13 +36,27 @@ def _unpickle_method(func_name, obj, cls):
...
+    """
+    This class is just a placeholder to share data amongsts different
+    versions of Brigade  after running ``filter`` multiple times.
+
+    Attributes:
+        failed_hosts (list): Hosts that have failed to run a task properly
+    """
+
+    def __init__(self):
+        self.failed_hosts = set()
```
ogenstad: Should we also have a `self.tracebacks = set()` here?

dbarrosop: what would be the purpose? Long term I think it'd be nice to just have a record of everything that was run, with what parameters and the result. Something you can actually dump for debugging purposes.

ogenstad: Just to have the tracebacks being saved so that people can collect all of them from the same location. This can of course be handled for each task too.

brigade/core/task.py
```diff
@@ -67,19 +69,22 @@ class Result(object):
...
     Attributes:
         changed (bool): ``True`` if the task is changing the system
         diff (obj): Diff between state of the system before/after running this task
         result (obj): Result of the task execution, see task's documentation for details
         host (:obj:`brigade.core.inventory.Host`): Reference to the host that lead ot this result
+        skipped (bool): ``True`` if the host was skipped because there was a previous error
     """
 
-    def __init__(self, host, result=None, changed=False, diff="", **kwargs):
+    def __init__(self, host, result=None, changed=False, diff="", skipped=False, **kwargs):
```
dbarrosop: I am wondering if it'd be interesting to not assume we skipped due to an error and add a "reason" as well.

ogenstad: What do you mean by a reason? Do you have any example of what that might be.

ogenstad:
Looks like the tests are failing because the xyz_connection tasks doesn't return a result object. In general it probably makes sense that a task would return a status/result.

However it would also be nice to be able to write simple tasks which should always end in a success where you don't care about the result object. So either just continue if the result doesn't have the skipped attribute, or enforce the use of the Result object.

brigade/core/__init__.py
dbarrosop: I don't think we should do that for tracebacks explicitly, we already have a mechanism for that in the `AggregatedResult` and I think we need a more generic way of handling this. See #63

brigade/core/task.py
dbarrosop: Right now we are assuming `skipped` can only be because a task failed so we took it from the pool. There might be other reasons in the future so we could have something like:

```
class Skipped(object):
    def __init__(self, reason=None):
        self.reason = reason  # probably a set of standard reasons?
```

and then use it as:

```
return Result(skipped=Skipped("errored"))
```

so the user can do:

```
if result.skipped.reason == "error":
    # oh well
```

dbarrosop:
> Looks like the tests are failing because the xyz_connection tasks doesn't return a result object. In general it probably makes sense that a task would return a status/result.

Yeah, will fix.

> However it would also be nice to be able to write simple tasks which should always end in a success where you don't care about the result object

Well, tasks are supposed to do something so they should report something, even if it's an empty result just to indicate everything went fine. Will think about this.

PR #64, opened 2017-12-19
Expand netmiko connection task to add additional arguments from inventory

ktbyers:


ktbyers:
Test failure is due to earlier doc commit that caused issue with the tests.

ktbyers:
Sometimes Git is very annoying...closing to fix all of unnecessary changes.

PR #67, opened 2017-12-26
Add Netmiko task for sending configuration changes

ktbyers:


ktbyers:
We could also potentially add in a way of doing a diff in some way. 

For example, pass in a callable/task (in some way) and then some sort of a string comparison between what already exists via the callable (like execute `show run` or `show run | inc whatever`) and the config commands you are passing in.

dbarrosop:
Sorry for the late review on this one. I have been too self-caught on that other massive PR... :(

> For example, pass in a show command that could be executed by Netmiko diff_cmd='show run' and then a string compare of the config changes being sent in.

Not sure I understand this.

> Also there is no write mem happening here...so could have a post config change command that gets passed in as an argument.

Yeah, I think a post hook for that purpose might be interesting.

ktbyers:
@dbarrosop The first comment was basically saying what I have in this PR simply executes config changes...there is no check of whether the configuration items being sent already exist in the config. So I was proposing a possibly way that the config could be checked beforehand (i.e. the config changes are only executed if this is True).

If we did this, I would want to make it generic enough such that it could work across a wide set of Netmiko platforms (basically end-user would need to provide a command to execute on the remote pattern and same regex pattern that the output would match).

Does that make sense?

dbarrosop:
> and if we did this we probably start re-creating that diff/tree-parser you started creating earlier in NAPALM

Yeah, I agree. If we want to do that it's probably better solved somewhere outside (netmiko/napalm/other).

ogenstad:
It would be nice to be able to include this in the tests. Can we mock something on the netmiko side, or use the Linux machine?

ktbyers:
I think we can probably test it using the Linux machine...let me try to add tests.

dbarrosop:
Yeah, tests would be great! :D

ktbyers:
Will close, git rebase got screwed up.

Issue #73, opened 2018-01-20
Alternative to `easy_brigade`


dbarrosop:
As discussed here: https://github.com/napalm-automation/brigade/pull/68#discussion_r161208081

Might be better to handle this via the Configuration object instead.

PR #77, opened 2018-01-21
generic mechanism to create brigade object from configuration

dbarrosop:
Basically removes `easy_brigade` in favor of a more generic mechanism that works for both loading from a configuration file or from variables for a more programmatic access (tests show both ways, which can be combined, btw)

Solves #73


dbarrosop:
@ktbyers @ogenstad any opinions about this one? Basically let's you easily and consistently get a brigade object fully instantiated and configured via a configuration file or programmatically. Actually, even both at the same time. See tests for examples.

P.S: I know documentation is missing, I was planning to work on the documentation from scratch after the couple of issues/PRs we have pending for the stable API were merged.

ogenstad:
I'm not really convinced that `from_conf` would make things easier for people. I think it's mostly due to the way the inventory is defined.

Is it easier for people to use this when you have to send in the full path of the inventory object?

I guess my question is do we really need a `from_conf` or `easy_brigade`?

dbarrosop:
> I think it's mostly due to the way the inventory is defined.

Not sure I understand what you mean with this.

> Is it easier for people to use this when you have to send in the full path of the inventory object?

Well, they will do whatever the documention tells them to do.

> I guess my question is do we really need a from_conf or easy_brigade?

Have you seen the average ansible user? :P Jokes aside, I have no particular opinion about this. If we decide we don't want them I am fine with that.

ogenstad:
>Not sure I understand what you mean with this.

I mean that you would still have to point to `brigade.plugins.inventory.simple.SimpleInventory` and then the files for the inventory. I think this would basically be a repetition of just
`from brigade.plugins.inventory.simple import SimpleInventory`

>Well, they will do whatever the documention tells them to do.

But in that case should the documentation present two ways of doing this or just the `from_conf` way?

dbarrosop:
I think we should do like flask and have an "easy" way for people scared of python and also document the "advanced" usage for people that wants to understand the API and have full control over it. So the basic tutorial should cover `from_conf` and then have an advanced tutorial in the how to section.

The real question is if `from_conf` really provides a simpler entrypoint. I agree with you it's not a huge simplification but I think it's a big psychological improvements. Just try to explain how to create a `brigade` object with only the API and with `from_conf` to someone that doesn't know python and you will see what I mean :P

ogenstad:
What's the easy way with Flask? :)

How about making "inventory=brigade.plugins.inventory.simple.SimpleInventory" default?

And perhaps `inventory_settings` an alias for `SimpleInventory`?

Would `get_brigade` or `make_brigade` be a better name than `from_conf`?

>Just try to explain how to create a brigade object with only the API and with from_conf to someone that doesn't know python and you will see what I mean :P

No comments..

ktbyers:
Disclaimer...I need to look at this more.

Initial reaction is I think it would be simpler/better to use default arguments in the Brigade Class.


To me I would ratherhave:

Simple case / default case
```
brg = Brigade()

# Where inventory and configuration file are in assumed paths.
# You could override defaults in the configuration file
```

Advanced case (whatever the arguments that are allowed for behavior changes)
```
brg = Brigade(num_workers=100,
                        config_file=...,
                         inventory=...,
                         SimpleInventory=...)
```

ktbyers:
Separate item which I will try to not bring up again after today (in the context of Brigade):

I also think it would be better to abstract away the internal directory structure from library consumers (i.e. integrate the library through the package __init__.py). 

For example (current):
```
from brigade.core import Brigade
from brigade.plugins.inventory.simple import SimpleInventory
from brigade.plugins.tasks.connections import netmiko_connection
from brigade.plugins.tasks.networking import netmiko_send_config
```

Put negatively this creates a situation where the API is very prone to breaking people's code (i.e. any relocation of things internally breaks existing code). 

It also requires end-users to have knowledge of a large number of defaults where the defaults is internal directory structure.

I think this is better (for library consumers):


Basically really start reducing the number of defaults people have to remember/copy (in the form of directory structure in the API).

dbarrosop:
@ktbyers regarding the defaults, I had mixed feelings about it but let me give it an extra thought. That might be just the right solution. Regarding the "imports" I actually thought about that and the reason I discarded it was to avoid making all dependencies mandatory. However, I might be suffering of premature optimization so I am going to discard that thought of mine and propose the following import paths:

```
brigade.core (Brigade, Tasks, Results, everything core hangs from here)
brigade.plugins.tasks (all tasks hang from here)
brigade.plugins.inventory (all inventory plugins hang from here)
brigade.plugins.functions (all functions hang from here)
```

ktbyers:
@dbarrosop Yes, let me look at it more too (the defaults in the class versus note). Hopefully, I can see if I can do some tests this weekend.

ktbyers:
@dbarrosop Yes dependencies are a fair issue.

As you pointed out, I am not sure it is worth worrying about yet (since NAPALM requires Netmiko and anyone doing anything with Brigade would probably want one of those two).

I have wondered whether we want some way of dynamically finding/loading dependencies. But this is probably something we can worry about later (if we have tasks being added/used in some way, but we don't want their dependencies directly in Brigade).

ogenstad:
>Regarding the "imports" I actually thought about that and the reason I discarded it was to avoid making all dependencies mandatory.

Perhaps this doesn't have to be an issue either, we could have non default dependencies in a try/except statement and perhaps even collect a list of the failed imports. Would make the plugins a bit more complex though.

```
brg = Brigade()

# Where inventory and configuration file have certain known defaults
# You could override other defaults in the configuration file 
# or through environment vars
```

I think I prefer this to using `from_conf`, it's reasonable to just assume that SimpleInventory can be the default one, then we can have `brigade_hosts.yaml` and an optional `brigade_groups.yaml` in the current path (or through environment variables)

dbarrosop:
Re imports, let's just worry about that later. I will change them to meet the ones I described before and we can look into doing the imports dynamically or with `try..except..` later on.

Regarding the defaults, the only problem with assuming the that `SimpleInventory` is the default one is that now we are tying the core API to a "plugin" potentially causing a circular import. A compromise might be:

```
from brigade.core import Brigade
from brigade.plugins.inventory import SimpleInventory

brigade = Brigade(SimpleInventory())
```

This method might work well as we discussed writing an AnisbleInventory plugin that feeds off the same source as ansible for reusability (which I expect to be more popular than the SimpleInventory) so in that case you would do:

```
from brigade.core import Brigade
from brigade.plugins.inventory import AnisbleInventory

brigade = Brigade(AnisbleInventory())
```

ogenstad:
```
from brigade.core import Brigade
from brigade.plugins.inventory import SimpleInventory

brigade = Brigade(SimpleInventory())
```
Something like that would be fine as far as I'm concerned. Or perhaps:

```
brigade = Brigade(inventory=SimpleInventory())
```

ktbyers:
@dbarrosop My vote would be to just move SimpleInventory into core and out of plugins. I think the vast majority of people are going to use SimpleInventory. 

We can then have the lesser used versions of inventory in plugins (for the more advanced users).

ogenstad:
>My vote would be to just move SimpleInventory into core and out of plugins. I think the vast majority of people are going to use SimpleInventory.

I'd rather keep the inventory as a plugin and have the two imports, or do something creative with the import orders to avoid the circular import issue.

ktbyers:
There are no circular imports here. Here is what SimpleInventory imports:

```
from builtins import super

from brigade.core.inventory import Inventory

import yaml
```

We can define SimpleInventory in `brigade.core.inventory`. It is a non-issue for SimpleInventory.

dbarrosop:
I don't think I like the idea of putting `SimpleInventory` in the core. I like having clear separations. One of the things I hate the most in certain other tool is how everything is intertwined without clear separations. I understand it is not the same but I'd rather compromise with the defaults and have two imports. Also, I have my doubts that `SimpleInventory` is going to be the de facto standard. I am sure it will be the Ansible one as most people will prefer to reuse it if they already have ansible playbooks.

ktbyers:
So if I understand your argument, you generally don't want any imports from plugins in Brigade core and you view SimpleInventory as belonging in plugins since it is a particular implementation/loading of the Inventory class? It looks like our only current cross-coupling is the 'connections' plugin.

Having a default Inventory implementation that is in core that is automatically used (yet can be overridden) is a good way to overcome the issue of increasing coupling versus unnecessary code duplication (i.e. adding additional Brigade() entry points) versus making it harder for end-users (forcing the extra imports and default specification on them).

I just don't like the extra complexity we are forcing on end-users. 

So my order of preference is:

1. Make SimpleInventory be the default use case for Brigade() i.e. so you can just do:

```
brg = Brigade()
```

2. Have a second entry point like EasyBrigade() (basically your previous easy_brigade idea).

```
brg = EasyBrigade()
```

Which then returns an instance of the Brigade class implemented with SimpleInventory, using default file locations, and a default config file location.

I would be fine with this. This also seems like a reasonable compromise between the coupling and simplicity issue.

3. Forcing the defaults and imports into end-user's code. I definitely wouldn't like this.

dbarrosop:
> you view SimpleInventory as belonging in plugins since it is a particular implementation/loading of the Inventory class

Correct, I think that **anything** that can have an alternative implementation should be under plugins.

> It looks like our only current cross-coupling is the 'connections' plugin

Yeah, partially. And I say partially because we have a mechanism that will allow to fully decouple that part (or rather to extend; https://github.com/napalm-automation/brigade/blob/develop/brigade/core/__init__.py#L61)  as well but haven't bothered much because what I wanted to do in the long run is implement something like this for tasks and plugins:

https://packaging.python.org/guides/creating-and-discovering-plugins/

Ok, so regarding the second option. I can totally rename `from_conf` to `EasyBrigade` or something else. I will also make sure it has sane defaults so a configuration file is not strictly necessary.

ktbyers:
Are there actions still to be done here? I am not sure where we left this at?

dbarrosop:
Yeah, this one is on me. I want to rename `from_conf` to `EasyBrigade` and propose some sane defaults. I guess there will be some arguing about what "sane" means but luckily we are three so it's easy to vote on them :P

dbarrosop:
@ktbyers @ogenstad 

Ok, I think we are getting to a point where this is looking good:

1. I added this line to `CONF` https://github.com/napalm-automation/brigade/pull/77/commits/0b8ffbeb197350997665873d01582b0e60b0781d#diff-bf23cf236c6094bb080cb565a830f5ffR7 which basically dictates what to use for defaults in `from_conf` (now renamed to `InitBrigade`).
2. Code for `InitBrigade` is the same as `from_conf`, however, after adding that line in `CONF` you can just do `InitBrigade()` and get a fully working `Brigade`: https://github.com/napalm-automation/brigade/pull/77/commits/bc10d99bd876afc99a184d0bc7ef97ad3967a27a#diff-1e15bf0d1e2f5eb9705651d6d8a9839cR11
3. Right now defaults for the Inventory are basically what the class dictates (https://github.com/napalm-automation/brigade/pull/77/commits/bc10d99bd876afc99a184d0bc7ef97ad3967a27a#diff-c8267f60e1e9e916929d767774cd6569R105). You can modify those via the configuration file as seen here: https://github.com/napalm-automation/brigade/blob/bc10d99bd876afc99a184d0bc7ef97ad3967a27a/tests/core/test_InitBrigade/a_config.yaml#L4 or even from the code as seen here: https://github.com/napalm-automation/brigade/pull/77/commits/bc10d99bd876afc99a184d0bc7ef97ad3967a27a#diff-1e15bf0d1e2f5eb9705651d6d8a9839cR32

So with very little code we actually get a very flexible entry point for Brigade (which means we actually managed to get a nice, flexible and decoupled core API, also, kudos to @ogenstad for the Config class that is "empowering" this `InitBrigade` function).

Thoughts?

ogenstad:
I think this looks good now! Like the name InitBrigade much better than the others we have seen.

My only issue is that I don't really like that SimpleInventory now sets `group_file="groups.yaml"`. For people just doing a simple test with just a few hosts they now need to override the group_file setting if they don't need any groups. However, I guess the alternative would be to make InitBrigade less generic and add the group_file argument from there. This isn't a problem for me, more something that I noticed. Overall I would say to merge it.

I think this section in InitBrigade will need some more documentation so people know how to use another inventory, but we can add that later.
```python
    module_path = ".".join(conf.inventory.split(".")[:-1])
    inv_class_name = conf.inventory.split(".")[-1]
    module = importlib.import_module(module_path)
    inv_class = getattr(module, inv_class_name)
```

>Conflicting files
>tests/inventory_data/nsot/nsot.sqlite3

This seems lika a recurring issue?

dbarrosop:
we can log a warning if the group file doesn't exist and keep going. As @ogenstad suggested, if a host tries to use a group that doesn't exist it will fail later on anyway.

Issue #71, opened 2018-01-20
Revisit logging


dbarrosop:
As discussed here: https://github.com/napalm-automation/brigade/pull/68#discussion_r160587483

PR #78, opened 2018-01-21
[WIP] Better mechanism to configure logging

dbarrosop:
Solves #71


brigade/core/configuration.py
```diff
@@ -24,6 +24,40 @@
         'default': os.path.join(os.path.expanduser("~"), ".ssh", "config"),
         'default_doc': '~/.ssh/config'
     },
+    'logging_dictConfig': {
```
ogenstad: While I realise dictConfig just refers to the Python object, I don't really like the camel case.

brigade/core/configuration.py
```diff
@@ -24,6 +24,40 @@
...
+        'description': 'Configuration dictionary schema supported by the logging subsystem. '
+                       'Overrides rest of logging_* parameters.',
+        'type': 'dict',
```
ogenstad: When there is a configuration object with the type of "dict" should we have some restriction so you aren't able to set that value using an environment variable?

brigade/core/configuration.py
dbarrosop: Yeah, we should. But I don't think it's super important so maybe we can create an issue and fix it later with calm.

brigade/core/configuration.py
dbarrosop: Yeah, we should. But I don't think it's super important so maybe we can create an issue and fix it later with calm.

brigade/core/configuration.py
dbarrosop: I don't either but I figured it was better to keep the original name. Can change if you have strong opinions ;)

ogenstad: Strong opinions about it might be to go a bit to far. :) Might just be more confusing to rename it just for the sake of renaming.

ogenstad:
Is this still a work in progress or are you done adding code?

dbarrosop:
Yeah, sorry. Just waiting for feedback.

ogenstad:
+1, looks good.

We can create an issue to remove the option to set environment variables for 'type': 'dict' and 'type': 'list' once this is merged since that doesn't have anything to do with logging.

PR #86, opened 2018-03-03
inventory now support defaults and inheriting from multiple groups

dbarrosop:
Ok, I think this is the last enhancement I want to do before reviewing the documentation and releasing the first stable version.

There are two changes introduced here:

1. Ability to associate multiple groups to a Group or Host
2. The introduction of "defaults"

# Multiple groups

Basically any group and host can have attached multiple groups to it instead of just one. The way resolution works is bottom up and left to right. For instance:

host(g1, g2)
g1(g3, g4)
g2(g5, g6)

In this case if we look for var X we will search in the following order:

host -> g1 -> g3 -> g4 -> g2 -> g5 -> g6

I decided to go with this method because it is very easy to reason about and as easy to implement.

# defaults

The introduction of defaults serves a "similar" purpose as the old "all" group. However, `defaults` is not really a group, it's a dictionary with default variables that is consulted if a var is not found in any of the parent groups. It's better to illustrate it with an example. Imagine the following:

host(g1, g2)
g1(g3, g4)
g2(all)
g3(all)
g4(all)

Now, if you search for a variable present in "all" your resolution order is going to be as follows:

host -> g1 -> g3 -> all

If you search for a variable present in "defaults" and no "all" group what you'd get is:

host -> g1 -> g3 -> g4 -> g2 -> defaults

So you'd traverse all the possibilities before using the default value.

P.S: I still have pending updating the docstrigs. I just wanted to get the discussion out before.


ktbyers:
The general proposal sounds reasonable for both items. I think for the groups hierarchy the key things is the `easy to reason about and easy to implement`.

The defaults sounds fine.

Is this what this would look like in practice at least for the `defaults` (from your changed files)? 

```
---
defaults:
    group: null
    domain: acme.com
bma-leaf:
    group: bma
bma-host:
    group: bma
bma:
cmh-leaf:
    group: cmh
cmh-host:
    group: cmh
cmh:
```

Why do we need the `group: null` in the defaults? I also assume the default behavior for Simple Inventory would be to create a blank `defaults` if none was provide.

In other words, this would be similarly valid (minus not specifying a default domain)

```
---
bma-leaf:
    group: bma
bma-host:
    group: bma
bma:
cmh-leaf:
    group: cmh
cmh-host:
    group: cmh
cmh:
```

dbarrosop:
> Is this what this would look like in practice at least for the defaults (from your changed files)?

Yes. Note though that I renamed group -> groups and that it's a list to acomodate multiple groups:

```
---
bma-leaf:
    groups: [bma]
bma-host:
    groups: [bma]
bma:
cmh-leaf:
    groups: [cmh]
cmh-host:
    groups: [cmh]
cmh:
```

> Why do we need the group: null in the defaults?

We don't. The one you are referring to is probably something I just half updated. Will fix docs regarding this once we agree on the PR. Here is a fully working and update groups file: https://github.com/napalm-automation/brigade/pull/86/files#diff-c8e57a8e6cc57f466e72bc0aafea8708

ogenstad:
I think this looks good. My only issue is that the groups key requires a list. Do we want this? I'm assuming this is just due to the fact that it hasn't been tested.

one.yaml
```yaml
---
nas-1:
    brigade_host: nas-1
    groups: sweden
```

Failure:
```python
  File "one.py", line 6, in <module>
    inventory=SimpleInventory("one.yml", "single_group.yml"),
  File "/Users/patrick/src/forks/brigade/brigade/plugins/inventory/simple.py", line 120, in __init__
    super().__init__(hosts, groups, defaults, **kwargs)
  File "/Users/patrick/src/forks/brigade/brigade/core/inventory.py", line 279, in __init__
    h.groups = self._resolve_groups(h.groups)
  File "/Users/patrick/src/forks/brigade/brigade/core/inventory.py", line 286, in _resolve_groups
    r = [self.groups[g] for g in groups]
  File "/Users/patrick/src/forks/brigade/brigade/core/inventory.py", line 286, in <listcomp>
    r = [self.groups[g] for g in groups]
KeyError: 's'
```

Based on the earlier code it looks like you intended this to be allowed? Other that issue I think this PR is a good addition.

ktbyers:
Looks good to me.   

I agree on @ogenstad point that we should allow a single element `groups` to be a string and not a list.

dbarrosop:
> Based on the earlier code it looks like you intended this to be allowed? Other that issue I think this PR is a good addition.

No, I never intended that to work and I am not a fan of allowing things to have multiple types. It just adds complexity for no real benefit other than accommodating mistakes. But i can make it work if you have strong opinions about it.

dbarrosop:
Yeah, I think it goes against the ducktyping principle. So unless you or kirk has strong opinions about it I will leave it as it is. As you pointed out it is easy to fix later on anyway.

dbarrosop:
Ok, just updated the docstrings so I will merge. If there is anything else it can be changed in small PRs.

PR #88, opened 2018-03-09
Netmiko file_transfer task

ktbyers:
Secure Copy task support.


ktbyers:
Here is an example:

```python
from brigade.core import Brigade
from brigade.plugins.inventory.simple import SimpleInventory
from brigade.plugins.tasks.networking import netmiko_file_transfer


def result_printer(result):
    for host, result_obj in result.items():
        print('-' * 80)
        print("Result: {}".format(result_obj[0].result))
        print("Changed: {}".format(result_obj[0].changed))
        print('-' * 80)

brigade = Brigade(inventory=SimpleInventory("hosts.yml", "groups.yml"), dry_run=False)
b = brigade.filter(group='cisco-ios')

result = b.run(netmiko_file_transfer, num_workers=60,
               source_file="testb.txt", dest_file="test100.txt",
               direction='put', overwrite_file=True)
result_printer(result)

result = b.run(netmiko_file_transfer, num_workers=60,
               source_file="test100.txt", dest_file="test_new.txt",
               direction='get', overwrite_file=True)
result_printer(result)
```

Here is the output from that:

```
python test_netmiko_file_transfer.py 
--------------------------------------------------------------------------------
Result: True
Changed: True
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Result: True
Changed: True
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Result: True
Changed: False
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Result: True
Changed: True
--------------------------------------------------------------------------------
```

The one changed being False is correct since the get is occurring from two remote routers. The file is right for one of them.

brigade/plugins/tasks/networking/netmiko_file_transfer.py
```diff
@@ -0,0 +1,28 @@
...
+    Execute Netmiko file_transfer method
+
+    Arguments:
+        source_file(str): Source file.
+        dest_file(str): Destination file.
+        kwargs (dict, optional): Additional arguments to pass to file_transfer
+
+    Returns:
+        :obj:`brigade.core.task.Result`:
+          * result (``bool``): file exists and MD5 is valid
```
dbarrosop: I am assuming you are using this to signal the transfer was correct. Shouldn't we return instead bytes written or nothing at all and signal with "failed" if there was some error?

https://github.com/napalm-automation/brigade/blob/develop/brigade/core/task.py#L91

ktbyers: It will raise an exception if it tries to transfer and fails. 

'changed' flag will indicate whether the file was transferred or not. It won't transfer the file if the MD5 is already right.

I am using 'result' to indicate the destination file exists and and has the right MD5 (i.e. the file is right).

On your statement what if the file is already right, so the bytes written are 0? This would be confusing to me.

Here is what gets returned from the Netmiko function:

```
{'file_exists': True, 'file_transferred': True, 'file_verified': True}
```

file_exists will always be True (or an exception would be raised)
file_transferred I am mapping to 'changed' in brigade
file_verified indicates whether the MD5 was used for verification or not (since this can be turned off).

ktbyers:
This is updated now so that it is using a released version of Netmiko (Netmiko 2.1.1).

dbarrosop:
Nice! Merging.

PR #91, opened 2018-03-19
support more nsot parameters and auth_methods

smnmtzgr:
I have added support for providing nsot parameters in config.yaml (and load it with InitBrigade). I also added support for additional nsot parameters, so that both auth_methods (auth_header/auth_token) can be used.

### DIFF
--- a/brigade/plugins/inventory/nsot.py
+++ b/brigade/plugins/inventory/nsot.py
@@ -25,7 +25,7 @@ class NSOTInventory(Inventory):
             for filtering hosts.
         nsot_url (string): URL to nsot's API (defaults to ``http://localhost:8990/api``)
         nsot_email (string): email for authtication (defaults to admin@acme.com)
-        nsot_auth_header (string): String for auth_header authentication
+        nsot_auth_header (string): String for auth_header authentication (defaults to X-NSoT-Email)
         nsot_secret_key (string): Secret Key for auth_token method. If given auth_token
             will be used as auth_metod.
     """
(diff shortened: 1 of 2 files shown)


brigade/plugins/inventory/nsot.py
```diff
@@ -15,22 +15,38 @@ class NSOTInventory(Inventory):
         the name of the site the host belongs to.
 
     Environment Variables:
-        * ``NSOT_URL``: URL to nsot's API (defaults to ``http://localhost:8990/api``)
```
dbarrosop: environment variables are usually uppercase. What I'd do here is move everything you have here under env variables to Arguments (specify the type) and then under env variables add:

```
NSOT_URL: Corresponds to nsot_url argument
NSOT_EMAIL: Corresponds to nsot_email argument
...
```

brigade/plugins/inventory/nsot.py
```diff
@@ -15,22 +15,38 @@ class NSOTInventory(Inventory):
...
-        * ``NSOT_EMAIL``: email for authentication (defaults to admin@acme.com)
+        * ``nsot_url``: URL to nsot's API (defaults to ``http://localhost:8990/api``)
+        * ``nsot_email``: email for authentication (defaults to admin@acme.com)
+        * ``nsot_auth_method``: auth_header or auth_token
```
dbarrosop: wouldn't it be better to infer this? So if `nsot_secret_key` is specified use one method, otherwise use the other? Happy to leave it as it is, just suggesting a simplification

smnmtzgr: So you mean just remove the nsot_auth_method parameter and rely only on nsot_secret_key is given or not?

brigade/plugins/inventory/nsot.py
smnmtzgr: You are correct, I will correct this!

PR #92, opened 2018-03-22
Transform function can be given as function arg (InitBrigade) or in config.yaml

smnmtzgr:
Hi,
as discussed I implemented the possibility to define a transform_function for an inventory (e.g. nsot) either as function arg of InitBrigade or directly in the config.yaml. 

Example of giving it in the config.yaml is:
transform_function: simon.test.transform_function.adapt_host_data

-> simon.test.transform_function has to be a module which could be imported. adapt_host_data() is the function.

### DIFF
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -237,24 +237,10 @@ def InitBrigade(config_file="", dry_run=False, **kwargs):
     """
     conf = Config(config_file=config_file, **kwargs)
 
-    module_path = ".".join(conf.inventory.split(".")[:-1])
-    inv_class_name = conf.inventory.split(".")[-1]
-    module = importlib.import_module(module_path)
-    inv_class = getattr(module, inv_class_name)
-    inv_class_args = getattr(conf, inv_class_name, {})
-
-    transform_function = kwargs.get("transform_function")
-
-    if callable(transform_function):
-        inv_class_args['transform_function'] = transform_function
-    else:
-        module_path = ".".join(conf.transform_function.split(".")[:-1])
-        transform_function_name = conf.transform_function.split(".")[-1]
-        module = importlib.import_module(module_path)
-        transform_function = getattr(module, transform_function_name)
-        inv_class_args['transform_function'] = transform_function
-
-    inv = inv_class(**inv_class_args)
+    inv_class = conf.inventory
+    inv_args = getattr(conf, inv_class.__name__, {})
+    transform_function = conf.transform_function
+    inv = inv_class(transform_function=transform_function, **inv_args)
 
     return Brigade(
         inventory=inv,
--- a/brigade/core/__init__.py
+++ b/brigade/core/__init__.py
@@ -242,7 +242,7 @@ def InitBrigade(config_file="", dry_run=False, **kwargs):
     module = importlib.import_module(module_path)
     inv_class = getattr(module, inv_class_name)
 
-    inv = inv_class(**getattr(conf, inv_class_name, {}), **getattr(conf, 'transform_function', {}))
+    inv = inv_class(**getattr(conf, inv_class_name, {}), transform_function: conf.transform_function)
 
     return Brigade(
         inventory=inv,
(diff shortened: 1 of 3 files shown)


brigade/core/__init__.py
```diff
@@ -241,8 +241,20 @@ def InitBrigade(config_file="", dry_run=False, **kwargs):
     inv_class_name = conf.inventory.split(".")[-1]
     module = importlib.import_module(module_path)
     inv_class = getattr(module, inv_class_name)
+    inv_class_args = getattr(conf, inv_class_name, {})
```
dbarrosop: ok, a priori looks good. I am going to suggest a few changes now that you are at it to make this functionality a bit cleaner.

1. First, I'd add function to the `config.py` to do most of this magic. Something like:

```
def resolve_import_from_string(import_path):
       if not import_path or callable(import_path):
            return import_path
       module_name = ".".join(import_path.split(".")[:-1])
       obj_name = import_path.split(".")[-1]
       module = importlib.import_module(module_name)
```
(approx. this may or may not work :P)

2. Then I'd move the inventory and transform_function resolution to `Config.__init__`. Something like:

```
inventory = resolve_import_from_string(kwargs.get("inventory"))
kwargs["inventory"] = inventory

transform_function = resolve_import_from_string(kwargs.get("transform_function"))
kwargs["transform_function"] = transform_function
``` 

3. Finally, `InitBrigade` should be able to just do:

```
inv_class = conf.inventory
inv_args = getattr(conf, inv_class.__name__, {})
transform_function = conf.transform_function 
inv = inv_class(transform_function=transform_function, **inv_args)
``` 

I think this should make it cleaner and also make this solution more generic so anyone using the `Config` class could benefit of this regardless of if they are using `InitBrigade` or their own code to initialize all this.

smnmtzgr:
Thanks for your comments. I like your ideas to make this more generic!
Here is what I had to change at your code snippets to get it work:

in configuration.py
```
import importlib

        inventory = (self._resolve_import_from_string(kwargs.get("inventory")) or
                     self._resolve_import_from_string(self.inventory))
        self.inventory = inventory

        transform_function = (self._resolve_import_from_string(kwargs.get("transform_function")) or
                              self._resolve_import_from_string(self.transform_function))
        self.transform_function = transform_function
```
import is the "or" and that we assign the things to self.X and not to kwargs.

brigade/core/configuration.py
```diff
@@ -96,6 +102,14 @@ def __init__(self, config_file=None, **kwargs):
         for k, v in kwargs.items():
             setattr(self, k, v)
 
+        inventory = (self._resolve_import_from_string(kwargs.get("inventory")) or
```
dbarrosop: this can be rewritten as:

```
inventory = self._resolve_import_from_string(kwargs.get("inventory", self.inventory))
```

which is a bit cleaner

brigade/core/configuration.py
```diff
@@ -96,6 +102,14 @@ def __init__(self, config_file=None, **kwargs):
...
+                     self._resolve_import_from_string(self.inventory))
+        self.inventory = inventory
+
+        transform_function = (self._resolve_import_from_string(kwargs.get("transform_function")) or
```
dbarrosop: same as the inventory comment :)

dbarrosop:
It actually looks quite good. I'd suggest adding a  couple test here: https://github.com/napalm-automation/brigade/blob/develop/tests/core/test_InitBrigade.py

Something like:

```
def test_different_inventory_by_string(self):
        brg = InitBrigade(config_file=os.path.join(dir_path, "a_config.yaml"), inventory="path.to.NSOTInventory")
        assert isinstance(brg.inventory, NSOTInventory)

def test_different_inventory_imported(self):
        brg = InitBrigade(config_file=os.path.join(dir_path, "a_config.yaml"), inventory=NSOTInventory)
        assert isinstance(brg.inventory, NSOTInventory)
```

smnmtzgr:
Tests are implemented and are finally working:
- test_InitBrigade_different_inventory_by_string
- test_InitBrigade_different_inventory_imported
- test_InitBrigade_different_transform_function_by_string
- test_InitBrigade_different_transform_function_imported

Thanks for your help David! Something more we should change here?

dbarrosop:
Nice!

PR #96, opened 2018-03-30
added http_method task

dbarrosop:
Note that the way it's designed it can be used with both:

```
b.run(http_method,
          method="get",
          url="http://.../")
```

and standalone as

```
result = http_method(url="http://.../", method="get", ...)
```

### DIFF
--- a/brigade/plugins/tasks/apis/__init__.py
+++ b/brigade/plugins/tasks/apis/__init__.py
@@ -1,0 +1,5 @@
+from .http_method import http_method
+
+__all__ = (
+    "http_method",
+)
--- a/brigade/plugins/tasks/apis/http_method.py
+++ b/brigade/plugins/tasks/apis/http_method.py
@@ -13,7 +13,8 @@ def http_method(task=None, method="get", url="", raise_for_status=True, **kwargs
         url (string): URL to connect to
         raise_for_status (bool): Whether to call `raise_for_status
             <http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status>`_
-            method automatically or not
+            method automatically or not. For quick reference, raise_for_status will consider an
+            error if the return code is any of 4xx or 5xx
         **kwargs: Keyword arguments will be passed to the `request
             <http://docs.python-requests.org/en/master/api/#requests.request>`_
             method
--- a/brigade/plugins/tasks/apis/http_method.py
+++ b/brigade/plugins/tasks/apis/http_method.py
@@ -1,0 +1,38 @@
+from brigade.core.task import Result
+
+import requests
+
+
+def http_method(task=None, method="get", url="", raise_for_status=True, **kwargs):
+    """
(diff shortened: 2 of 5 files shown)


ogenstad:
I think the code for this looks good. The only thing which might be worth considering is this:

```python
result = r.json() if "application/json" == content_type else r.text
```
It might be that we should have `in` instead of `==`, the first example on http://docs.python-requests.org/en/master/ has:

```python
>>> r.headers['content-type']
'application/json; charset=utf8'
```
It could also be a list.

Other than that I think one thing to consider could be the name of the task/function, since it's so closely tied to requests that could be part of the name. Would also fit with the other ones like "napalm" and "netmiko". As we even refer to the requests documentation I think it could make sense to call it `http_requests` (even if `http_request` sounds better).

While I like the idea of keeping it simple as it is it could be helpful to introduce more named arguments so that people won't have to check documentation in an extra place.

I don't care much for the name `raise_for_status`, however I see the point of keeping the original name from requests. I think the documentation for that can be improved though. Even the one in requests isn't that good. I had to check in the requests source to get a complete picture of what it does. Might be confusing to have the argument be `raise_on_error`..

```python
return Result(host=task.host if task else None, response=r, result=result)
```

Another thought I had was the return value, for methods such as post or delete it might make more sense to have the status code in the result key and instead use a content or payload key for the returned data. I.e. something like this, but switch the name of the result variable to avoid confusion:

 ```python
return Result(host=task.host if task else None, content=result, result=r)
```

Even though this was a long comment I think it looks good and while I think it should look a bit different when checking the content-type the rest are just suggestions.

dbarrosop:
> I think it could make sense to call it http_requests (even if http_request sounds better)

I named it originally `http_requests` but backed out because it felt misleading, as if you could make multiple requests which is not the case.

> `raise_on_error`

Problem is that `raise_on_error` already has meaning in brigade: http://brigade.readthedocs.io/en/latest/ref/configuration/index.html#raise-on-error

I think I will document it a bit better.

> but switch the name of the result variable to avoid confusion

The main argument for keeping as it is is that "result" is the standard attribute all tasks use for returning "useful" information. This is interesting for generic tasks/workflows like `print_result`, for instance. If we had the response object in the `result` attributed `print_result` (or any other task) would get the request object (instead of a dict or text) and probably do something not very useful with it

ogenstad:
>Problem is that raise_on_error already has meaning in brigade: 

Hence my comment about that it could be confusing, perhaps `raise_on_http_error`, but that's a bit long. :) Anyway, I think the answer is just better documentation.

>The main argument for keeping as it is is that "result" is the standard attribute all tasks use for returning "useful" information.

What I meant was that for an http post the useful result might be a 201 reply or a status code, but now the `result` might be `None` which could indicate success. For the majority of cases, I think it's good as it is now. But I think either way is workable. Since the response object is returned it's easy to check anyway.

PR #97, opened 2018-03-31
[for discussion] add task wrapper and resolve_args functionality

dbarrosop:
Two things I'd like to discuss here:

1. Adding an optional wrapper where we can put some common logic for tasks. This common logic could have things like resolving arguments. For instance: https://github.com/brigade-automation/brigade/compare/task_wrapper?expand=1#diff-dbc265077515c480e1d473ace71d98b4R9
2. Do we want to resolve arguments automatically like we do in some places? I think @ktbyers doesn't really like it and I am starting to agree with him and we should probably remove it. The reason why I don't like it it's because it's not consistent between tasks and subtasks. for instance:

```
brigade.run(commands.command,
                    command="echo {name}") 
```

while most likely in a subtask you'd end up doing something like:

```
def my_subtask(task):
    task.run(commands.command,
             command=f"echo {task.host.name}") 
```

or:

```
def my_subtask(task):
    task.run(commands.command,
             command="echo {host.name}".format(host=task.host)) 
```

In any case, I think I'd rather encourage people to group tasks instead of running tasks in the main thread so I vote to remove the automatic resolver. In which case the wrapper can be scrapped or decide if we want to use it for something else. For instance, to "capture" non-`Result` objects and wrap them for consistency. For instance:

```
def my_subtask(task):
    ...
    return "YASSSSS!"


result = b.run(my_subtask)
print(result[my_host][0].result)  # this sould print "YASSSS!"
```

### DIFF
--- a/brigade/core/helpers/__init__.py
+++ b/brigade/core/helpers/__init__.py
@@ -7,6 +7,13 @@ def merge_two_dicts(x, y):
     return z
 
 
-def format_string(text, task, **kwargs):
-    return text.format(host=task.host,
-                       **merge_two_dicts(task.host.items(), kwargs))
+def format_object(obj, host, **kwargs):
+    if isinstance(obj, dict):
+        return {format_object(k, host, **kwargs): format_object(v, host, **kwargs)
+                for k, v in obj.items()}
+    elif any([isinstance(obj, t) for t in [list, set]]):
+        return [format_object(e, host, **kwargs) for e in obj]
+    elif isinstance(obj, str):
+        return obj.format(**merge_two_dicts(host.items(), kwargs))
+    else:
+        return obj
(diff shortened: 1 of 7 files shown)


dbarrosop:
@ktbyers @ogenstad

ogenstad:
>In any case, I think I'd rather encourage people to group tasks instead of running tasks in the main thread so I vote to remove the automatic resolver. 

Haven't looked at the code for this, but I agree about the grouping of tasks, that will probably be the most common way of running things. Might just add confusion if the documentation shows things which only works in certain scenarios. 

I haven't used the resolver, so I don't know if I like that feature or not. :)

dbarrosop:
> I haven't used the resolver, so I don't know if I like that feature or not. :)

will remove then

PR #99, opened 2018-04-01
Change repo links to brigade-automation

ogenstad:


dbarrosop:
Good catch, I didn't even think about this... Thanks! :D

PR #100, opened 2018-04-01
Move dry_run logic to the task

dbarrosop:
The purpose of this is two-fold:

1. Simplify dry_run logic by moving it to the task directly.
2. Make it clearer which tasks support "dry_run" or not

By removing the `dry_run` logic from the core and transferring it to the tasks we basically get rid of code which only purpose is to move things around and we also make it clearer which tasks support it as now to override argument is part of the task signature.


dbarrosop:
> It's not really related to this PR, but we were talking about merging functions and tasks

Yeah, I haven't gotten there.

PR #101, opened 2018-04-01
Added AnsibleInventory

dbarrosop:


ogenstad:
It looks like Ansible is moving to Yaml inventories, even though the INI option is still first in the documentation. Should this be named ansible_ini? Or if we add another one should that just be called ansible_yaml, or should this one be able to handle both options?

Might also be nice to be able to dump a Brigade inventory to files.

dbarrosop:
Would you mind testing?

brigade/plugins/inventory/ansible.py
```diff
@@ -0,0 +1,179 @@
+import configparser
```
ogenstad: Something weird is going on here. In Python 2.7 this module is called `ConfigParser` it was renamed to `configparser` in Python 3. I don't understand why this would work and how come Tox doesn't pick it up. Unless I'm missing something obvious there's a bug somewhere.

dbarrosop: yeah, that's weird...

brigade/plugins/inventory/ansible.py
```diff
@@ -0,0 +1,193 @@
...
+                result[k]["hosts"] = {
+                    host: self.normalize_content(data) for host, data in v.items()
+                }
+        return result
+
+    def load_hosts_file(self):
+        hostsfile = cp.ConfigParser(
+            interpolation=None, allow_no_value=True, delimiters=" "
+        )
+        hostsfile.read(os.path.join(self.path, "hosts"))
```
ogenstad: We shouldn't assume that the file is named `hosts`, it should be allowed to have any name.

brigade/plugins/inventory/ansible.py
```diff
@@ -0,0 +1,193 @@
...
+        data = self.normalize(hostsfile)
+        data.pop("DEFAULT")
+        if "all" not in data:
+            self.hostsfile = {"all": {"children": data}}
+
+
+class YAMLParser(AnsibleParser):
+
+    def load_hosts_file(self):
+        with open(os.path.join(self.path, "hosts"), "r") as f:
```
ogenstad: Same here with the name of the inventory.

ogenstad:
Looks good to me.

PR #103, opened 2018-04-02
Move test file out from root of repo

ogenstad:
The test file shouldn't be in the root of the repo.


tests/plugins/tasks/networking/test_netmiko_file_transfer.py
```diff
@@ -1,9 +1,13 @@
...
+
 from brigade.plugins.tasks import networking
 
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+
 
 class Test(object):
     def test_netmiko_file_transfer(self, brigade):
-        source_file = 'test_file.txt'
+        source_file = os.path.join(THIS_DIR, 'data/test_file.txt')
```
dbarrosop: you should probably do `os.path.join(THIS_DIR, 'data', 'test_file.txt')`. For consistency mostly :P (not that we are very good at it anyway :(

ogenstad: Sure. The contribution guidelines never mentioned this. ;)

PR #106, opened 2018-04-02
Contribution guidelines

ogenstad:


CONTRIBUTING.rst
```diff
@@ -0,0 +1,100 @@
+How to contribute to Brigade
+============================
+
+First of all, thank you for conidering to contribute to this project!
+
+Support questions
+-----------------
+
+While we are happy to help, the `GitHub issues <https://github.com/brigade-automation/brigade/issues>`_ are intended for bugs and discussions about new features. If are struggling to get something to work but don't believe its due to a bug in Brigade, the place to ask questions is in the #brigade channel in the `networktoCode Slack team <https://networktocode.herokuapp.com/>`_.
```
dbarrosop: maybe we need a mailing list for those that doesn't want to use slack. A google group maybe?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,100 @@
...
+If you have written your custom plugin for Brigade there's a good chance that it can be useful for others as well. General guidelines when writing plugins are:
+
+- Make them as generic as possible, it doesn't help others if they only work in your environment
+- Make sure that it's possible to have unit tests which automatically test that the plugins are working
+
+
+Contributing to the Brigade core
+--------------------------------
+
+[xyz]
```
dbarrosop: I would just add here to make sure that existing tests pass and to add tests for the new code to make sure future people working around that code doesn't break anything.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Coding style
+------------
+
+[xyz]
+
+Questions?
+----------
+
+[xyz]
```
dbarrosop: Slack, mailing list?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,100 @@
...
+Running tests
+-------------
+
+pip install tox
+tox
+
+Coding style
+------------
+
+[xyz]
```
dbarrosop: We could probably leave it as TBD and integrate with `black` later.

CONTRIBUTING.rst
ogenstad: Perhaps.. Not terribly keen on using something which generates more emails though and a semi-dead mailing list can give the wrong impression. But it might be good to have something else, or just say that if you want to ask questions Slack is where to find us. :)

ogenstad:
Was thinking we could soft link this into Sphinx too, do we want that or should I leave it as an independent file?

CONTRIBUTING.rst
dbarrosop: as discussed out of band; stackoverflow?

dbarrosop:
I think a link in RTD would be good

ogenstad:
Added a section in RTD and added some text. Removed the question section at the bottom. We can insert some info about Stackoverflow there and in the support section once we know what to write. Other than that I think this is done.

Perhaps when we know what that should be we can create the templates here on GitHub.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,117 @@
...
+
+	pip install tox
+	tox
+
+The test above will run the tests against the Brigade code. Aside from the code the documentation generated by `Spinx <http://www.sphinx-doc.org/>`_ is also validated. In order to test this you will be required to also install `Pandoc <https://pandoc.org/installing.html>`_. 
+
+Coding style
+------------
+
+TBD
```
dbarrosop: we should add here a reminder to run `ambv/black`

CONTRIBUTING.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Coding style
+------------
+
+Brigade uses `Black <https://github.com/ambv/black>`_, the the uncompromising Python code formatter. Black makes it easy for you to format your code as you can do so automatically after installing it. Note that Python 3.6 is required to run Black.
+
+.. code-block:: bash
+
+	pip install  black
+	black [source_code_directory]
```
dbarrosop: `black .` should do. Again, we can also hide this in the `make tests` or `make format`.

ogenstad:
Removed the rookie section and moved tests and formatting to the Makefile. Also added black in dev-requirements if the user is running Python 3.6.

CONTRIBUTING.rst
ktbyers: I would vote just using Slack.

dbarrosop: I think we should give an option for those who doesn't want to use slack. I am actually not a fan of slack myself. Stackoverflow has the advantage that questions/answers can be googgled so it serves as a KB. With slack you end up answering the same questions over and over.

ktbyers: Yes, I know you don't like Slack...you have a tendency to keep quiet on things you don't like, but I worked it out. :-)

Yes, we could try stack overflow; you can also use GitHub issues in a similar way.

I am not sure how much it will save us in answering the same-question over and over. It seems like to effectively avoid that you have to roll up the solution into some FAQ, that you can just remember/link to whenever anyone asks it. I say this somewhat from issues we keep getting asked over and over that are already answered in various places that are online.

I get worried about having two forums and then one of them basically doesn't get used (which frustrates people that ask questions there).

But am open to trying StackOverflow (and it decouples GitHub issues from FAQ-beginner questions).

dbarrosop: > you have a tendency to keep quiet on things you don't like, but I worked it out. :-)

http://s2.quickmeme.com/img/8a/8aef39928e5c5edcfe8328b798266de50eeae07fb36429d24ee483c3bb9deb70.jpg

I agree with everything you said but I think it's worth trying. Worst case we change our minds and remove that option from the README.

ktbyers: Sounds good...yes, was sarcasm ;-)

dbarrosop: I think we should move this section to the README and add stackoverflow.

PR #112, opened 2018-04-08
remove automatic resolver

dbarrosop:
It was used very inconsistently so let's remove it. If we want it later on we can try to add it via a decorator or something like that.

Solves #107

### DIFF
--- a/tests/plugins/tasks/commands/test_command.py
+++ b/tests/plugins/tasks/commands/test_command.py
@@ -2,13 +2,18 @@ from brigade.core.exceptions import CommandError
 from brigade.plugins.tasks import commands
 
 
+def echo_hostname(task):
+    command = "echo {host.name}".format(host=task.host)
+    task.run(task=commands.command, command=command)
+
+
 class Test(object):
 
     def test_command(self, brigade):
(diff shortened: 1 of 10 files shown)


tests/plugins/tasks/commands/test_command.py
```diff
@@ -2,13 +2,18 @@
...
+
 class Test(object):
 
     def test_command(self, brigade):
-        result = brigade.run(commands.command, command="echo {host.name}")
+        result = brigade.run(echo_hostname)
         assert result
         for h, r in result.items():
-            assert h == r.stdout.strip()
+            assert h == r[1].stdout.strip()
```
ktbyers: This seemed a bit strange...why the addition of the [1] i.e. why did it go from `r.stdout.strip()` to `r[1].stdout.strip().

ktbyers: Okay, I understand this behavior now.

dbarrosop: Just for reference; the reason is because instead of calling a task directly we are calling now a grouped task. Grouped tasks return a dict of results (key is host) with a list of results of which the first element is the result for the whole group.

Issue #94, opened 2018-03-27
easy way of adding jinja filters


dbarrosop:
easy way of adding jinja filters

PR #113, opened 2018-04-09
add mechanism to pass jinja_filters

dbarrosop:
For instance:

```
# jinja_filters/__init__.py

def get_filters():
    return {"test": test}


def test(data):
    return "yes!!!"
```

and then:

```
    brg = InitBrigade(
        dry_run=dry_run,
        SimpleInventory={"host_file": HOSTS_FILE, "group_file": GROUPS_FILE},
        jinja_filters="jinja_filters.get_filters",
    )
```


dbarrosop:
Solves #94

PR #115, opened 2018-04-10
add explanation about how versioning works

dbarrosop:
Once the API is stable I will release a 1.0.0 release. I was thinking in the following versioning scheme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,19 @@ While Brigade still supports Python 2.7 the recommended version is 3.6. Install 
 pip install brigade
 ```
 
+Versioning
+==========
+
+It is recommended to understand what the versioning of brigade tells you. The format is as follows; `x.y.z`, where:
+
+- `x` is the API version. This will rarely change and if it does it means it's a major release that breaks completely backwards compatibility.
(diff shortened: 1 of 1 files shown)


dbarrosop:
@ogenstad @ktbyers

README.md
```diff
@@ -19,6 +19,19 @@ While Brigade still supports Python 2.7 the recommended version is 3.6. Install
 pip install brigade
 ```
 
+Versioning
+==========
+
+It is recommended to understand what the versioning of brigade tells you. The format is as follows; `x.y.z`, where:
+
+- `x` is the API version. This will rarely change and if it does it means it's a major release that breaks completely backwards compatibility.
```
ogenstad: I think this would be worded a bit differently. The "breaks completely backwards compatibility" would be the worst case scenario. I would tone it down a bit. 

>Has a high probability of introducing backward incompatible changes, make sure you read the release notes.

Or something like that?

ktbyers: Here is the terminology `semantic versioning` uses:

https://semver.org/

```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.
```

I would probably use something similar to that for the MAJOR version change.

PR #119, opened 2018-04-16
Remove link to specific build on RTD

ogenstad:


ktbyers:
The unit test failing is pip10 issue.

PR #120, opened 2018-04-16
Add result key and cleanup

ogenstad:
Should we also add / change other plugins to always have the most important output in "result"?

Don't know if that would be diff or changed in something like napalm_configure().


dbarrosop:
I think result should be whatever is considered to be printable and most meaningful for the user. I am merging this one, we can evaluate the rest one by one if needed.

PR #121, opened 2018-04-16
Add description to package for display on pypi.

ogenstad:
With today's launch of the new pypi we can now have markdown files to describe the project. Currently Brigade doesn't have a description on https://pypi.org/project/brigade/, this will add the README.md to the package so that it can be displayed on the pypi page.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include requirements*
+include README.md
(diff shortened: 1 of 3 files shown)


ktbyers:
Looks good to me. We will just have to verify it the next time we do a release.

ktbyers:
+1

PR #124, opened 2018-04-22
dont mask error when running a grouped task

dbarrosop:
Without this a grouped task with a failing task won't stop and will keep running the rest of the tasks in the same group. For instance:

```
def grouped_tasks(task):
     task.run(task=this_task_fails)
     task.run(task=another_task)

brigade.run(task=grouped_tasks)
```

Without this patch `another_task` will run even if `this_task_fails`. With this patch, you have to catch if if you want to keep going.

### DIFF
--- a/brigade/core/task.py
+++ b/brigade/core/task.py
@@ -93,6 +93,10 @@ class Task(object):
         if "severity" not in kwargs:
             kwargs["severity"] = self.severity
         r = Task(task, **kwargs).start(self.host, self.brigade)
+        if r.failed:
+            # Without this we will keep running the grouped task
+            raise r.exception
+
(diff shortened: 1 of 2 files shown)


brigade/core/task.py
```diff
@@ -93,6 +93,10 @@ def grouped_tasks(task):
         if "severity" not in kwargs:
             kwargs["severity"] = self.severity
         r = Task(task, **kwargs).start(self.host, self.brigade)
+        if r.failed:
+            # Without this we will keep running the grouped task
+            raise r.exception
```
ogenstad: Hm, how does this work with the `raise_on_error` setting?

ogenstad:
Looks good to me.

PR #126, opened 2018-04-26
Change severity to severity_level on tasks, add tests to print_output

ogenstad:
Adding a test to print_result as discussed in https://github.com/brigade-automation/brg-tools/pull/1

Copied the wrapper from the same PR, but modified it slightly for Python 2 and also added an exception to make it easier to see why a task was failing as the `assert` seemed to be running outside of the test environment and only gave a true/false statement, it wasn't possible to see how the results differed.

I took a shortcut by only returning a bare string from a fake task. Other return values gets converted to unicode in Python 2 so the asserts fail. Could revisit that later.


ogenstad:
@dbarrosop, so the `decorator` package was installed before. However, it was installed as a dependency to the Sphinx setup. When you moved that to a different Tox environment this PR broke.

It's fixed now.

ogenstad:
Sure, was mostly thinking about making the badge prettier for the first release. :) 

But I can probably add some more tests tomorrow. 

It will be with separate output files for Python 2 though. As there will be slight differences in the output if we do anything more advanced than the basic test I did here.

dbarrosop:
haha let's do something useful now that we are at it :D

ogenstad:
Will also add the severity tests.

```python
    @wrap_cli_test(output="{}/basic_inventory_one_task".format(output_dir))
    def test_print_basic_inventory_one_host(self, brigade):
        result = brigade.run(data_with_greeting)
        print_result(result["dev2.group_1"][1])
```

With the above test, do we want the last task to be inserted first in the list?

ogenstad:
What a surprise, the paths in the traceback looks completely different on Travis compared to my system.

Might have to mock traceback.format_exc().

https://github.com/brigade-automation/brigade/blob/develop/brigade/core/task.py#L64-L67

Or do you have any other suggestions?

ogenstad:
The simplest solution is perhaps just to use the vars argument of print_result and not include the result property.

ogenstad:
https://github.com/brigade-automation/brigade/blob/develop/brigade/core/task.py#L96-L100

```python
        if r.failed:
            # Without this we will keep running the grouped task
            raise r.exception

        self.results.append(r[0] if len(r) == 1 else r)
```
Changing the order of these could solve the issue of seeing which task actually failed. However, the error is also caught by the grouped task so just changing the above would cause the error to appear twice.

dbarrosop:
> What a surprise, the paths in the traceback looks completely different on Travis compared to my system.

We can probably just check the exception you are looking for is there. Something like:

```
assert "blah.bleh.MyTerribleException" in stderr
```

dbarrosop:
> Changing the order of these could solve the issue of seeing which task actually failed. However, the error is also caught by the grouped task so just changing the above would cause the error to appear twice.

I think that's actually fine-ish. Maybe we should modify how the group task captures the exception and mark somehow that as a BrigadeSubtaskExeption or something like that to mark a subtask failed. I think the grouped task should have a real exception only if it was triggered by the code in it. For instance:

```
def grouped_task_a(task):
    with open("file_doesnt_exist", "r") as f:
    ...

def grouped_task_b(task):
    task.run(task=load_yaml, file="file_doesnt_exist)
    ...
```

* grouped_task_a - should have the OSerror exception as it was trying to open a non-existent file
* grouped_task_b - should have the OSerror on the result for the load_yaml task and have the `BrigadeSubtaskExeption` as the result for the grouped task

Does this make sense?

ogenstad:
>assert "blah.bleh.MyTerribleException" in stderr

I did a workaround so we don't have to change the wrapper:
https://github.com/brigade-automation/brigade/blob/92854761209b921f3dfaf01b1a5c0c02c236414f/tests/plugins/functions/text/test_print_result.py#L80-L83

I think that works too.


>Maybe we should modify how the group task captures the exception and mark somehow that as a BrigadeSubtaskExeption or something like that to mark a subtask failed.

That sounds good. Perhaps open an issue for that and handle it later? I think it should be fixed but it isn't urgent.

PR #128, opened 2018-04-29
add pylama job

dbarrosop:


brigade/core/__init__.py
```diff
@@ -119,7 +119,7 @@ def configure_logging(self):
             "handlers": {},
             "loggers": {},
             "root": {
-                "level": "CRITICAL" if self.config.logging_loggers else self.config.logging_level.upper(),
+                "level": "CRITICAL" if self.config.logging_loggers else self.config.logging_level.upper(),  # noqa
```
ktbyers: Why not something like this:

```
                "level": "CRITICAL" if self.config.logging_loggers 
                else self.config.logging_level.upper(),
```

dbarrosop: because we are using black to avoid having this sort of discussions :D The `noqa` is just so `pylama` doesn't complain. I am with you that this could look better as you suggested and I think they were discussing fixing it somehow but the version we are pinned to right now wants that format

ktbyers: Yes, totally agree on avoiding pointless linter discussions.

So black doesn't actually format it to comply with line length requirements for a ternary operator?

ktbyers: To me that is a black-formatting bug (i.e. it should auto format to meet the line length restriction)...of course, is easier said than done with all the possible formatting requirements.

dbarrosop: Yes, it is a bug and 18.4a3 fixes it. However, as the project is fairly new and they are still tweaking certain bits I pinned a particular version in the requirements to avoid having issues with different people having different versions. The plan is to pin it to 18.4 once released and update the code so it complies with that particular version.

Issue #129, opened 2018-04-29
Exception doesn't exist for self defined failures


ogenstad:
This code causes an error for self-inflicted failures.
https://github.com/brigade-automation/brigade/blob/develop/brigade/core/task.py#L96-L98

For example:
```python
def echo_task(task, msg):
    failed = False
    if msg == "5":
            failed = True
    return Result(host=task.host, result=f"Hello from Brigade: {msg}", failed=failed)


def data_with_greeting(task):
    task.run(task=echo_task, name="First", msg='1')
    task.run(task=echo_task, name="Second", msg='5')
    task.run(task=echo_task, name="Third", msg='3')
    task.run(task=echo_task, name="Fourth", msg='4', severity=logging.DEBUG)
    task.run(task=echo_task, name="Fifth", msg='5', severity=logging.DEBUG)

Traceback (most recent call last):
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 61, in start
    r = self.task(self, **self.params)
  File "test-run.py", line 32, in data_with_greeting
    task.run(task=echo_task, name="Second", msg='5')
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 98, in run
    raise r.exception
TypeError: exceptions must derive from BaseException
```
The Result.exception is set to None by default and doesn't get changed when a failed task doesn't throw an error.
https://github.com/brigade-automation/brigade/blob/develop/brigade/core/task.py#L58-L67

```python
        if r.failed:
            # Without this we will keep running the grouped task
            raise r.exception
```
Either we need to check if r.exception=None, and raise some other generic Exception. Or we can let users return a failure reason. 

Thoughts?

dbarrosop:
I'd say that shouldn't be allowed (maybe we need to rename the argument to `_failed` to make it clearer). If the task is unable to do what was supposed to do the task should raise an Exception.

dbarrosop:
no, I don't think it was. Maybe it surfaced the problem but I don't think tasks should return failed. Maybe we should revisit this and make sure certain fields that are supposed to be filled by brigade core itself (and not by a plugin) is properly marked as private with prepending `_`.

I will leave this issue open for that particular purpose :D

ogenstad:
Just to illustrate what I meant, if we have a file which looks like this:

```python
from brigade.core import InitBrigade
from brigade.plugins.functions.text import print_result
from brigade.core.task import Result


brg = InitBrigade(
    raise_on_error=False)


[...]

PR #131, opened 2018-05-02
Add BrigadeSubTaskError for grouped tasks

ogenstad:
Raises BrigadeSubTaskError when a subtask to a grouped task has failed,
this allows for a prettier error message when using print_output()

I haven't changed the output from the tests for print_output yet so currently this will fail.

With a test script:
```python
from brigade.core import InitBrigade
from brigade.plugins.functions.text import print_result
from brigade.core.task import Result


brg = InitBrigade(
    raise_on_error=False)


def greet(task, msg):
    return Result(host=task.host, result="{}".format(msg))


def fail_on_me(task):
    with open('does_not_exist.txt', 'r') as fs:
        data = fs.read()
    return Result(host=task.host, result=data)


def sub_task(task):
    task.run(task=greet, msg="first")
    task.run(task=greet, msg="second")
    task.run(task=fail_on_me)
    task.run(task=greet, msg="third")
    return Result(host=task.host, result="grouped_task")


result = brg.run(sub_task)
print_result(result)
```

Running the script with the current develop branch:
```bash
sub_task************************************************************************
* srv1 ** changed : False ******************************************************
vvvv sub_task ** changed : False vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv ERROR
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 61, in start
  File "will-fail.py", line 23, in sub_task
    task.run(task=fail_on_me)
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 98, in run
    raise r.exception
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 61, in start
  File "will-fail.py", line 15, in fail_on_me
    with open('does_not_exist.txt', 'r') as fs:
FileNotFoundError: [Errno 2] No such file or directory: 'does_not_exist.txt'

---- greet ** changed : False -------------------------------------------------- INFO
first
---- greet ** changed : False -------------------------------------------------- INFO
second
^^^^ END sub_task ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
* srv2 ** changed : False ******************************************************
vvvv sub_task ** changed : False vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv ERROR
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 61, in start
  File "will-fail.py", line 23, in sub_task
    task.run(task=fail_on_me)
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 98, in run
    raise r.exception
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 61, in start
  File "will-fail.py", line 15, in fail_on_me
    with open('does_not_exist.txt', 'r') as fs:
FileNotFoundError: [Errno 2] No such file or directory: 'does_not_exist.txt'

---- greet ** changed : False -------------------------------------------------- INFO
first
---- greet ** changed : False -------------------------------------------------- INFO
second
^^^^ END sub_task ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Running the script with this PR:
```bash
sub_task************************************************************************
* srv1 ** changed : False ******************************************************
vvvv sub_task ** changed : False vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv ERROR

########################################
# Subtask: fail_on_me (failed)
########################################

---- greet ** changed : False -------------------------------------------------- INFO
first
---- greet ** changed : False -------------------------------------------------- INFO
second
---- fail_on_me ** changed : False --------------------------------------------- ERROR
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 62, in start
  File "will-fail.py", line 15, in fail_on_me
    with open('does_not_exist.txt', 'r') as fs:
FileNotFoundError: [Errno 2] No such file or directory: 'does_not_exist.txt'

^^^^ END sub_task ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
* srv2 ** changed : False ******************************************************
vvvv sub_task ** changed : False vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv ERROR

########################################
# Subtask: fail_on_me (failed)
########################################

---- greet ** changed : False -------------------------------------------------- INFO
first
---- greet ** changed : False -------------------------------------------------- INFO
second
---- fail_on_me ** changed : False --------------------------------------------- ERROR
  File "/Users/patrick/src/brigade/brigade/core/task.py", line 62, in start
  File "will-fail.py", line 15, in fail_on_me
    with open('does_not_exist.txt', 'r') as fs:
FileNotFoundError: [Errno 2] No such file or directory: 'does_not_exist.txt'

^^^^ END sub_task ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

With something like this in place I think we can also ignore #129.

### DIFF
--- a/brigade/core/exceptions.py
+++ b/brigade/core/exceptions.py
@@ -71,8 +71,4 @@ class BrigadeSubTaskError(Exception):
         self.result = result
 
     def __str__(self):
-        text = "\n"
-        text += "{}\n".format("#" * 40)
-        text += "# Subtask: {} (failed)\n".format(self.task)
-        text += "{}\n".format("#" * 40)
-        return text
+        return "Subtask: {} (failed)\n".format(self.task)
--- a/brigade/core/exceptions.py
+++ b/brigade/core/exceptions.py
@@ -58,13 +58,16 @@ class BrigadeSubTaskError(Exception):
     has failed
 
     Arguments:
-        task (str): Name of the sub task that failed
+        task (:obj:`brigade.core.task.Task`): The subtask that failed
+        result (:obj:`brigade.core.task.Result`): The result of the failed task
     Attributes:
-        task (str): Name of the sub task that failed
+        task (:obj:`brigade.core.task.Task`): The subtask that failed
+        result (:obj:`brigade.core.task.Result`): The result of the failed task
     """
 
-    def __init__(self, task):
+    def __init__(self, task, result):
         self.task = task
+        self.result = result
 
     def __str__(self):
         text = "\n"
--- a/brigade/core/exceptions.py
+++ b/brigade/core/exceptions.py
@@ -52,6 +52,7 @@ class BrigadeExecutionError(Exception):
                 text += "{}\n".format(sub_r)
         return text
 
+
 class BrigadeSubTaskError(Exception):
     """
     Raised by brigade when a sub task managed by :meth:`brigade.core.Task.run`
--- a/brigade/core/exceptions.py
+++ b/brigade/core/exceptions.py
@@ -51,3 +51,24 @@ class BrigadeExecutionError(Exception):
                 text += "**** {}\n".format(sub_r.name)
                 text += "{}\n".format(sub_r)
         return text
+
+class BrigadeSubTaskError(Exception):
+    """
+    Raised by brigade when a sub task managed by :meth:`brigade.core.Task.run`
+    has failed
+
+    Arguments:
+        task (str): Name of the sub task that failed
+    Attributes:
+        task (str): Name of the sub task that failed
+    """
+
+    def __init__(self, task):
+        self.task = task
+
+    def __str__(self):
+        text = "\n"
+        text += "{}\n".format("#" * 40)
+        text += "# Subtask: {} (failed)\n".format(self.task)
+        text += "{}\n".format("#" * 40)
+        return text
--- a/brigade/core/task.py
+++ b/brigade/core/task.py
@@ -100,12 +100,13 @@ class Task(object):
 
         if "severity_level" not in kwargs:
             kwargs["severity_level"] = self.severity_level
-        r = Task(task, **kwargs).start(self.host, self.brigade)
+        task = Task(task, **kwargs)
+        r = task.start(self.host, self.brigade)
         self.results.append(r[0] if len(r) == 1 else r)
 
         if r.failed:
             # Without this we will keep running the grouped task
-            raise BrigadeSubTaskError((r.name))
+            raise BrigadeSubTaskError(task=task, result=r)
 
         return r
 
--- a/brigade/core/task.py
+++ b/brigade/core/task.py
@@ -3,6 +3,7 @@ import traceback
 from builtins import super
 
 from brigade.core.exceptions import BrigadeExecutionError
+from brigade.core.exceptions import BrigadeSubTaskError
 
 
 class Task(object):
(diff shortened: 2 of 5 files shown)


brigade/core/exceptions.py
```diff
@@ -51,3 +51,24 @@ def __str__(self):
...
                 text += "{}\n".format(sub_r)
         return text
+
+class BrigadeSubTaskError(Exception):
+    """
+    Raised by brigade when a sub task managed by :meth:`brigade.core.Task.run`
+    has failed
+
+    Arguments:
+        task (str): Name of the sub task that failed
```
dbarrosop: I think task should be a reference the task itself, not just the name. I think we should also have a reference to the result. So you'd basically call this like (see comment below)

brigade/core/task.py
```diff
@@ -93,11 +101,12 @@ def grouped_tasks(task):
         if "severity_level" not in kwargs:
             kwargs["severity_level"] = self.severity_level
         r = Task(task, **kwargs).start(self.host, self.brigade)
+        self.results.append(r[0] if len(r) == 1 else r)
+
         if r.failed:
             # Without this we will keep running the grouped task
-            raise r.exception
+            raise BrigadeSubTaskError((r.name))
```
dbarrosop: Given my [previous](https://github.com/brigade-automation/brigade/pull/131/files#r185722111) comment I think this should be:

```
raise BrigadeSubTaskError(task=task, error=r)
```

That will also mean we have to change:

```
r = Task(task, **kwargs).start(self.host, self.brigade)
```

to:

```
task = Task(task, **kwargs)
r = task.start(self.host, self.brigade)
```

brigade/core/task.py
```diff
@@ -61,10 +62,17 @@ def start(self, host, brigade):
             r = self.task(self, **self.params)
             if not isinstance(r, Result):
                 r = Result(host=host, result=r)
+
+        except BrigadeSubTaskError as e:
```
dbarrosop: I don't think we need this block, do we?

tests/plugins/functions/text/output_data/failed_with_severity.stdout
```diff
@@ -13,7 +13,9 @@ Hello from CRITICAL
 [1m[32m^^^^ END read_data ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 [1m[34m* dev4.group_2 ** changed : False **********************************************
 [1m[31mvvvv read_data ** changed : False vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv ERROR
-Exception('Unknown Error -> Contact your system administrator',)
+BrigadeSubTaskError()
```
dbarrosop: Shouldn't this have been something like `Subtask: parse_data (failed)\n`?

ogenstad: It's due to this:
https://github.com/brigade-automation/brigade/blob/develop/tests/plugins/functions/text/test_print_result.py#L101

Because the result key isn't printed. That's to avoid showing the traceback which contains my local paths.

brigade/core/task.py
ogenstad: There's a small difference as I don't save the traceback in the result key after BrigadeSubTaskError. Do you want to remove it?

tests/plugins/functions/text/output_data/failed_with_severity.stdout
dbarrosop: I think that should output the task name and the arguments used if possible. If it's too hard we can open an issue and fix it later

PR #137, opened 2018-05-09
NAPALM connection fallback to ssh_port if no optional_args and no API port defined

ktbyers:
For Cisco IOS and for NXOS_SSH, Brigade user's might logically define a `brigade_ssh_port` in inventory.

Consequently, I think it is logical that we should process them in this order:

1. optional_args['port']
2. brigade_network_api_port
3. brigade_ssh_port


dbarrosop:
We probably need a better way of documenting this but let’s merge this for now.

PR #145, opened 2018-05-21
Clarify which key is used for the Napalm port

ogenstad:
The current docstring is wrong and can cause confusion.

Another solution would be to point to `task.host.network_api_port` but before we have a specific section in the documentation which talks about how these are connected I think it's better to point to the key as in this PR.


ogenstad:
Removed two extra ` characters to make pylama happy. Not sure if they are needed for anything?

PR #146, opened 2018-05-22
add to_dict function so the inventory is serializable

dbarrosop:
To discuss, might be nice if we can dump with a `to_dict` an inventory object and make it load back into the same object. I think we should bother about doing it as a `dict` instead of as a `json` or any other format as each one of them have a different mechanism to deal with custom objects and we should assume they should work.

Basically what I'd like to do here is make sure that we can do something equivalent to `my_inv == Inventory(my_inv.to_dict())`

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -58,6 +58,10 @@ class Data(object):
         """Reset failed hosts and make all hosts available for future tasks."""
         self.failed_hosts = set()
 
+    def to_dict(self):
+        """ Return a dictionary representing the object. """
+        return self.__dict__
+
 
 class Nornir(object):
     """
@@ -248,6 +252,10 @@ class Nornir(object):
         else:
             self.data.failed_hosts.update(result.failed_hosts.keys())
         return result
+
+    def to_dict(self):
+        """ Return a dictionary representing the object. """
+        return {"data": self.data.to_dict(), "inventory": self.inventory.to_dict()}
 
 
 def InitNornir(config_file="", dry_run=False, **kwargs):
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -106,6 +106,10 @@ class Host(object):
         """
         return self._resolve_data().items()
 
+    def to_dict(self):
+        """ Return a dictionary representing the object. """
+        return self.data
+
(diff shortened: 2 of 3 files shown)


ogenstad:
>my_inv == Inventory(my_inv.to_dict())

This sounds good. One thing that is missing are the values from the `defaults` group.

Should we have anything to prevent people from shooting themselves in the foot?

```python
import json
from nornir.core import InitNornir
from netaddr import IPAddress


def set_ip(task):
    task.host["ip_address"] = IPAddress("10.0.0.1")


nr = InitNornir()
nr.run(set_ip)
my_inv = nr.inventory.to_dict()

print(json.dumps(my_inv))
````

```python
(nornir-dev) ~/src/playground/nornir ᐅ python test_serial.py
  File "test_serial.py", line 14, in <module>
    print(json.dumps(my_inv))
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'IPAddress' is not JSON serializable
(nornir-dev) ~/src/playground/nornir ᐅ
```

dbarrosop:
> Should we have anything to prevent people from shooting themselves in the foot?

If you have a suggestion for that, sure... Otherwise I'd say no.

dbarrosop:
Good catch on the missing `defaults` btw :)

Is there something else you think might need to be added?

ogenstad:
I don't have a solution was only thinking about what could come up.

No, I can't think of anything else to add to this.

PR #152, opened 2018-06-04
Implement django style queries

dbarrosop:
Alternative to #149 

So for fun I implemented this. Thoughts? This way the easiest way to implement this I could think of.

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -249,6 +249,18 @@ class Nornir(object):
             self.data.failed_hosts.update(result.failed_hosts.keys())
         return result
 
+    def __len__(self):
+        return self.inventory.__len__()
+
+    def __contains__(self, other):
+        return self.inventory.__contains__(other)
+
+    def __and__(self, other):
+        return self.inventory.__and__(other)
+
+    def __or__(self, other):
+        return self.inventory.__or__(other)
+
 
 def InitNornir(config_file="", dry_run=False, **kwargs):
     """
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -366,6 +366,29 @@ class Inventory(object):
     def __len__(self):
         return self.hosts.__len__()
 
+    def __contains__(self, hostname):
+        return hostname in self.hosts
+
+    def __and__(self, other):
+        return Inventory(
+            hosts={
+                name: host for name, host in self.hosts.items() if name in other.hosts
+            },
+            groups={
+                name: group
+                for name, group in self.groups.items()
+                if name in other.groups
+            },
+            nornir=self.nornir,
+        )
+
+    def __or__(self, other):
+        return Inventory(
+            hosts={**self.hosts, **other.hosts},
+            groups={**self.groups, **other.groups},
+            nornir=self.nornir
+        )
+
     @property
     def nornir(self):
         """Reference to the parent :obj:`nornir.core.Nornir` object"""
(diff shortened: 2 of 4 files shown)


dbarrosop:
@ktbyers @ogenstad

ogenstad:
For me, it was mostly the fact that if we say to people that this will be easy that these filters look different than more simple ones. Perhaps I should just have more faith in humanity. :)

Currently this works:
```python
routers = nr.filter(groups__contains="routers")
```
In #149, there was talk about also having something like this:

```python
routers = nr.filter(groups__not__contains="switches")
```

I think something along those lines would be good to have too.

dbarrosop:
> For me, it was mostly the fact that if we say to people that this will be easy that these filters look different than more simple ones. Perhaps I should just have more faith in humanity. :)

Well, the simple ones still work. This is just if you need it, if you don't nothing changes.

> `routers = nr.filter(groups__not__contains="switches")`

Makes sense, will look into it, the CI is also broken because py2.7 is not nearly as nice as py3 as well so I have to fix it... I can't wait to get rid of py2.7. Should we drop support? :P (not kidding)

ogenstad:
Yeah saw that the old stuff works which is nice.

Since we have a real version out that is installable with pip for Python 2.7 which works quite nicely I'm fine with dropping support for it. In this case it would complicate the code to keep it in which isn't very attractive.

ktbyers:
@dbarrosop Dropping PY27 support is fine with me.

ktbyers:
I will try to experiment some more with this shortly.

ogenstad:
Now I want a filter to include all hosts from groupA, groupB or groupC. Should we have something for that or just write a custom function?

dbarrosop:
to do that I was thinking in implement the `+` and `-` operation so you can do something like:

```
blah_or_bleh = nr.filter(my_var="blah") +  nr.filter(my_var="bleh")
```

ogenstad:
Ah yes, remember that. Looks good.

targuan:
I was thinking about proposing to implement django like or SQLAchemy like filtering but you beat me to it. I got stuck will wondering if this should be implemented natively in the inventory filtering system or as a helper in the plugins.

I started an [implementation](https://github.com/nornir-automation/nornir/compare/develop...targuan:feature/filter_helper) more like the [Q object](https://docs.djangoproject.com/en/2.0/topics/db/queries/#complex-lookups-with-q-objects) in django and use it as a func_filter:
```python
f = F(site="site1")
nornir.filter(filter_func=f.match)
```
I thought it allowed more complexe/lighter filtering than just combination operations on arrays:
```python
f = F(groups__contains="group1") & (F(role="www") | F(site__in=["site1","site2"]))
nornir.filter(filter_func=f.match)
```

ktbyers:
Somewhat related to @targuan comments, but should the operations/operators be set operations (unions, intersections, difference, et cetera).

dbarrosop:
Regarding set operations I agree, if you check the last commit I pushed that’s actually what I ended up implemented.

Regarding @targuan’s suggestion I like it a lot so I’d suggest closing this PR and working on his idea. How far are you in the implementation?

thoughts?

targuan:
I have the basic filtering (ie ```__eq__```),  ```and```, ```or``` and ```invert``` operators.
If you allow me to integrate your verify_rules logic, I will also have the \__*\__ operators.

dbarrosop:
Closing in favor of #155

PR #153, opened 2018-06-06
Add Dockerfile

jsimonetti:
This Dockerfile allows executing your python scripts from inside a
docker container.
It requires your scripts to be mounted to the volume at /nornir as this
is set as the working directory.

Usage: # docker run -it --rm -v `pwd`:/nornir nornir:latest get_facts.py

Signed-off-by: Jeroen Simonetti <jeroen@simonetti.nl>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:3.6
+RUN set -x && \
+	pip install nornir
+VOLUME [ "/nornir" ]
+WORKDIR "/nornir"
+ENTRYPOINT [ "/usr/local/bin/python" ]
+#usage:
+# from inside your nornir base directory run: docker run -it --rm -v `pwd`:/nornir nornir:latest get_facts.py


jsimonetti:
How would you like to handle these errors?

The workdir is full path, so that error does noet make sense.
The version should just be latest (imho)
Remains the maintainer. Should I add myself as maintainer or the project? In the latter case, what should I use?

dbarrosop:
Ignore codacy, I thought I had disable that one :S

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6
+RUN set -x && \
+	pip install nornir
+VOLUME [ "/nornir" ]
```
dbarrosop: I don't think exposing this volume adds any benefit. Same for the workdir

dbarrosop:
Don't add the maintainer, it's a deprecated directive anyway: https://docs.docker.com/engine/reference/builder/

Dockerfile
jsimonetti: The workdir should (ideally) be set to the same directory as where you mount your scripts.
Obiously this can be configured using the docker cli, but I thought adding some sane defaults made sense. Both can be overridden if a user wants too.

PR #154, opened 2018-06-09
Proposed fix for SSH and API port mapping issues

ktbyers:


ktbyers:
@dbarrosop @ogenstad Any update on this...just so I don't forget about the context.

ogenstad:
I'm getting ready to vacation in Spain tomorrow, have to get everything in order and have an early flight. Will be back in about a week and won't have much time until then..

ktbyers:
@ogenstad No worries at all :-)

dbarrosop:
Sorry for the delay on this one, would you mind describing the logic? I feel it's quite complex. I expected something more like:

```
optional_args = {}
if self.network_api_port:
    optional_args["port"] = self.network_api_port
...
device = driver(..., optional_args)
```

and let `napalm` decide.

ktbyers:
The main issue is that Nornir has two ways to specify ports: `network_api_port` and `ssh_port`. So the question is what is the proper behavior when Nornir inventory has none of these, one of these, and both of these specified.

In the case of none of these being specified that is pretty easy--just use the NAPALM, netmiko, paramiko default port.

In the case of both of these being specified, then you need to make a logical decision regarding which one should be used. The logical choice is to use the relevant one for the given NAPALM driver. In other words, 'eos' should the API port, 'ios' should use the SSH port. For the NAPALM community drivers, I thought it made sense to prioritize the API port over the SSH port (and I also didn't want to get into the need of updating this code as new NAPALM community drivers were added).

In the case of only one port being specified (either API port of SSH port), this is already handled by the case of both ports being specified. In other words, use the API port if it is an API driver; otherwise, use an SSH port if it is an SSH driver (and use the API > SSH port for the community drivers).

Note, there are two other problems being solved in this PR:
1. Nornir shouldn't default the SSH port if it is not specified in inventory. We have already seen this causes issues for end-users. 
2. There is ambiguity in the 'nxos' driver on whether NX-API should be used or SSH should be used. So we need a way to determine which of the two we should use. So this adds that logic as well.

dbarrosop:
Ok, makes complete sense. Merging.

PR #155, opened 2018-06-11
Filter object

targuan:
As discussed in #152, this is a basic implementation of Django's Q object filtering schema.
In order to use it, one have to construct a filtering object F which will represent how one want to filter.
The objects can be combined with the logicals operators 
- ```&``` which allow object matching the conditions of both F object 
- ```|``` which allow object matching the conditions of one of the F object
- ```~``` which allow object matching none of the conditions of the F object

Tests are quite explicit I think but I can clarify if needed.
I have integrated the others django operators (mostly the contains operator) as in #152.

This implementation is done in the plugins section as I think as this as a helper to filter with more granularity than the basic ```key=value``` method.
If for consistency reason (or other reasons) the F filtering method should be kept as the only method (by converting the **kwargs to a F object) it can be moved to the core.


nornir/plugins/inventory/helpers/filters.py
```diff
@@ -0,0 +1,79 @@
+import copy
```
dbarrosop: I think we should consider this core functionality so I'd place it under `nornir/core/filter.py`

tests/plugins/inventory/helpers/test_filter.py
```diff
@@ -0,0 +1,43 @@
+from nornir.plugins.inventory.helpers.filters import F
```
dbarrosop: As per above comment this should be "tests/core/test_filter.py".

Also, please, follow the same pattern as with the rest of the tests, please.

dbarrosop:
Overall looks great, I have to do some more testing of my own and play a bit with yet but I think this is the right direction :)

targuan:
I moved the filter to nornir/core.
Can you check if the updated tests are what you had in mind.
> follow the same pattern as with the rest of the tests, please.

dbarrosop:
Awesome work! :D

I am going to merge it into a branch and play a bit with it as I add a note in the tutorial about this :)

ktbyers:
Thanks for doing this work @targuan !

PR #157, opened 2018-06-18
Migrate to ruamel.yaml and add `ordered_dict` parameter to `load_json` and `load_yaml` plugins

### LINKED ISSUE #109: migrate to ruaml.yaml
Looks like pyyaml is pretty much abandoned. We should migrate to ruaml.yaml (AnsibleInventory) is already using it.

dbarrosop:
* Provides `OrderedDict` support for load_yaml and load_json
* Fixes #109


ktbyers:
PY27 ordered test failed...but we can also get rid of the PY27 support.

ogenstad:
I think we already agreed that everyone was fine with removing Python 2.7. So I say go ahead. :)

ktbyers:
@ogenstad Yes, I think we agreed Python 3.6 and greater.

PR #158, opened 2018-06-19
WIP - thread safety

dbarrosop:
The idea is that now you can run tasks passing `thread_safe=True` and make sure that those tasks run only one at a time. This is useful if only one of your grouped tasks needs to be thread-safe.

I want to extend this to the `print_result` function so you can print them safely even during subtasks.

If we all agree we want this I will update the docstrings and probably add a note in the execution model.

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -1,6 +1,7 @@
 import logging
 import logging.config
 import sys
+import threading
 from multiprocessing.dummy import Pool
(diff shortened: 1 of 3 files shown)


dbarrosop:
After looking a bit more on this, this makes no sense. You get the same functionality with `num_workers=1` and for more complex workflows users can use  their own locks in a grouped_task.

PR #159, opened 2018-06-23
Support SSH Agent forwarding for paramiko SSH connections

enzzzy:
Adds the ssh_forwardagent (bool) attribute to a Host.
This value can be set in the inventory (nornir_ssh_forwardagent), or can be read from a ssh_config file using the host ForwardAgent flag (see man ssh_config).

Tasks, like remote_command, can then enable Agent Forwarding on the paramiko channel.


ktbyers:
I think we should also consider how this integrates to Netmiko when an SSH config file is used (and also potentially NAPALM for the Netmiko-based drivers).

Unit tests are failing looks like there are some linting and black errors.

enzzzy:
Linting issue has been fixed. 
Are you aware of devices that support ssh agent forwarding, for which there is a netmiko driver?

nornir/core/inventory.py
```diff
@@ -191,6 +191,15 @@ def ssh_port(self):
         """Either ``nornir_ssh_port`` or 22."""
         return self.get("nornir_ssh_port", 22)
 
+    @property
```
dbarrosop: Is it worth adding this as a property? I think I'd rather keep things in the ssh config to avoid making this more complex. What do you think? Is there a strong case where you think it'd be worth managing this in an inventory file instead of in a proper ssh config file?

ktbyers: I thought pretty much the same thing...i.e. that it probably didn't make sense to have a separate attribute for this and that we should just do this through the SSH config file.

enzzzy:
My use case(s) definitely don’t have the need for it to be a separate property. 
I just wanted to make it accessible in other ways then the config file, but it might not make a lot of sense now that I think about it again. (Keep it similar to the proxy command support)
I’ll remove it, if that is the consensus.

dbarrosop:
> I’ll remove it, if that is the consensus.

let's do that, let's just read it from the ssh config. If someone comes with a compelling use case we can revisit this.

nornir/plugins/tasks/connections/paramiko_connection.py
```diff
@@ -35,6 +35,11 @@ def paramiko_connection(task=None):
     if "proxycommand" in user_config:
         parameters["sock"] = paramiko.ProxyCommand(user_config["proxycommand"])
 
+    task.host["ssh_forwardagent"] = False
```
dbarrosop: can you turn into a private attribute (i.e. `task.host._ssh_forward_agent`), please? I don't want to write into the `Host` inventory data. To avoid problems make sure you initialize the same attribute in `Host.__init__()` as `None`.

dbarrosop:
I did minor changes to your code and merged it here: https://github.com/nornir-automation/nornir/commit/0baf86fb76b966be0da65df85eccda1cf64f00f2

If there is anything wrong let me know.

PR #170, opened 2018-07-08
Updating Netmiko plugins to disconnect at the end of a session.

lykinsbd:
When utilizing SSH ProxyCommand to jump through a bastion/jump box, the lack of utilizing `disconnect()` leaves dozens of stale sessions open on the bastion.

This becomes a resource constraint/problem at scale, and starts to make Paramiko choke and die with similar stack traces to the below:

```    'Error reading SSH protocol banner' + str(e)
paramiko.ssh_exception.SSHException: Error reading SSH protocol banner
```

(Sensitive data redacted below)
```    raise ProxyCommandFailure(' '.join(self.cmd), e.strerror)
paramiko.ssh_exception.ProxyCommandFailure: ('ssh USER@BASTION nc IPADDRESS 22', 'Broken pipe')
```

If I force `.disconnect()` on each connection, it largely eliminates these problems.

### DIFF
--- a/nornir/plugins/tasks/networking/netmiko_send_command.py
+++ b/nornir/plugins/tasks/networking/netmiko_send_command.py
@@ -19,4 +19,8 @@ def netmiko_send_command(task, command_string, use_timing=False, **kwargs):
         result = net_connect.send_command_timing(command_string, **kwargs)
     else:
         result = net_connect.send_command(command_string, **kwargs)
+
+    # Cleanup the open net_connect connection, prevents stale sessions when using ssh ProxyCommand
+    net_connect.disconnect()
+
     return Result(host=task.host, result=result)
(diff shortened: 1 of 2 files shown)


dbarrosop:
As commented on slack, this basically removes the ability to reuse connections. We should probably have a "close" task for all the connection plugins we can call either automatically or manually. For instance, right below [this](https://github.com/nornir-automation/nornir/blob/develop/nornir/plugins/tasks/connections/netmiko_connection.py#L12) we could have a `netmiko_close_connection` and have as many tweaks there as necessary. Then we could either let users call that task whenever they want with the tweaks they want or maybe even provide a `Nornir.close_connections()` function that closes all the open connections by calling those tasks with the default parameters (or maybe even reading from the configuration file). Any thoughts?

cc @ktbyers @ogenstad

dbarrosop:
Two more reasons to not do it on the tasks:

1. We'd have to implement on every task, existing or future for consistently.
2. If there is an exception of the task you may leave the connection hanging.

lykinsbd:
Closing this PR and tracking larger scoping of overall connection closure in #171.

PR #172, opened 2018-07-09
replace nsot container with requests-mock

dbarrosop:


ogenstad:
>tests/plugins/inventory/nsot/normal/devices.json

A thought would be to replace `normal` with the version of NSOT that was use, could be helpful if something changes in the future.

dbarrosop:
Done!

PR #174, opened 2018-07-09
remove support for <py36

dbarrosop:
TODO: Update installation guide


ktbyers:
+1 except for the Python3.7 not existing on Travis.

dbarrosop:
Ok, will fix that. I want to first merge as many PRs as possible and make a 1.1 release before removing the support though.

ogenstad:
Might be going back and forth with this, but perhaps we should leave 3.4 and 3.5 until we actually start using a 3.6 only feature.

ktbyers:
I suspect PY3.4 is probably pretty close to EOL (given that we now have 3.5, 3.6, and 3.7)...though definite assumption on my part and I couldn't find any exact date on this.

My preference is 3.6 and greater.

dbarrosop:
Yeah, let's stop messing around with this and go directly with py3.6.

PR #176, opened 2018-07-10
Filter object

dbarrosop:
I made some improvements to the F object and expanded on the filters to build a more generic solution as the current one had a few shortcomings and doing something like `F(site="site1") & ~F(role="www")` wouldn't work. It had a couple of other problems when building more complex queries, hopefully this solution is generic and simple enough.

I also added changed the `filter` function to allow us to do `nr.filter(F(blah="bleh"))` without having to pass it via the `filter_func` argument.

Finally, updated the tutorial.

* TODO: check if `has_parent_group` works


dmfigol:
This looks interesting and much better without `filter_func` argument!
Though, I don't think it current covers the group selection, especially selecting high-level group with hosts several levels down.
Not sure if this should be implemented as part of F filters using a reserved keyword or the `run` method should be available for `Group` object or may be both.

dbarrosop:
I think that should work like this `F(has_parent_group="a_group")`. I suspect it will work out of the box but I will test just in case.

dbarrosop:
@dmfigol I made two changes relevant for the `has_parent_group` discussion:

1. `has_parent_group` now supports testing against a name: https://github.com/nornir-automation/nornir/pull/176/files#diff-e3f6654f418e5eb1d370a0287a483253R208
2. The F object now supports attributes and callables so you can do `F(name="hostname")` or `F(has_parent_group="parent_group")` https://github.com/nornir-automation/nornir/pull/176/files#diff-f6d20f682614ef9064d2f542855fdd65R104

PR #178, opened 2018-07-11
Fix typos

MatthiasGabriel:
Hi,
I fixed a typo and rewrote a small part in the tutorial section.


ogenstad:
This looks good to me.

However, as we are removing Python 2.7 support the second part should perhaps be rewritten a bit. Not sure if there's an issue with older pip versions on Python 3 but it's probably a good idea to mention that it's good to be of Pip 9.0 or above regardless to avoid any pip problems.

dbarrosop:
It's planned to be removed quite soon as we have to keep dealing with py2 and py3 differences and limiting ourselves by the support of it so we don't think it's worth at this point to continue doing so.

PR #181, opened 2018-07-12
Prepare release 1.1.0

dbarrosop:


dbarrosop:
I want to make a release before we remove py2.7.

Issue #186, opened 2018-07-17
is_dry_run per task/per nornir instance handling


dmfigol:
The current implementation for the `dry_run` argument has the following logic:
`dry_run` set per nornir object
`dry_run` set per task, which overrides whatever was set per nornir object.
Here is the code doing that in `nornir.core.task`:
```
    def is_dry_run(self, override=None):
        """
        Returns whether current task is a dry_run or not.

        Arguments:
            override (bool): Override for current task
        """
        return override if override is not None else self.nornir.dry_run
```
In practice, this means that you can set `dry_run=True` per the nornir object and then set `dry_run=False` per your task. I am not sure it is a valid use-case.
Another consequence of the current implementation is that it leads to the following repetition:
`dry_run = task.is_dry_run(dry_run)`
Also, it is very easy to make a mistake if you extend it on your own by using `dry_run` argument directly (which you shouldn't).

Here are proposed changes:
1) In my personal opinion, the logic should be OR, so whenever one of those is set to True, the task will be run as `dry_run`.
2) Add `dry_run` to the Task class constructor as save it as `self._dry_run`
3) Add a decorator to the Task class:
```
@property
def dry_run(self):
  return self._dry_run or self.nornir.dry_run
```
4) Remove `dry_run` as an argument for existing plugins
5) If the user chooses to implement `dry_run` for this plugin, the usage would be:
```
if task.dry_run:
    pass
```

dmfigol:
> meaning with your implementation a host could change the behavior of another host

I am not sure I understand. What do you mean?

> That'd be confusing. You have an argument you claim "overrides the global dry_run"

I don't consider it an override. You can either set it per nornir instance, to run everything as `dry_run` or per current Task `nr.run(dry_run=True)` for a selected number of hosts. 

> Being in the method signature has the advantage if makes it easy to understand where it's supported and where it's not.

Signature is not something users see when they run their code. `dry_run` could (should) be part of the plugin documentation, describing what exactly with happen differently with `dry_run` set.  
On the opposite side, by having `dry_run` as a function argument, it is very easy to use `dry_run` directly instead of `is_dry_run(run)`, creating inconsistencies/unexpected behavior

> There might be other less repetitive solutions if we move towards having tasks be proper classes instead of just functions with a proper signature (which I want to explore)

Very likely, I agree

dmfigol:
[...]

PR #188, opened 2018-07-18
Change dry_run from task override to task.dry_run or nornir.dry_run

dmfigol:
Refers to #186


dmfigol:
I will close this PR until it is decided what is the best approach after 2.0 refactoring

PR #195, opened 2018-07-22
Refactor connection plugins

dbarrosop:
From an API perspective the user can:

```
def connect_and_close_manually(task):
    task.host.open_connection("napalm", ...)  # to open a connection manually
    task.host.close_connection("napalm", ...)  # to close the connection manually
```

You can't do like before:

```
nr.run(task=napalm_connection, ...)
```

Although if we think that's valuable it should be easy to write connection tasks that are just wrappers like this:

```
def napalm_connection(task, ...):
    task.host.open_connection("napalm", ...)
    return Result("it works")
```

The method `Host.get_connection` works as before. However, I moved there the logic to be able to override parameters or default to the ones specified by the inventory. That way the plugins can assume the parameters they got are the correct ones. I want to discuss the behavior or napalm/netmiko/paramiko as I think they were unnecessarily complex (I rewrote napalm and netmiko but let paramiko ummodified).

Interestingly this code is consistently 12s faster when running tests than the previous one.

@dmfigol @ktbyers @ogenstad


dmfigol:
@dbarrosop 
It will take me a couple of days to go through PR.
Meanwhile, just a couple of comments:
1) I am thinking if we could (or should) convert InitNornir to context manager. So the usage will be something like:
```
with InitNornir("config.yaml") as nr:
  ...
```
`__exit__` would go through the list of all open connections and close them even if an exception occurred 
2) Do you know where speed increase is coming from?

dbarrosop:
1. It's a bit off topic, I'd suggest opening a new issue if you want to discuss that but I'd suggest first thinking what would be the gain of it. To me it feels like it's just going to add an extra level of indentation to all of your code without much benefit (if any at all).
2. I haven't profiled the code but I suspect it's the layer of indirection that allowed the connection plugins to behave like tasks.

dbarrosop:
There are basically two motivations:

1. Get rid of some complexities and indirections.
2. Allow connection plugins to know how to open and close connections and expose those via a coherent API.

Regarding breaking changes, there is only one breaking change here and it affects only users opening the connections manually (which I assume it's just a few). The breaking change is explained in the description and we can actually avoid it.

I will be sending other PRs in the upcoming weeks changing some other internals although I don't expect those to affect end users.

In summary, I don't think nornir 2.x will affect regular users but it might slightly change how task and connection plugins are written.

dbarrosop:
I added a context manager and a method to close all the connections manually. I have to add tests for this and other things though but I won't do it until we have agreed on that this is the right direction.

ktbyers:
@dbarrosop Okay, I had not read @dmfigol earlier issue on connections so I am looking at that now.

nornir/plugins/tasks/networking/napalm_cli.py
```diff
@@ -14,6 +14,6 @@ def napalm_cli(task: Task, commands: List[str]) -> Result:
         Result object with the following attributes set:
           * result (``dict``): result of the commands execution
     """
-    device = task.host.get_connection("napalm")
+    device = task.host.get_connection("napalm").connection
```
ktbyers: This is strange to me...

I would think `.get_connection("napalm")` should return the napalm connection objection and not the nornir connection object.

In almost every case this is what we want:

```
./plugins/tasks/networking/napalm_validate.py:    device = task.host.get_connection("napalm").connection
./plugins/tasks/networking/napalm_configure.py:    device = task.host.get_connection("napalm").connection
./plugins/tasks/networking/netmiko_file_transfer.py:    net_connect = task.host.get_connection("netmiko").connection
./plugins/tasks/networking/netmiko_send_command.py:    net_connect = task.host.get_connection("netmiko").connection
./plugins/tasks/networking/napalm_get.py:    device = task.host.get_connection("napalm").connection
./plugins/tasks/networking/napalm_cli.py:    device = task.host.get_connection("napalm").connection
./plugins/tasks/networking/netmiko_send_config.py:    net_connect = task.host.get_connection("netmiko").connection
./plugins/tasks/commands/remote_command.py:    connection = task.host.get_connection("paramiko")
./plugins/tasks/files/sftp.py:    client = task.host.get_connection("paramiko").connection
```

The exception being the remote_command Paramiko task for the SSH forwarding agent.  

IMO it should be the other way i.e. get_connection("napalm") returns the napalm connection object then if a task wants the Nornir connection object they should have to do something else special to return the nornir connection object.

The thing I don't like about it is that from a library user perspective it doesn't do what you would expect.

In the SSH agent case, we have inventory information (i.e. the intent to use SSH agent, that we are storing in an alternate config file 'ssh_config' that we are then trying to stuff into Nornir connection state). This doesn't sound like the right way to do this...but would have to dig into it more.

ktbyers:
In general, I like @dmfigol design proposal for this and your implementation, I think it is a better solution than what we had previously.

nornir/core/inventory.py
```diff
@@ -230,43 +232,108 @@ def nos(self):
...
+                operating_system,
+                nos,
+                connection_options,
+                configuration,
+            )
         return self.connections[connection]
 
+    def close_connection(self, connection: str) -> None:
+        """ Close the connection"""
+        self.connections[connection].close()
```
ktbyers: When this is done, shouldn't it remove the dictionary entry from connections?

It surprised that after I did this:

```
task.host.close_connection("netmiko")
```

That I would then see:

```
(Pdb) task.host.connections
{'netmiko': <nornir.plugins.connections.netmiko.Netmiko object at 0xb640b6bc>}
(Pdb) task.host.connections['netmiko']
<nornir.plugins.connections.netmiko.Netmiko object at 0xb640b6bc>
(Pdb) task.host.connections['netmiko'].connection
<netmiko.cisco.cisco_ios.CiscoIosSSH object at 0xb640f12c>
(Pdb) task.host.connections['netmiko'].connection.remote_conn
(Pdb) p task.host.connections['netmiko'].connection.remote_conn
None
```

i.e. it was still in the Nornir connections dictionary even though there was no active connection.

ktbyers: Related to this should `open_connection()` do any checking with respect to the connection already being in the `connections` table?

open_connection() will just create a new object regardless of whether an entry already exists in the connections table or not:

```
        self.connections[connection] = self.nornir.get_connection_type(connection)()
        if default_to_host_attributes:
```

nornir/plugins/connections/napalm.py
```diff
@@ -0,0 +1,50 @@
...
+    ) -> None:
+        connection_options = connection_options or {}
+        if network_api_port:
+            connection_options["port"] = network_api_port
+
+        parameters = {
+            "hostname": hostname,
+            "username": username,
+            "password": password,
+            "timeout": connection_options.get("timeout"),
```
ktbyers: timeout being set to None causes issues on NAPALM (I saw this on NX-OS)

ktbyers:
These changes break this config for `nxos_ssh`

```
nxos:
  nornir_nos: nxos
  nornir_ssh_port: 22
```

i.e. it tries to use NX-API and not `nxos_ssh`.

i.e. some of these issues still exist:

https://github.com/nornir-automation/nornir/pull/154

nornir/core/connections.py
```diff
@@ -0,0 +1,58 @@
...
+        self.state: Dict[str, Any] = {}
+
+    @abstractmethod
+    def open(
+        self,
+        hostname: str,
+        username: str,
+        password: str,
+        ssh_port: int,
+        network_api_port: int,
```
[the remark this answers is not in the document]
dbarrosop: I am glad you bring this topic because I was thinking exactly the same. I suggest ticketing this one and discussing this before releasing 2.0.0.

nornir/core/connections.py
```diff
@@ -0,0 +1,58 @@
...
+        operating_system: str,
+        nos: str,
```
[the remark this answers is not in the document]
dbarrosop: Problem using `os` is that it shadows `import os`. We should probably discuss getting rid of `os` and `nos` and just converging in `operating_system`. We can discuss this in the same issue as the `ssh_port/network_api_port`.

nornir/core/connections.py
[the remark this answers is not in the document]
dbarrosop: Issue: #202

nornir/core/connections.py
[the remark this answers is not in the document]
dbarrosop: Issue: #202

nornir/core/inventory.py
dmfigol: I agree with Kirk, I'd expect something like this:
```
    def close_connection(self, connection: str) -> None:
        self.connections.pop(connection).close()
```

dmfigol: I disagree about `open_connection` though. IMO, `open_connection` should literally open a new connection. Plugins should use `get_connection`, which gets an existing connection if exists, or opens a new one.

nornir/core/inventory.py
```diff
@@ -230,43 +232,108 @@ def nos(self):
...
+
+    def close_connections(self) -> None:
+        for connection in self.connections.values():
+            if connection is not None:
+                connection.close()
+
```
dmfigol: I propose:
```
    def close_connections(self) -> None:
        for connection in self.connections:
            self.close_connection(connection)
```

nornir/plugins/tasks/networking/napalm_cli.py
dmfigol: I agree with Kirk. I don't really like this repetition.
I propose having:
`def get_connection(self, connection: str) -> Any`
which returns `self.connections[connection].connection`
instead of
`def get_connection(self, connection: str) -> ConnectionPlugin`
and
```
def get_connection_state(self, connection: str) -> Dict[str, Any]:
        raise ValueError(f'Connection {connection} is not open`)
    return self.connections[connection].state
```

nornir/core/__init__.py
```diff
@@ -79,9 +82,13 @@ def __init__(
...
-            self.available_connections = available_connections
-        else:
-            self.available_connections = connections.available_connections
+            self.data.available_connections = available_connections
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close_all_connections()
```
dmfigol: Currently there isn't a method `self.close_all_connections()`, only `self.close_connections()`. It can be fixed either here or where `def close_connections` is defined. I am ok with both naming.
Also, I think it is important to do it safely either here or under the method definition with catch-all try/except block. Since `__exit__` will be called before exit, either graceful or due to an exception, getting another exception during `__exit__` will make it harder to debug. I know it is currently being called using task, but users may use `raise_on_error=True`.

nornir/core/inventory.py
dbarrosop: Re `close_connection`, yeah, we should remove it from available connections. Re `open_connection`, we could raise an exception if the connection is already open.

nornir/core/connections.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class UnestablishedConnection(object):
+    def close(self) -> NoReturn:
+        raise ValueError("Connection not established")
+
+    disconnect = close
+
+
+class Connections(Dict[str, ConnectionPlugin]):
```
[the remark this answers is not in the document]
dbarrosop: > I am not sure about the purpose of this class.

The purpose is to have a distinct object in case we want to add methods to it.

> It also inherits from the type Dict, which is not correct.

What would be the correct way? From what I read this seems to be the correct way of doing this (it certainly works as I expected). For instance:

```
class Asd(Dict[str, int]):
    pass


a = Asd()
a["asd"] = "asd"

error: Incompatible types in assignment (expression has type "str", target has type "int")
```

> this class could replace both available_connections and connection instances for Host

Not sure they are the same thing though. For instance, the `register_connection` you mention may make sense in one but not the other but I will look into it as I address some of the other comments.

dbarrosop:
> So I could have host.connections['netmiko'] and host.connections['netmiko_console'] and both of those would create Netmiko object but with different connection parameters.

That's doable but I'd leave that outside of this PR to avoid delaying this work because of new features.

nornir/core/inventory.py
```diff
@@ -230,43 +232,108 @@ def nos(self):
...
+        Raises:
+            AttributeError: if it's unknown how to establish a connection for the given type
+
+        Returns:
+            An already established connection
+        """
+        self.connections[connection] = self.nornir.get_connection_type(connection)()
+        if default_to_host_attributes:
+            self.connections[connection].open(
+                hostname if hostname is not None else self.host,
```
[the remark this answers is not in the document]
dbarrosop: I am going to fix it even though I think we shouldn't as it makes the code unnecessarily more verbose. IMO, we shoudn't have to account for users not following the specs.

dbarrosop:
> i.e. it tries to use NX-API and not nxos_ssh.

Yeah, I don't think we should address this problem. Nornir is not an abstraction layer for netmiko/napalm. More about this [here](https://github.com/nornir-automation/nornir/issues/202#issuecomment-408598841)

nornir/core/connections.py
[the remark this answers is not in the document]
dbarrosop: I agree this is the right thing to and I even tried a similar implementation, however, this change is a bit more tricky than it looks as it runs into circular imports if you pre-register the connections shipped with nornir. In any case, this is just a minor OCD fix to the internal API so feel free to ticket it so we can take a closer look but I am not going to address it in this PR.

nornir/core/inventory.py
```diff
@@ -230,43 +232,108 @@ def nos(self):
...
+                else self.network_api_port,
+                operating_system if operating_system is not None else self.os,
+                nos if nos is not None else self.nos,
+                connection_options
+                if connection_options is not None
+                else self.get(f"{connection}_options"),
+                configuration if configuration is not None else self.nornir.config,
+            )
+        else:
+            self.connections[connection].open(
```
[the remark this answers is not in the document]
dbarrosop: Will fix as well.

dbarrosop:
@ktbyers @dmfigol I think all the comments have either been ticket for further discussion or addressed. Would you mind taking another quick look before merging and we start discussing the other more "complex" issues raised?

nornir/core/connections.py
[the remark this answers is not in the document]
dmfigol: #211

nornir/core/exceptions.py
```diff
@@ -1,6 +1,19 @@
...
+        self.connection = connection
+
+
+class ConnectionAlreadyOpen(ConnectionException):
+    pass
+
+
+class ConnectionNotOpen(Exception):
+    pass
+
```
dmfigol:
```
class ConnectionNotOpen(ConnectionException): 
```
?

dbarrosop: Not sure what you are asking.

dmfigol: if `ConnectionNotOpen` should inherit from `ConnectionException`?

ktbyers:
@dbarrosop @dmfigol I think we should go ahead and merge this PR and then move forward with this as the new base.

We can then discuss the other open issues in separate issues/PRs.

So +1 from my perspective.

nornir/core/exceptions.py
dbarrosop: Right, that's a typo. Fixing.

dmfigol:
LGTM!

PR #200, opened 2018-07-26
Allow setting of attribute for nornir_ attributes

ktbyers:
Do Not Merge:

Allow the nornir_ attributes that are using @property to have a setter.

This would allow you to do:

```
nornir_host.username = 'whatever'
nornir_host.password = getpass()
```

Could also implement using the @property .setter format if that is preferred.

If this looks reasonable, I will update and add unit tests (add cleanup changes unrelated to `__setattr__`).

### DIFF
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -199,20 +199,40 @@ class Host(object):
         """String used to connect to the device. Either ``nornir_host`` or ``self.name``"""
         return self.get("nornir_host", self.name)
 
+    @host.setter
+    def host(self, value):
+        """Allow host to be assigned as an attribute."""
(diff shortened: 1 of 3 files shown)


dbarrosop:
Thanks. Can we use the `@property.setter` as you suggest though?

dmfigol:
Makes perfect sense!
+1 to `@property.setter`

ktbyers:
That is fine. I had `@property.setter` here so I will just rollback to that:

https://github.com/nornir-automation/nornir/pull/200/commits/d97fc0ca552a7f5b2e70218b30336c60ed04be10

dbarrosop:
I guess we can hold onto this one until we close #202

PR #215, opened 2018-07-31
Rebase '2.0' on develop to add in 'netbox' inventory plugin

bdlamprecht:


dbarrosop:
Not sure what happened but your branch seems to be duplicating some commits so I opened #220 instead.

PR #218, opened 2018-08-01
Patch connection close

ktbyers:
Calling this in some Nornir code:

```
brg.close_connections()
```

Generated a bunch of exceptions pertaining to dictionary size changing while iterating over dictionary:

```
hp_procurve: Traceback (most recent call last):
  File "/home/gituser/nornir/nornir/core/task.py", line 63, in start
  File "/home/gituser/nornir/nornir/core/__init__.py", line 247, in close_connections_task
    task.host.close_connections()
  File "/home/gituser/nornir/nornir/core/inventory.py", line 353, in close_connections
    for connection in self.connections:
RuntimeError: dictionary changed size during iteration
```

### DIFF
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -350,7 +350,9 @@ class Host(object):
         self.connections.pop(connection).close()
 
     def close_connections(self) -> None:
-        for connection in self.connections:
+        # Decouple deleting dictionary elements from iterating over dict
+        existing_conns = list(self.connections.keys())
+        for connection in existing_conns:
             self.close_connection(connection)
 
 
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -350,7 +350,7 @@ class Host(object):
         self.connections.pop(connection).close()
 
     def close_connections(self) -> None:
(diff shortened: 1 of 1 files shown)


dbarrosop:
Oh, I just approved it, but would you mind adding a test here?

https://github.com/nornir-automation/nornir/blob/2.0/tests/core/test_connections.py#L8

something like:

```
def open_connections(task):
    task.host.open_connection("dummy")
    task.host.open_connection("dummy2")
    assert len(task.host.connections) == 2

...

    def test_close_not_opened_connection(self, nornir):
        nornir.data.available_connections["dummy"] = DummyConnectionPlugin
        nornir.data.available_connections["dummy2"] = DummyConnectionPlugin

        nr = nornir.filter(name="dev2.group_1")
        r = nr.run(task=open_connections, num_workers=1)
        assert not r.failed
        nr.close_connections()
        assert len(nr.inventory.hosts["dev2.group_1"].connections) == 0
```

Or something like that.

ktbyers:
@dbarrosop Sure sounds good (on adding the test).

dbarrosop:
I am going to merge even without the test I asked for as I think the benefit of the fix outweighs the lack of test.

PR #222, opened 2018-08-03
Include interface name of 'nornir_host'

bdlamprecht:
There was some functionality that was lacking which I thought would be useful for generating configurations from `jinja` templates.

The current implementation simply added values to a `host` that were returned by `NetBox`'s API call to `/api/dcim/devices/`.  This was missing the name of the `interface` associated with `nornir_host`.

I [asked](https://github.com/digitalocean/netbox/issues/2302) the developer of `NetBox` to include but it by default.
However, it turns out doing so would be an expensive DB operation.
So I added another API call internal to this plugin.

The explanation for adding this attribute is as follows:

>Network devices using `junos` use the format `source-address X.X.X.X` in their configuration.
>However, network devices using `ios` use the format `source-interface [NAME]` instead.

This change allows you to reference `{{ primary_interface }}` within the template.

### DIFF
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -45,10 +45,10 @@ class NBInventory(Inventory):
                     headers=headers,
                 )
                 i.raise_for_status()
-                i = i.json()
+                int_name = i.json()
 
                 # Strip extraneous information and add it to the dictionary
-                temp["primary_interface"] = i["results"][0]["interface"]["name"]
+                temp["primary_interface"] = int_name["results"][0]["interface"]["name"]
 
             # Add values that don't have an option for 'slug'
             temp["serial"] = d["serial"]
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -40,10 +40,9 @@ class NBInventory(Inventory):
                 # Find interface name using additional API call
                 i = requests.get(
                     "{}/api/ipam/ip-addresses/?address={}".format(
-                        nb_url, temp["nornir_host"]
-                    ),
-                    headers=headers,
-                ).json()
+                        nb_url, temp["nornir_host"]),headers=headers)
+                i.raise_for_status()
+                i = i.json()
 
                 # Strip extraneous information and add it to the dictionary
                 temp["primary_interface"] = i["results"][0]["interface"]["name"]
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -40,7 +40,10 @@ class NBInventory(Inventory):
                 # Find interface name using additional API call
                 i = requests.get(
                     "{}/api/ipam/ip-addresses/?address={}".format(
-                        nb_url, temp["nornir_host"]),headers=headers)
+                        nb_url, temp["nornir_host"]
+                    ),
+                    headers=headers,
+                )
                 i.raise_for_status()
                 i = i.json()
 
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -37,6 +37,17 @@ class NBInventory(Inventory):
             if d.get("primary_ip", {}):
                 temp["nornir_host"] = d["primary_ip"]["address"].split("/")[0]
 
+                # Find interface name using additional API call
+                i = requests.get(
+                    "{}/api/ipam/ip-addresses/?address={}".format(
+                        nb_url, temp["nornir_host"]
+                    ),
+                    headers=headers,
+                ).json()
+
+                # Strip extraneous information and add it to the dictionary
+                temp["primary_interface"] = i["results"][0]["interface"]["name"]
+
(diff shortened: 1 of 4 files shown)


bdlamprecht:
I'm not sure how `requests.mock` is supposed to work but it is failing because that was written without this functionality added.

dbarrosop:
You will have to save in different files the output of all the calls the test needs to perform and tell request-mock which one to use for which call in a code similar to this:

https://github.com/nornir-automation/nornir/blob/develop/tests/plugins/inventory/test_netbox.py#L14-L19

nornir/plugins/inventory/netbox.py
```diff
@@ -37,6 +37,17 @@ def __init__(
...
                 temp["nornir_host"] = d["primary_ip"]["address"].split("/")[0]
 
+                # Find interface name using additional API call
+                i = requests.get(
+                    "{}/api/ipam/ip-addresses/?address={}".format(
+                        nb_url, temp["nornir_host"]
+                    ),
+                    headers=headers,
+                ).json()
+
```
dmfigol: By default, `requests` does not raise an exception when something goes wrong.
Could you please add something like this:
```
response = requests.get(...)
if response.ok:
    temp["primary_interface"] = response.json()...
```
Alternatively, you can achieve similar thing using `response.raise_for_status()` and catching `requests` exceptions.

dbarrosop: I'd suggest using `raise_for_status` as above.

bdlamprecht: Technically, the first `r.raise_for_status()` line [27](https://github.com/nornir-automation/nornir/pull/222/files#diff-71d4c60e78b698e99f071d4074c15052R27) will catch any errors in the submitted request (i.e. an interface **MUST** have an address first, only if that is successful, then I'm querying for the name).

The only way a second error would occur is if the entire DB is screwed up.
Then if that is the case, you have some other issues to worry about.

All that being said, the likelihood that this explanation is going to be remembered in even 1 months time-frame is small, so I'm okay adding the additional `r.raise_for_status()` simply for documentation purposes.

However, I don't have the time this week to add data for the `requests.mock` tests that are failing, so I'll do this next week so as to not have several failed `travis-ci` tests on record :smile:

dbarrosop: > The only way a second error would occur is if the entire DB is screwed up

There could be other reasons, load causing a timeout, lots of connections opened against the databases leading to starvation of resources, a service crash... We should check that on every request just for safety.

> I don't have the time this week to add data for the requests.mock tests that are failing

No rush, anything you need let us know.

bdlamprecht:
As asked, I altered the code to include `requests.raise_for_status()` in the 2nd API call to get the interface name.

I also added some `JSON` files showing what the output of querying `api/ipam/ip-addresses/?address=10.0.1.1` should look like, but I have no idea how to integrate that check in your `test_netbox.py` code.

nornir/plugins/inventory/netbox.py
bdlamprecht: As I said in the comment on the recent commit I made, I don't know how to integrate the `JSON` files for the `address_[1,2,3]` in your test file which is why test 1 (`py36`) is failing.

And after spending _some_ learning about what `mypy` is, I have no idea why test 5 is failing.

nornir/plugins/inventory/netbox.py
```diff
@@ -37,6 +37,19 @@ def __init__(
...
 
+                # Find interface name using additional API call
+                i = requests.get(
+                    "{}/api/ipam/ip-addresses/?address={}".format(
+                        nb_url, temp["nornir_host"]
+                    ),
+                    headers=headers,
+                )
+                i.raise_for_status()
+                i = i.json()
```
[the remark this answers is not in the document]
bdlamprecht: I renamed it to `int_name` on line 48 and the later reference on line 50.
Still need help understanding how to integrate the JSON data correctly for `requests_mock`.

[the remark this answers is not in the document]
dbarrosop: You need to call requests-mock like this:

```
https://github.com/nornir-automation/nornir/blob/develop/tests/plugins/inventory/test_netbox.py#L15-L19
```

The first parameter is the URL you want to mock, the `json` parameter is the json output that call should return. So basically you can save the `json` output in a file, load it and pass it to that function.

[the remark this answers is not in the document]
bdlamprecht: Okay, _think_ I can see how that works, but I'm struggling to understand how to correctly structure the `JSON` files I included within the tests you wrote to allow to solve this error `py36`:
```
>       raise exceptions.NoMockAddress(request)
E       requests_mock.exceptions.NoMockAddress: No mock address: GET http://localhost:8080/api/ipam/ip-addresses/?address=10.0.1.1
.tox/py36/lib/python3.6/site-packages/requests_mock/adapter.py:233: NoMockAddress
```

bdlamprecht:
Closing this PR as it is not longer relevant to `nornir v2`

PR #223, opened 2018-08-03
POC on connection ConnectionPlugin arguments moving to object

ktbyers:


nornir/core/connections.py
```diff
@@ -18,12 +18,7 @@ class ConnectionPlugin(ABC):
 
     __slots__ = ("connection", "state")
```
dbarrosop: how is it possible this works without updating `__slots__`?

ktbyers: Yeah, don't know...would have to dig into that :-)

dbarrosop:
Should we close this one or is this something you still want to discuss?

PR #227, opened 2018-08-13
Migrate how tos to testable notebooks

dbarrosop:


dbarrosop:
This one is ready. I really like the pattern of building testable notebooks to explain nornir as you can conflate documentation and tests to make sure the documentation stays true and relevant.

nornir/core/inventory.py
```diff
@@ -70,11 +70,11 @@ class Host(object):
     def __init__(self, name, groups=None, nornir=None, defaults=None, **kwargs):
         self.nornir = nornir
         self.name = name
-        self.groups = groups or []
+        self.groups = groups if groups is not None else []
```
ogenstad: What am I missing, why was this done? Do we want groups to be able to be `False`?

nornir/core/inventory.py
```diff
@@ -70,11 +70,11 @@ class Host(object):
...
         self.data = {}
         self.data["name"] = name
         self.connections = Connections()
-        self.defaults = defaults or {}
+        self.defaults = defaults if defaults is not None else {}
```
ogenstad: Same as above with groups.

nornir/core/inventory.py
dbarrosop: If the inventory creates a host passing an empty list of groups it has to be accepted. Otherwise, when the inventory adds a group later on the Host won't have access to it.

PR #251, opened 2018-10-01
Always go into enable mode for Netmiko configuration changes

ktbyers:


ktbyers:
The build failure is totally unrelated to the change (i.e. it is a sphinx error probably related to some underlying package change).

dbarrosop:
I am merging this PR as the issue is unrelated and it's going to popup somewhere else anyway.

PR #254, opened 2018-10-13
Refactor Configuration

dbarrosop:
Same pattern as the inventory:

1. A deserializer takes care of validating user input and implementing the logic of transforming some strings into objects (like loading the right inventory plugin)
2. A set of config objects hold the real data inside nornir
3. I also moved all the parameters inside a section, meaning that `num_workers` and `raise_on_error` are now under `core`  (ie, `config.core.num_workers`) and `jinja_filters` unless `jinja2`(ie, `config.jinja2.filters`)


dbarrosop:
@ktbyers @ogenstad @dmfigol unless someone says something against it in the next following days my plan is to:

1. Merge this PR into #243 
2. Merge #243 into 2.0
3. Rebase 2.0 branch
4. Release 2.0.0alpha1
4. Start working on issues with milestone 2.0.0

There are a number of opened PRs that will require rebasing as well

docs/_data_templates/configuration-parameters.j2
```diff
@@ -1,29 +1,29 @@
...
+   * - **Descrpiption**
+     - {{ v["description"] }}
+   * - **Type**
+     - ``{{ v["type"] }}``
+   * - **Default**
+     - {{ "``{}``".format(v["default"]) if v["default"] else "" }}
+   * - **Required**
+     - ``{{ v["required"] }}``
+   * - **Environment Variable**
+     - ``{{ "NORNIR_{}_{}".format(section, k).upper() }}``
```
ogenstad: I haven't looked at the entire PR, along with all of the changes to Nornir that have been made while I haven't been paying attention. However, my idea for this was to write out "N/A" for the variable types which take a list or dictionary as those can't be set through environment variables. Just seeing at this it looks like all settings can be configured using environment variables. So, that's why the if statement which is removed is included.

dbarrosop: As per pydantic documentation:

> Complex types like list, set, dict and submodels can be set by using JSON environment variables.

So it should work.

dmfigol:
I disagree. We haven't released 2.0 and if someone is using that branch (I am using it too for my projects), they should be prepared that breaking changes could be added to that branch.

dbarrosop:
@dmfigol I agree with you but there is no point in arguing about it. The important thing is the merge to `develop` anyway.

bradh11:
For what it's worth I am a user of the 2.0 branch...  I willingly selected
it knowing that there could be breaking changes when it was formally
released.  That is one of the points of having a major release revision in
the first place (that breaking changes could be introduced).  Go forth and
don't look back!  :-)

docs/_data_templates/configuration-parameters.j2
ogenstad: Ah, wasn't aware that was possible.

docs/configuration/index.rst
```diff
@@ -1,4 +1,25 @@
...
+        core={"num_workers": 20},
+        logging={"file": "mylogs", "level": "debug"}
+    )
+
+A similar example using a ``yaml`` file:
+
+.. include:: ../howto/advanced_filtering/config.yaml
+   :code: yaml
+
+A continuation you can find each section and their corresponding options.
```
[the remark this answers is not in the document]
dbarrosop: Would "Next" or "Below" sound better? Other? "A continuation" makes perfect sense to me but I am not a native speaker :P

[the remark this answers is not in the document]
dbarrosop: I am changing it to "Next". If that sounds odd as well let me know and feel free to propose an alternative.

nornir/core/deserializer/configuration.py
```diff
@@ -0,0 +1,184 @@
...
+
+    class Config:
+        env_prefix = "NORNIR_SSH_"
+        ignore_extra = False
+
+    @classmethod
+    def deserialize(self, **kwargs) -> configuration.SSHConfig:
+        s = SSHConfig(**kwargs)
+        return configuration.SSHConfig(**s.dict())
+
```
dmfigol: I propose `cls` instead of `self` to follow convention and `s = cls(**kwargs)`
The same would apply to other classes below

nornir/core/deserializer/configuration.py
```diff
@@ -0,0 +1,184 @@
...
+class CoreConfig(BaseSettings):
+    num_workers: int = Schema(
+        default=20,
+        description="Number of Nornir worker threads that are run at the same time by default",
+    )
+    raise_on_error: bool = Schema(
+        default=False,
+        description=(
+            "If set to ``True``, (:obj:`nornir.core.Nornir.run`) method of "
+            "will raise an exception if at least a host failed"
```
dmfigol: Can we please add a reference to `nornir.core.exceptions.NornirExecutionError` here?

nornir/core/deserializer/configuration.py
```diff
@@ -0,0 +1,184 @@
...
+def _resolve_import_from_string(import_path: Any) -> Optional[Callable[..., Any]]:
+    try:
+        if not import_path:
+            return None
+        elif callable(import_path):
+            return import_path
+        module_name = ".".join(import_path.split(".")[:-1])
+        obj_name = import_path.split(".")[-1]
+        module = importlib.import_module(module_name)
+        return getattr(module, obj_name)
```
dmfigol: Some nit-picking, but I think this is a little bit more readable:
`module_name, obj_name = import_path.rsplit('.', 1)`

docs/howto/advanced_filtering/config.yaml
```diff
@@ -1,4 +1,6 @@
 ---
+core:
+    num_workers: 20
 inventory:
     plugin: nornir.plugins.inventory.simple.SimpleInventory
     options:
```
dmfigol: can we add here `user_defined` for completeness?

nornir/core/deserializer/configuration.py
```diff
@@ -0,0 +1,184 @@
...
+
+class InventoryConfig(BaseSettings):
+    plugin: Any = Schema(
+        default="nornir.plugins.inventory.simple.SimpleInventory",
+        description="Import path to inventory plugin",
+    )
+    options: Dict[str, Any] = Schema(
+        default={}, description="kwargs to pass to the inventory plugin"
+    )
+    transform_function: Any = Schema(
```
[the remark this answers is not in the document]
dbarrosop: Not really, it's an error. Should be `str`

docs/howto/advanced_filtering/config.yaml
dbarrosop: I think we should add a proper example about that instead of in here, which is for something else so let's treat this as a quick and dirty documentation hack to avoid hogging this even more

nornir/core/deserializer/configuration.py
dbarrosop: done

nornir/core/deserializer/configuration.py
dbarrosop: done

dbarrosop:
Thanks for the review. This was a big one. I am going to merge this into `inventory_rework` and address some of the comments there before merging it into a new `2.0.0` branch

PR #263, opened 2018-10-24
remove need of Config object in the Inventory and Hosts

dbarrosop:
I am trying to get rid of as much magic as possible.

Solves #252

### DIFF
--- a/docs/howto/handling_connections.ipynb
+++ b/docs/howto/handling_connections.ipynb
@@ -103,7 +103,7 @@
    ],
    "source": [
     "def task_manages_connection_manually(task):\n",
-    "    task.host.open_connection(\"napalm\")\n",
+    "    task.host.open_connection(\"napalm\", configuration=task.nornir.config)\n",
     "    r = task.run(\n",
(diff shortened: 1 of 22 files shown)


ktbyers:
+1

I am fine with this as-is. 

The only part I wonder is for get_connection() whether we should have a default value for configuration (i.e. configuration=None).

In other words, whether we should allow the below to work (in the case where the connections already exist):

ipdb> p norn.inventory.hosts['pynet-rtr1'].get_connection("netmiko")                                                                                                                                                   
<netmiko.cisco.cisco_ios.CiscoIosSSH object at 0x7f102814ca20>

And then have it fail on the open_connection() part i.e. have it fail if configuration=None.

I am fine either way on this, but it will probably imply I will use the below as the default way to access the connection table for existing connections.

```
ipdb> p norn.inventory.hosts['pynet-rtr1'].connections['netmiko']                                                                                                                                                      
<nornir.plugins.connections.netmiko.Netmiko object at 0x7f1028170c60>
```

PR #271, opened 2018-11-08
Plugin for netmiko save_config() method

enzzzy:
Add plugin for netmiko's save_config() method.
Not sure how I can develop a test case for this, so help / suggestions would be welcome.

### DIFF
--- a/nornir/plugins/tasks/networking/netmiko_save_config.py
+++ b/nornir/plugins/tasks/networking/netmiko_save_config.py
@@ -1,0 +1,27 @@
+from __future__ import unicode_literals
+
+from nornir.core.task import Result, Task
+
+
+def netmiko_save_config(
+    task: Task, cmd: str = "", confirm: bool = False, confirm_response: str = ""
+) -> Result:
+    """
+    Execute Netmiko save_config method
+    Arguments:
+        cmd(str, optional): Command used to save the configuration.
+        confirm(bool, optional): Does device prompt for confirmation before executing save operation
+        confirm_repsonse(str, optional): Response send to device when it prompts for confirmation
+
+    Returns:
+        :obj: `nornir.core.task.Result`:
+          * result (``dict``): dictionary showing the CLI from the save operation
+    """
+    conn = task.host.get_connection("netmiko", task.nornir.config)
+    if cmd:
+        result = conn.save_config(
+            cmd=cmd, confirm=confirm, confirm_response=confirm_response
+        )
+    else:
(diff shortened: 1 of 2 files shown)


ktbyers:
@enzzzy You should probably submit this against the `nornir-automation:2.0.0-beta` branch (instead of against `develop`).

ktbyers:
You might want to add a `confirm_response` argument also. Here is the default method definition inside of Netmiko (for Cisco-ios like devices):

save_config has the following arguments: `cmd`, `confirm`, `confirm_response` (by default confirm_response will just be a newline, but this allows you to change it.

dbarrosop:
I just changed the base of the PR to `2.0.0-beta`

enzzzy:
Branched of `2.0.0-beta` now. Added confirm_resonse argument and added annotations.
Not sure why the check is failing?

dbarrosop:
You need to format your code with black to adhere to the code style. If you install the requirements specified in `requirements-dev.txt` and execute `black` in the root of the repo it should fix itself.

nornir/plugins/tasks/networking/netmiko_save_config.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def netmiko_save_config(
+    task: Task, cmd: str = "", confirm: bool = False, confirm_response: str = ""
+) -> Result:
+    """
+    Execute Netmiko save_config method
+    Arguments:
+        cmd(str, optional): Command used to save the configuration.
+        confirm(bool, optional): Does device prompt for confirmation before executing save operation
```
ktbyers: Can you add `confirm_response` to the docstring? This would be the string that gets sent in response to confirmation if required (for example a "Y").

nornir/plugins/tasks/networking/netmiko_save_config.py
```diff
@@ -0,0 +1,26 @@
...
+        :obj: `nornir.core.task.Result`:
+          * result (``dict``): dictionary showing the CLI from the save operation
+    """
+    conn = task.host.get_connection("netmiko", task.nornir.config)
+    if cmd:
+        result = conn.save_config(
+            cmd=cmd, confirm=confirm, confirm_response=confirm_response
+        )
+    else:
+        result = conn.save_config(confirm=confirm)
```
ktbyers: You should have `confirm_response` here also

enzzzy:
I’ve made the requested changes.
Ps test test_print_result.py seems to fail using python3.7.1. The test succeeds with 3.6.5. Haven’t looked into it though.

PR #282, opened 2018-11-22
Add support to pass path to dir of templates that the template extends

omershtivi:


dbarrosop:
CI is broken, you need to lint your code with `black`

omershtivi:
Yes yes saw it will take a try hopefully tonight.
Do you want tests also? And in docs can under how-to I can add templates
how to use them?

nornir/core/helpers/jinja_helper.py
```diff
@@ -11,9 +11,12 @@ def render_from_file(path, template, jinja_filters=None, **kwargs):
     return template.render(**kwargs)
 
 
-def render_from_string(template, jinja_filters=None, **kwargs):
+def render_from_string(template, path=None, jinja_filters=None, **kwargs):
     jinja_filters = jinja_filters or {}
-    env = Environment(undefined=StrictUndefined, trim_blocks=True)
+    if path:
```
dbarrosop: this should be `if path is not None`

nornir/plugins/tasks/text/template_string.py
```diff
@@ -2,13 +2,14 @@
...
 
-def template_string(task, template, jinja_filters=None, **kwargs):
+def template_string(task, template, jinja_filters=None, path=None, **kwargs):
     """
     Renders a string with jinja2. All the host data is available in the tempalte
 
     Arguments:
         template (string): template string
         jinja_filters (dict): jinja filters to enable. Defaults to nornir.config.jinja_filters
+        path (string): path that to the folder that the template extends. Defaults to None
```
dbarrosop: We will need to rephrase this. Something like `path to folder with additional jinja2 templates and resources`.

nornir/core/helpers/jinja_helper.py
omershtivi: Hi,
Nope, it is if path, because if path is none call the function without loader, if it isn't none call the function with Loader

dbarrosop:
tests and docs are always welcome :)

nornir/plugins/tasks/text/template_string.py
omershtivi: So it is actually the path that the templates extends.. so maybe base_template_path? it is too long, so happy for suggestions

nornir/core/helpers/jinja_helper.py
dbarrosop: you need to account for things that are not `None` but might be `falsey`. For instance, if the user passes `path=""`, that's why `if path` should be `if path is not None`

nornir/plugins/tasks/text/template_string.py
dbarrosop: So this enables more than just the `extend` as in your use case. It also enables the imports and includes, it's basically a path to jinja2 assets. In any case, I think it's best to document as `if set, load FileSystemLoader with this path`.

omershtivi: You mean in the comment? sure NP

nornir/core/helpers/jinja_helper.py
omershtivi: Nice! didn't thought about it! thanks!

omershtivi:
Can't understand why CI broke

omershtivi:
Thanks! I will add some docs and tests.

omershtivi:
Code - V
Test to cover the use case - V
Docs how to use the templates - V

I think I'm done let me know if need to add/rephrase/remove something (:

dbarrosop:
Just changed the base branch of the PR as this should be against `2.0.0-beta`. Couple of comments:

1. Needs to be rebased.
2. The how to should be a jupyter notebook. It requires a tiny bit more effort to write but if done properly it serves as tests as well, which means docs are always up to date.

dbarrosop:
Sorry this fell through the cracks. I have been a bit busy preparing 2.0.0 release. Looks like your rebase went a bit wild. Would you mind resubmitting from a clean branch? (use develop as base)

omershtivi:
Will do,
Hopefully next week

dbarrosop:
Are you still planning to work on this?

ktbyers:
I am going to close this. If you are still working on it--please rebase and resubmit thanks.

PR #286, opened 2018-12-16
Release 2.0.0

dbarrosop:


dbarrosop:
Ping @ktbyers @dmfigol @ogenstad

ktbyers:
+1

ogenstad:
I've been absent for so long so this is too much for me to say much about. However, it looks like the project README.md file still says "While Nornir still supports Python 2.7 the recommended version is 3.6. Install it with pip.", would be good to at least scratch that part before a 2.0 release.

PR #292, opened 2018-12-24
Transform function options

### LINKED ISSUE #290: Passing arguments to inventory transform_function
Hi all,

I'm using the `transform_function` in what I think is a fairly canonical way, to pass host authentication data (and other configuration secrets) from a secrets storage system into the inventory data (which initially contains only non-sensitive data) so that nornir can connect to hosts. Because the transform function has never accepted any arguments other than the current `host`, I've done this in the past using function currying, and lexical scoping (example below).

Under nornir v2 this no longer works as the transform_function is serialised to a string, using a custom `cls_to_string` method which doesn't handle anonymous functions/lambdas at all (it naively generates a bogus path name). I would like to find a way to continue to support this in v2+.

Example:
```python
# Reusable function in a library, which isn't concerned with where the auth data comes from
def transform(auth_data):
  def _transform(host):
    if not auth_data:
      return
    # Pass in user/pass for nornir to use during connection
    if 'username' in auth_data:
      host.data['nornir_username'] = auth_data['username']
    if 'password' in auth_data:
      host.data['nornir_password'] = auth_data['password']
    # Store a copy of the auth data for use by other tasks
    host.data['auth_data'] = auth_data
  return _transform # partially curried function

# nornir script
auth_data = retrieve_from_secrets_system(token=os.env['SECRET_TOKEN'])

nr = InitNornir(
    config_file="config.yaml",
    dry_run=True,
    inventory={
        'plugin': 'netops.inventory.foo.FooInventory',
        'transform_function': transform(auth_data), # fully-curry the transform function with the auth data
    }
)
```

Currently fails with:
``` 
File "/data/home/brobert/git/netops/env3/lib64/python3.6/site-packages/nornir/init_nornir.py", line 43, in InitNornir
    conf = Config.load_from_file(config_file, **kwargs)
  File "/data/home/brobert/git/netops/env3/lib64/python3.6/site-packages/nornir/core/deserializer/configuration.py", line 172, in load_from_file
    return Config.deserialize(**{**config_dict, **kwargs})
  File "/data/home/brobert/git/netops/env3/lib64/python3.6/site-packages/nornir/core/deserializer/configuration.py", line 158, in deserialize
    inventory=InventoryConfig.deserialize(**c.inventory.dict()),
  File "/data/home/brobert/git/netops/env3/lib64/python3.6/site-packages/nornir/core/deserializer/configuration.py", line 59, in deserialize
    transform_function=_resolve_import_from_string(inv.transform_function),
  File "/data/home/brobert/git/netops/env3/lib64/python3.6/site-packages/nornir/core/deserializer/configuration.py", line 183, in _resolve_import_from_string
AttributeError: module '__main__' has no attribute '_transform'
```

I understand passing through the anonymous function pointer is no longer going to work because of serialisation/deserialisation. Would you consider adding a `transform_function_options` parameter that would pass through `InventoryConfig` as either a single serialisable value, or as `**kwargs` to permit this use case going forwards?

Happy to work on a PR if so.

Thanks,
Ben

optiz0r:
Adds support, documentation and tests for a new `transform_function_options` inventory configuration option.

Fixes #290

### DIFF
--- a/docs/howto/transforming_inventory_data.ipynb
+++ b/docs/howto/transforming_inventory_data.ipynb
@@ -256,7 +256,44 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "For more information on how inheritance works check the tutorial section [inheritance model](../tutorials/intro/inventory.rst)"
+    "For more information on how inheritance works check the tutorial section [inheritance model](../tutorials/intro/inventory.rst)\n",
+    "\n",
+    "## Passing options to the transform function\n",
+    "\n",
+    "You can also pass options to the transform function, like in this example:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from nornir import InitNornir\n",
+    "\n",
+    "\n",
+    "def adapt_host_data(host, password, something_else):\n",
+    "    host.data[\"password\"] = password\n",
+    "\n",
+    "    \n",
+    "options = {\n",
+    "    \"password\": \"a_secret_password\",\n",
+    "    \"something_else\": \"unused\"\n",
+    "}\n",
+    "    \n",
+    "nr = InitNornir(\n",
+    "    inventory={\n",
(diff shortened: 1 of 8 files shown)


dbarrosop:
Awesome!

PR #293, opened 2018-12-27
Remove the external ping test

optiz0r:
By having a test for external connectivity which isn't stubbed,
what's actually being tested is the network environment of the development
machine, and not the nornir code. There are several examples where
this test would legitimately fail:
- Offline laptop
- Corporate environment with restrictive firewall policies.

As such, a failing test does not necessarily indicate an issue with the
code. A failed test will cause downstream tools that depend on passing
tests to fail, or leave developers wasting time looking for non-existent
faults.

### DIFF
--- a/tests/plugins/tasks/networking/test_tcp_ping.py
+++ b/tests/plugins/tasks/networking/test_tcp_ping.py
@@ -1,7 +1,6 @@
 import os
 
 
-from nornir import InitNornir
 from nornir.plugins.tasks import networking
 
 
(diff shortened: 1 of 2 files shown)


dbarrosop:
I agree, might be nice to mock though. I am going to merge it to avoid unblocking this issue but I think we should basically create a docker-compose file to orchestrate a few containers. I will create an issue for that.

tests/plugins/tasks/networking/test_tcp_ping.py
```diff
@@ -42,18 +41,3 @@ def test_tcp_ping_invalid_ports(self, nornir):
             processed = True
             assert isinstance(result.exception, ValueError)
         assert processed
-
-    def test_tcp_ping_external_hosts(self):
-        external = InitNornir(
-            inventory={"options": {"host_file": ext_inv_file}}, dry_run=True
```
dbarrosop: Is the file needed at all?

Issue #295, opened 2018-12-28
orchestrate tests with docker-compose


dbarrosop:
To avoid issues like #293

PR #296, opened 2018-12-28
isolate test environment with docker-compose

dbarrosop:
Solves #295

This tries to accomplish a few things:

1. Isolate completely the testing environment so everybody has the same regardless of if it's local or in travis (pandoc issues anyone? ;) )
2. Allow orchestrating containers more easily. Right now there is a bash script that starts containers that needs to be instantiated with pytest. With this you can start them and have them running in the background which may allow you to iterate faster.
3. Avoids issues like #293 as we can add local resources and link them. So no more tests pointing to localhost :)
4. Gets rid of tox which I never liked.


dbarrosop:
@ogenstad @dmfigol @ktbyers 

This changes slightly the development workflow (I updated the contributing guidelines) but I think it's worth the change as it will lead to less obscure crap going on in travis.

ogenstad:
Looks reasonable. Have only looked on my phone. Only comment is in wrapper.py. 

If I remember correctly the assert in this code will only show true/false. So if the output doesn’t match the reference the user will only see that something is wrong but not what. I.e., that’s a Python assert not a Pytest assert. 

if screen_output != reference_output:
raise Exception(screen_output, reference_output)
assert screen_output == reference_output

dbarrosop:
> the assert in this code will only show true/false

assert throws an Assertion error, for instance, the REPL:

```
>>> assert 1 == 2
  File "<stdin>", line 1, in <module>
AssertionError
```

ogenstad:
That was my point a user will only see that an error occurred, but not why.

ktbyers:
I tested this in my AWS AMI environment (fedora-based). It was fine. I had to install docker-compose using curl, but everything was pretty straight-forward after that.

+1

PR #298, opened 2018-12-30
Add support for SSH config file via Nornir config for Netmiko and NAPALM

### LINKED ISSUE #294: Netmiko connection plugin is not using ssh_config from Nornir config
Also need to make this work for NAPALM-Netmiko and add appropriate tests.

ktbyers:
Fixes #294


ktbyers:
I tested this in a lab environment for both Netmiko and NAPALM using a Linux proxy server and connecting to Cisco IOS devices (via the intermediate Linux server).

PR #299, opened 2018-12-31
Expand user path to default ssh config

floatingstatic:
After upgrading from v1.1.0 to v2.0.0 any inventory hosts that depend on a "proxycommand" in ssh config files appear to fail connecting when the ssh config file path is not explicitly set in InitNornir ssh config options. After digging into this it appears that the default `~/.ssh/config` setting is not expanded to the full path of the user home directory thus not loading altogether. Looking at how this was handled in v1.1.0 the default was set as follows:
```
    "ssh_config_file": {
        "description": "User ssh_config_file",
        "type": "str",
        "default": os.path.join(os.path.expanduser("~"), ".ssh", "config"),
        "default_doc": "~/.ssh/config",
    },
``` 

This change sets the default setting in the configuration deserializer back to how it was handled in v1.1.0 by using `os.path.expanduser`

### DIFF
--- a/nornir/core/deserializer/configuration.py
+++ b/nornir/core/deserializer/configuration.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+import os
 from typing import Any, Callable, Dict, List, Optional, Type, cast
 
 from nornir.core import configuration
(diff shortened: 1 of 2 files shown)


dbarrosop:
Yeah, this should be a function of the de-serialization process. I opened #304, would you mind testing?

PR #304, opened 2019-01-06
expand user home when deserializing configuration

dbarrosop:


ogenstad:
This would still cause the documentation to become incorrect?

PR #306, opened 2019-01-06
fix pytest

dbarrosop:
Looks like pytest 4.1.0 breaks things...


ogenstad:
Looks like it's an issue with the pytest-cov plugin, which is probably using a deprecated feature.

Issue #291, opened 2018-12-20
Merging of config file and InitNornir params (breaking change relative to v1)


optiz0r:
Hi

I had an issue in v1 where it is not possible to set some Inventory options in the config file, and set other options in the call to `InitNornir`. This is still present in v2 despite the rewrite because of the dict merge here: https://github.com/nornir-automation/nornir/blob/develop/nornir/core/deserializer/configuration.py#L170.

However in v1 it was possible to set options in the config file, and set a `transform_function` in the call to InitNornir because they were separate settings. In v2, transform_function has now been moved to a child of the inventory config, so trying to pass this into the InitNornir call means none of the inventory configuration from the config file is respected any more.

Having to define the `transform_function` in the config file, I think:
- blurs the lines between code and config data separation
- makes it much harder to reuse one config file across multiple nornir-based tools which might need different transforms, but otherwise reuse the inventory plugin options

IMHO `Config.load_from_file` does a partial merge between config data and parameters at too high a level (choosing to take one block at a time from either source). Especially now that the config file has been refactored to make the structures deeper, the merging should happen at least one level deeper than it currently is (it would also be useful if `inventory.options` dict were merged too, but that's not a change in behaviour from v1).

Example change which merges one level below the top, letting kwargs take precedence. This is perhaps not the nicest code (and does not handle merging of the inventory.options), but does restore what I think the previous behaviour was:
```python
    @classmethod
    def load_from_file(cls, config_file: str, **kwargs) -> configuration.Config:
        config_dict: Dict[str, Any] = {}
        if config_file:
            yml = ruamel.yaml.YAML(typ="safe")
            with open(config_file, "r") as f:
                config_dict = yml.load(f) or {}

        # Merge the config and parameters together
        merged_config = config_dict
        for k1 in kwargs:
            if k1 in merged_config:
                for k2 in kwargs[k1]:
                    merged_config[k1][k2] = kwargs[k1][k2]
            else:
                merged_config[k1] = kwargs[k1]

        return Config.deserialize(**merged_config)
```

What do you think?

optiz0r:
if all variables were top-level variables the current merging behaviour would be fine. Because the config is deeply structured and the merging applies to the very top-most level only this is less useful than it could be, and forces a lot of complexity down to the end user code.

Imagine this more concrete example:
```yaml
# config.yaml
inventory:
  plugin: MyInventory
  options:
    # These are sane defaults, which we might want to override at runtime with cli option, or env var
    cache: true
[...]

PR #309, opened 2019-01-06
fix order of preference when deserializing config

dbarrosop:


dbarrosop:
@ktbyers @dmfigol can I get a review here? I'd like to merge and release soon

dmfigol:
@dbarrosop could you briefly describe what this PR is supposed to do so that it is easier to review? I tried following #291 but at some point I gave up.

dbarrosop:
It's basically a bugfix. In the documentation we state that the order of preference of the configuration is `config_file < env variables < code`, however, that wasn't true after the move to pydantic. This fixes the issue and makes the order correct.

Issue #230, opened 2018-08-15
Nornir is hijacking logger


dmfigol:
```
import logging

from nornir.core import InitNornir

logging.basicConfig(
    format="[%(asctime)s] %(levelname)-8s {%(filename)s:%(lineno)d} %(message)s",
    level=logging.INFO,
    handlers=[
        logging.FileHandler("my-app.log"),
        logging.StreamHandler(),
    ],
)

logger = logging.getLogger(__name__)


def main() -> None:
    logger.info("Starting script")
    nr = InitNornir("config.yaml")
    logger.info("Ending script")

main()
```
```
-> % python main.py
[2018-08-15 20:14:29,622] INFO     {main.py:72} Starting script
```
Even after 6 years python logging is still a mystery to me, so I'd greatly appreciate if someone could help with triage and fixing the issue.

dbarrosop:
Probably not the best implementation but the behavior is intended:

https://github.com/nornir-automation/nornir/blob/develop/nornir/core/__init__.py#L110

We should probably call it instead from `InitNornir` and move the function into a helper. Anyway, your best solution is to define a `dictConfig` and add it to the config:

https://github.com/nornir-automation/nornir/blob/develop/nornir/core/__init__.py#L122

Alternatively, all the parameters you are trying to set are actually supported by the configuration so it should be quite straightforward to configure all that.

dmfigol:
I am failing to understand how this behavior is intended.
I can't recall any other Python library that hijacks your root logger and changes its configuration. 
Shouldn't all nornir logs have `nornir` logger as root and then branch either to `nornir.core` or `nornir.plugins`?
Nornir is just part of my application and I am not sure why I would want to additionally configure `nornir` logging using a different way than standard Python logging configuration mechanism.

Currently I have found the following workaround - branch my loggers with different root:
```
import sys
import logging.config

logging_dict = {
    "version": 1,
    "disable_existing_loggers": True,
    "formatters": {
        "standard": {
            "format": "[%(asctime)s] %(levelname)-8s {%(name)s:%(lineno)d} %(message)s"
        }
    },
    "handlers": {
        "default": {
            "level": "INFO",
            "class": "logging.handlers.RotatingFileHandler",
            "filename": "lab-system.log",
            "maxBytes": 1024 * 1024 * 5,
            "backupCount": 5,
            "formatter": "standard",
        },
        "console": {
            "level": "INFO",
            "class": "logging.StreamHandler",
[...]

PR #316, opened 2019-01-20
Improve logging

### LINKED ISSUE #302: Duplicate output in nornir.log file
Hi,

It looks to me that the nornir.log file has duplicate output. Below is some content in my nornir.log file:

```
2019-01-03 16:30:44,584 -  nornir.core -     INFO -        run() - Running task 'netmiko_send_command' with num_workers: 20
2019-01-03 16:30:44,584 -  nornir.core -     INFO -        run() - Running task 'netmiko_send_command' with num_workers: 20
2019-01-03 16:30:44,604 -  nornir.core -    DEBUG -        run() - {'command_string': 'show interfaces status'}
2019-01-03 16:30:44,604 -  nornir.core -    DEBUG -        run() - {'command_string': 'show interfaces status'}
2019-01-03 16:30:44,637 - nornir.core.task -     INFO -      start() - test_sw_01: netmiko_send_command: running task
2019-01-03 16:30:44,637 - nornir.core.task -     INFO -      start() - test_sw_01: netmiko_send_command: running task
2019-01-03 17:16:56,505 -  nornir.core -     INFO -        run() - Running task 'netmiko_send_command' with num_workers: 20
2019-01-03 17:16:56,505 -  nornir.core -     INFO -        run() - Running task 'netmiko_send_command' with num_workers: 20
2019-01-03 17:16:56,524 -  nornir.core -    DEBUG -        run() - {'command_string': 'show interfaces status'}
2019-01-03 17:16:56,524 -  nornir.core -    DEBUG -        run() - {'command_string': 'show interfaces status'}
2019-01-03 17:16:56,597 - nornir.core.task -     INFO -      start() - test_sw_02: netmiko_send_command: running task
2019-01-03 17:16:56,597 - nornir.core.task -     INFO -      start() - test_sw_02: netmiko_send_command: running task
```

Is it really an issue, or did I set something wrong in my Nornir configuration?

Regards,

dmfigol:
Fix #302, #230

* Better interoperability with native Python logging module
* No more duplicate logs
* Replace format and f-strings in logs to % strings according to best practices
* Improve some logging messages
* `InitNornir` `configure_logging` parameter is deprecated, replaced with `config.logging.enabled`

TODO:
* [x] Write a separate logging section in docs


Makefile
```diff
@@ -34,7 +34,7 @@ pytest:
 .PHONY: black
 black:
 	${DOCKER_COMPOSE} \
-		run nornir black --check .
```
dbarrosop: why these changes?

nornir/core/__init__.py
```diff
@@ -118,12 +115,16 @@ def run(
...
                     run_on.append(host)
 
-        self.logger.info(
-            "Running task '{}' with num_workers: {}".format(
-                kwargs.get("name") or task.__name__, num_workers
+        num_hosts = len(self.inventory.hosts)
+        task_name = kwargs.get("name") or task.__name__
+        if num_hosts:
+            logger.info(
+                "Running task %r with args %s on %d hosts", task_name, kwargs, num_hosts
```
dbarrosop: use f-string, please

nornir/core/__init__.py
```diff
@@ -118,12 +115,16 @@ def run(
...
+            )
+        else:
+            logger.warning(
+                "Task %r has not been run due to 0 hosts selected", task_name
```
dbarrosop: f-string, please. Mostly for consistency.

nornir/core/deserializer/configuration.py
```diff
@@ -65,30 +65,32 @@ def deserialize(cls, **kwargs) -> configuration.InventoryConfig:
...
+    file_path: str = Schema(default="nornir.log", description="Logging file")
     format: str = Schema(
-        default="%(asctime)s - %(name)12s - %(levelname)8s - %(funcName)10s() - %(message)s",
+        default="[%(asctime)s] %(levelname)-8s {%(name)s:%(lineno)d} %(message)s",
         description="Logging format",
     )
     to_console: bool = Schema(
         default=False, description="Whether to log to console or not"
     )
-    loggers: List[str] = Schema(default=["nornir"], description="Loggers to configure")
```
dbarrosop: I don't think it's a widely used feature but the idea was to allow users to enable debugging in paramiko/netmiko/others without having to change the code. How do you envision that being done now and why was it removed?

nornir/core/deserializer/configuration.py
```diff
@@ -185,6 +187,6 @@ def _resolve_import_from_string(import_path: Any) -> Optional[Callable[..., Any]
         module_name, obj_name = import_path.rsplit(".", 1)
         module = importlib.import_module(module_name)
         return getattr(module, obj_name)
-    except Exception as e:
-        logger.error(f"failed to load import_path '{import_path}'\n{e}")
+    except Exception:
+        logger.error("Failed to import %r", import_path, exc_info=True)
```
dbarrosop: I'd like to use f-strings everywhere as they are easier to read and execution time they are identical

nornir/plugins/inventory/ansible.py
```diff
@@ -251,9 +251,7 @@ def parse(hostsfile: str) -> Tuple[HostsDict, GroupsDict, DefaultsDict]:
         try:
             parser = YAMLParser(hostsfile)
         except (ScannerError, ComposerError):
-            logger.error(
-                "couldn't parse '{}' as neither a ini nor yaml file".format(hostsfile)
-            )
+            logger.error("couldn't parse %r as neither a ini nor yaml file", hostsfile)
```
dbarrosop: f-string

nornir/core/deserializer/configuration.py
```diff
@@ -65,30 +65,32 @@ def deserialize(cls, **kwargs) -> configuration.InventoryConfig:
...
-    level: str = Schema(default="debug", description="Logging level")
-    file: str = Schema(default="nornir.log", descritpion="Logging file")
+    enabled: bool = Schema(
+        default=True, description="Whether to configure logging or not"
+    )
+    level: str = Schema(default="DEBUG", description="Logging level")
+    file_path: str = Schema(default="nornir.log", description="Logging file")
     format: str = Schema(
-        default="%(asctime)s - %(name)12s - %(levelname)8s - %(funcName)10s() - %(message)s",
+        default="[%(asctime)s] %(levelname)-8s {%(name)s:%(lineno)d} %(message)s",
```
dbarrosop: breaking change, might break people's logging system

dbarrosop:
> Replace format and f-strings in logs to % strings according to best practices

which best practices?

There is a third breaking change, which is the format of the log.

dbarrosop:
I am a bit concerned about this PR introducing so many breaking changes. I think we need to either have very good reasons for them, revert them or delay this to 3.0.0.

@ktbyers thoughts?

dbarrosop:
Ok, read a bit more about it and looks like the "recommended" method is just because it's a lazy evaluator for the string. I don't think we should be concerned about that as that's going to save us 0.0000000001us of execution time and I'd rather have readable string but to avoid bikeshedding I will let ktbyers chime in and break the tie :)

Makefile
dmfigol: The current dockerfile is built in such a way that, it is mapping the whole project dir to the container, which causes the problems for my setup when I run tests... Decided to change to include directories explicitly when black should run.  
I can change it back, but I'd strongly advocate to change volume mounting in docker-compose then.

nornir/core/deserializer/configuration.py
dmfigol: How is it breaking? It's just the format of logs.

nornir/core/deserializer/configuration.py
dmfigol: I remove it, because I didn't really understand what problem it solves. If you want to enable logging for other modules, you can do it with standard python logging (basicConfig, dictConfig or programmatically). I can put it back if you want, but I'd vote against it.

ktbyers:
I vote f-strings versus the % strings.

I also vote if they are backwards incompatible that we defer until Nornir 3.0 (or maybe some way to specify new-logging format in Nornir-config that then becomes the default in Nornir 3.0)? Disclaimer: I have mostly been out of the loop on this and only loosely following it.

dmfigol:
Those are not really % strings as % strings are interpolated at declaration time, the syntax is `"%s", %var` (difference: comma). f-strings in logs are not "best practice" and both pylint and pycharm will yell at you unless suppressed. But I also agree that they increase readability so I guess I will change it if my opinion didn't consensus ;)
I am honestly not sure about "breaking changes" and 3.0:
1) file -> file_path, we can add warning to "file" being set to something other `nornir.log`, still use it, but `file_path` would take precedence
2) we can add "loggers" back, but honestly I think they are useless. I could add a warning that they are not used anymore. I could make them work back. My take in general is that it is not nornir's job to touch logging. So if we are touching it already, the impact should be minimum. Built-in Python logging module is very powerful.
3) I don't consider new logging format a breaking change at all.

ktbyers:
Okay, so it really is a tuple of arguments that is passed to logger:

```
>>> a = "foo"
>>> b = "bar"
>>> "Running task %s whatever %s", a, b
('Running task %s whatever %s', 'foo', 'bar')
```

Here is more discussion on pylint doing this (wasn't very well received):

https://github.com/PyCQA/pylint/issues/1788

Looks like we can just disable that warning in Pylint.

Makefile
dbarrosop: if that's the case let's discuss it in a separate PR to avoid noise

nornir/core/deserializer/configuration.py
dbarrosop: Well, the point is that a developer builds a tool, deploys it in a management server and then the neteng folks (or other people) consume that tool, which is installed in the $PATH and don't have access to modify it, how would you enable netmiko's debug then because something is not going as expected? You could have the developer bake it in into its own application, and that'd be a valid answer, however, if there isn't a good reason to remove it at this point (like introducing an insurmountable amount of complexity) I'd gate it until nornir 3.0.0.

dbarrosop:
> file -> file_path

Why that change? `file` is not masking any actual global variable.

> so I guess I will change it if my opinion didn't get consensus

I still think readability wins over this supposedly perf improvement, specially when the perf improvements is going to be measured in microseconds while nornir's application are mostly going to be bound by IO and measured in ms. So whatever kirk says as he is the one breaking the tie :)

> but honestly I think they are useless

I already commented on the reasoning, and I am fine deprecating it in nornir 3 if we decide the benefits don't overweight the complexity but without a very good reason we shouldn't introduce breaking changes as per semver rules

> My take in general is that it is not nornir's job to touch logging

I agree, notice it's not nornir who is doing that, it's `InitNornir`, which is a convenient entrypoint to the framework for people that want convenience. If you want full control you can implement your own InitNornir and instantiate the objects yourself.

> I don't consider new logging format a breaking change at all.

For people sending logs to a logging system and parsing the logs to build dashboards it's going to be breaking change.

requirements.txt
```diff
@@ -7,4 +7,4 @@ future
 requests
 ruamel.yaml
 mypy_extensions
-pydantic
+pydantic==0.16.1
```
dbarrosop: don't forget to rebase as new develop uses poetry :)

nornir/core/deserializer/configuration.py
dmfigol: You are saying that a neteng would change env var in this case? I don't think it will work for lists (may be I wrong here). Or you are saying that a neteng would need to change `config.yaml`? If that is the case, how different it is from changing the logging setup in the entrypoint script, if both `config.yaml` and the application code will be in VCS?
I do think you are onto something here and may be it is a valid use-case, however it needs to be described in our docs, because until you told it, I didn't even think about it.

dmfigol:
> Why that change? file is not masking any actual global variable.

To make it clear that it is not filename it is a path. I will return it back.

> I agree, notice it's not nornir who is doing that, it's InitNornir, which is a convenient entrypoint to the framework for people that want convenience. If you want full control you can implement your own InitNornir and instantiate the objects yourself.

We have every single example with `InitNornir`, we don't even discuss (by discuss I mean docs) the possibility of creating `Nornir` objects using some other mechanism.

> For people sending logs to a logging system and parsing the logs to build dashboards it's going to be breaking change.

The current setup does not add Syslog handler at all... Do you want me to remove it back anyway?
I guess for my projects I will always do `logging.enabled = False`.

Something that I didn't see in your comments so far. Could you please go through 4 conditions when nornir logging is not configured and confirm that they are all right? I think first two completely make sense, but for the last two where I decide based on the root logger config there can be some disagreement.

dmfigol:
> So let's recover the old InitNornir flag and let the user decide instead, please. It's also a breaking change.

I can find a way to convert it to a non-breaking change

> I don't want to add any logic here. I don't want to have to document under which circumstances the logging subsystem is configured automagically or have to answer support questions about that

Then what does this PR solve? One of my problems with `nornir` is that if you do Python logging (using `basicConfig` or `dictConfig`) and you don't disable `nornir` logging, the result will be a mess. It is not documented properly either. One of the things this PR solves is that if you do that, the end result will NOT be a mess.

dbarrosop:
> If I remove that, what does this PR solve?

My understanding is that, besides the breaking changes, you are doing some general cleaning and improvements as well. For instance, getting rid of the duplicates.

> if you do Python logging (using basicConfig or dictConfig) and you don't disable nornir logging, the result will be a mess

I don't see that as a problem. You have the option today of using either way, we are not forcing anyone to use nornir's logging helper function.

> It is not documented properly either

If it's not properly documented let's fix that.

> One of the things this PR solves is that if you do that, it will not result in a mess

In general, instead of taking arbitrary decisions (which the user may or may not want), I'd suggest logging warning.

bradh11:
Does this PR also address the duplicate log entries in the logfile that are poduced??

dmfigol:
@bradh11 yes

dbarrosop:
@dmfigol let us know when this is ready for review again. If we get this one sorted out this week I will release 2.1.

dmfigol:
@dmfigol @ktbyers this is ready for review

dbarrosop:
I will try to find some time during the weekend. Thanks for the work :)

docs/howto/handling_connections.ipynb
```diff
@@ -73,34 +73,9 @@
   },
```
dbarrosop: did this break?

nornir/core/deserializer/configuration.py
```diff
@@ -65,8 +65,11 @@ def deserialize(cls, **kwargs) -> configuration.InventoryConfig:
 
 
 class LoggingConfig(BaseSettings):
-    level: str = Schema(default="debug", description="Logging level")
-    file: str = Schema(default="nornir.log", descritpion="Logging file")
+    enabled: Optional[bool] = Schema(
+        default=None, description="Whether to configure logging or not"
```
dbarrosop: shouldn't this be true to avoid changing default behavior?

dbarrosop: ok, I see what you are doing down below.

dbarrosop:
Ok, this looks awesome to me. The only comment is related to the notebook that seems to have lost the output. Can you confirm that's expected? Feels broken to me. Once we clear that out I will merge this :)

Awesome work.

dmfigol:
@dbarrosop 
I've rebased and fixed the notebook. You were right, for some reason this notebook didn't have an output and was not checked by `nbval`. I've stumbled upon an issue with `\r\n` and `\n`, which I didn't find the root cause for, so I had to use nbval sanitization, to make sure that `\r\n` is converted to `\n` everywhere.

dbarrosop:
> so I had to use nbval sanitization

Good catch :)

PR #323, opened 2019-02-17
Fix/319

bradh11:
seeing if proposed fix to remove the colorama init(convert) method will result in passed unit tests.


bradh11:
ok, unit tests have passed now.  I think this resolves the conflict we were seeing between ansii color in jupyter notebook (being used for the docs), and windows terminals.  please feel free to do a few more manual spot checks before merging into Develop.

ktbyers:
Okay, that fix makes sense to me:

The colorama docs state this:

```
init(convert=None):
Pass True or False to override whether to convert ANSI codes in the output into win32 calls. The default behaviour is to convert if on Windows and output is to a tty (terminal).
```

So it sounds like this convert has three modes...the default which is None does a convert but only on windows, but not elsewhere.

bradh11:
@ktbyers  I agree - 
I found the same reference in my research about the default for windows being `None`.  I suppose the other option we have is to allow a toggle in nornir init method to allow user override in case of environmental conflicts.  I think however, for now this solution seems like it fits the most scenarios.  works on windows, 'nix/Mac, and jupyter based environments.

PR #324, opened 2019-02-17
Add gitlab file plugin

enzzzy:
I have been using this plugin mainly for backing up device configurations and saving of staged configurations in a git repo on Gitlab. 
Thought it might be useful for other people, so I'm sharing this. 

A quick example to show how this can be used
```
from nornir.plugins.tasks.networking import netmiko_send_command
from nornir.plugins.tasks.files.gitlab import gitlab

def backup(task):
    res = task.run(
        task = netmiko_send_command,
        name = "grab config",
        command_string="show running-config"
    )
    task.host["config"] = res.result

    res = task.run(
        task = gitlab,
        action = "create",
        url = "https://gitlab.local",
        token = "mygitlabapitoken",
        respository = "backup",
        branch = "master",
        content = task.host["config"],
        filename = task.host.name,
        commit_message = "backup"
    )
```


dbarrosop:
Hello,
I have a couple of questions. I am not familiar with gitlab but my understanding is that it's a git repository, isn't it? If yes, why not a generic set of functions to deal with git repositories that can be reused?

Regards
David

dbarrosop:
By the way, nice work. This is something I have been wanting for a while now :) I just want to discuss the possibility of generalizing this so it's not gitlab specific :)

enzzzy:
Yes, it is a git repository and I agree that there is a need to have a generic git plugin that is not dependend on Gitlab or others. 

But in my usecase, we are deploying nornir in multiple small containers, let’s say they are workers. I don’t want to have the full git toolset in those containers and I also don’t want to have the full repository pulled into the container. 

That’s why interfacing with gitlab through its API made a lot of sense to me.

I understand that it is very specific for my usecase. If someone else has a similar requirements but uses a different git provider, then they can’t reuse this.

So to conclude:
Yes I would love to have a generic implementation , which doesn’t depend on a specific provider. But I also think there are usecases for specific implementations for different git providers, if they provide an API to interface with them.

dbarrosop:
ok, finally found some time to look into this a bit more in depth. A few comments:

1. Good job, specially in the testing part
2. I think we should move the plugin to a dedicated section for `source_control` plugins. Although I am not 100% sure. I was thinking in `tasks.plugins.source_control.gitlab`. Thoughts?
3. Regarding documentation, I was wondering if you could rewrite it using a jupyter notebook. I am trying to use notebooks as much as possible for documentation as they are more visual and can be tested. For instance:

https://github.com/nornir-automation/nornir/blob/76247d3684ff06e2f2b43908bc9bf838ee2badaf/nornir/plugins/tasks/data/echo_data.py
https://github.com/nornir-automation/nornir/blob/344fa4ead1edaec8f554323e267993f745ca701a/docs/plugins/tasks/data/echo_data.ipynb

which renders:

https://nornir.readthedocs.io/en/stable/plugins/tasks/data/echo_data.html

and is tested:

https://travis-ci.org/nornir-automation/nornir/jobs/498697369#L553

Not sure how familiar you are with notebooks but happy to help out. If you don't have the time or the energy we can proceed as it is, not a big deal.

enzzzy:
The dedicated section definitely makes sense to me, but personally I would name it version_control. I think that name covers the package a bit better no? It can be used for more then 'source'.
I'm not really familiar with jupyter notebooks yet, but this might be a good way to start learning about it. I'll have a look.

dbarrosop:
> I would name it version_control

Yes, I think that's better too.

enzzzy:
Moved plugin to version_control directory & created documentation.
Still working on a "get" action to retrieve file(s) from a repository ( + test(s) & doc).

dbarrosop:
Awesome work so far! Let me know when you are done :)

enzzzy:
I'm done.
But the build is failing due to a strict implementation of PEP517 from pip v19
Basically pip doesn't allow a editable install if a pyproject.toml file is present.

More info can be found here:
https://github.com/pypa/pip/issues/6434

There seem to be multiple issues opened for this with poetry as well:
https://github.com/sdispater/poetry/issues/1049

dbarrosop:
Impressive job!

Looks like the issue with poetry is solved :)

PR #326, opened 2019-02-24
Bump jsonschema to ^3

optiz0r:
Previous requirement on 2.6.0 causes poetry commands to fail
with error:

    [AttributeError]
    module 'jsonschema' has no attribute 'Draft7Validator'

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -253,7 +253,13 @@ description = "An implementation of JSON Schema validation for Python"
 name = "jsonschema"
 optional = false
 python-versions = "*"
-version = "2.6.0"
+version = "3.0.0"
+
+[package.dependencies]
+attrs = ">=17.4.0"
+pyrsistent = ">=0.14.0"
+setuptools = "*"
+six = ">=1.11.0"
 
 [[package]]
 category = "main"
@@ -449,6 +455,7 @@ textfsm = "*"
 reference = "9c0996ddc3d2de18242bf1a722188bd5aa4c8983"
 type = "git"
 url = "https://github.com/ktbyers/netmiko"
+
 [[package]]
 category = "main"
 description = "A library for managing Cisco devices through NX-API using XML or jsonrpc."
@@ -658,6 +665,17 @@ cffi = ">=1.4.1"
 six = "*"
 
 [[package]]
+category = "dev"
+description = "Persistent/Functional/Immutable data structures"
+name = "pyrsistent"
+optional = false
+python-versions = "*"
+version = "0.14.11"
+
+[package.dependencies]
+six = "*"
+
+[[package]]
 category = "main"
 description = "Python Serial Port Extension"
 name = "pyserial"
@@ -856,7 +874,7 @@ python-versions = "*"
 version = "0.1.7"
 
 [metadata]
-content-hash = "ad8c5de57301b5d529395ec487563c7bac46db9fabfb549331c319a518b40989"
+content-hash = "8174af0974174a191774604c5b1e19db91f1891554a18875da5287330367cab2"
 python-versions = ">= 3.6, < 3.8"
 
 [metadata.hashes]
@@ -885,7 +903,7 @@ ipython = ["6a9496209b76463f1dec126ab928919aaf1f55b38beb9219af3fe202f6bbdd12", "
 ipython-genutils = ["72dd37233799e619666c9f639a9da83c34013a73e8bbc79a7a6348d93c61fab8", "eb2e116e75ecef9d4d228fdc66af54269afa26ab4463042e33785b887c628ba8"]
 jedi = ["571702b5bd167911fe9036e5039ba67f820d6502832285cde8c881ab2b2149fd", "c8481b5e59d34a5c7c42e98f6625e633f6ef59353abea6437472c7ec2093f191"]
 jinja2 = ["74c935a1b8bb9a3947c50a54766a969d4846290e1e788ea44c1392163723c3bd", "f84be1bb0040caca4cea721fcbbbbd61f9be9464ca236387158b0feea01914a4"]
-jsonschema = ["000e68abd33c972a5248544925a0cae7d1125f9bf6c58280d37546b946769a08", "6ff5f3180870836cae40f06fa10419f557208175f13ad7bc26caa77beb1f6e02"]
+jsonschema = ["acc8a90c31d11060516cfd0b414b9f8bcf4bc691b21f0f786ea57dd5255c79db", "dd3f8ecb1b52d94d45eedb67cb86cac57b94ded562c5d98f63719e55ce58557b"]
 junos-eznc = ["b96bccd4ce9c9127d91ca6145b868ae53a378186455eb087d57420d3dae276bd", "d97d8babf650abca25a096825aa6d88573d340481a0b0793afcdac4a7bee09d3"]
 jupyter-client = ["b5f9cb06105c1d2d30719db5ffb3ea67da60919fb68deaefa583deccd8813551", "c44411eb1463ed77548bc2d5ec0d744c9b81c4a542d9637c7a52824e2121b987"]
 jupyter-core = ["927d713ffa616ea11972534411544589976b2493fc7e09ad946e010aa7eb9970", "ba70754aa680300306c699790128f6fbd8c306ee5927976cbe48adacf240c0b7"]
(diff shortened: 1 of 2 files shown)


dmfigol:
I don't think we use `jsonschema` anywhere, that's why it should not be in `pyproject.toml`, which describes direct dependencies.
Could you please clarify?

optiz0r:
Can you clarify, this should be listed in poetry.lock only, but not pyproject.toml? Presently, poetry insists v2.6.0 be installed, which then renders poetry unusable. `pip install --upgrade jsonschema` resolves it temporarily, but poetry will revert back to 2.6.0 on any action. This commit is the result of `poetry add --dev jsonschema=^3` (the `--dev` is only to minimise the diff against the current poetry.lock file, where this is listed as a dev dependency, though I think it's really a runtime dependency from pydantic?)

dbarrosop:
> Previous requirement on 2.6.0 causes poetry commands to fail

Would you mind clarifying that statement, please? Mostly trying to reproduce as I haven't seen that error before.

What @dmfigol is trying to explain is that we shouldn't have to deal in `pyproject.toml` with dependencies that are not directly to the project. (jsonschema is an indirect dependency and should be handled by whichever library is using it).

optiz0r:
I totally get your point that this is an implicit dependency. Without this change, I can't actually use poetry to manage dependencies. It may not be "correct", but working is better than not-working?

A log, captured on Fedora:
```bash
cd ~/git/nornir
python -m venv .env3
source .env3/bin/activate
pip install poetry
poetry install
```

For reference, poetry install does the following:
```
[~/git/nornir] [.env3] develop+ ± poetry install
Installing dependencies from lock file


Package operations: 72 installs, 1 update, 0 removals

  - Installing pycparser (2.19)
  - Installing asn1crypto (0.24.0)
  - Installing cffi (1.12.1)
  - Installing bcrypt (3.1.6)
  - Installing cryptography (2.5)
  - Installing decorator (4.3.2)
  - Installing ipython-genutils (0.2.0)
  - Installing pyasn1 (0.4.5)
  - Installing pynacl (1.3.0)
  - Installing paramiko (2.4.2)
  - Installing parso (0.3.4)
  - Installing ptyprocess (0.6.0)
  - Installing traitlets (4.3.2)
  - Installing wcwidth (0.1.7)
  - Installing backcall (0.1.0)
  - Installing colorama (0.4.1)
  - Installing jedi (0.13.2)
  - Installing jupyter-core (4.4.0)
  - Installing lxml (4.3.1)
  - Installing markupsafe (1.1.0)
  - Installing pexpect (4.6.0)
  - Installing pickleshare (0.7.5)
  - Installing prompt-toolkit (2.0.8)
  - Installing pygments (2.3.1)
  - Installing pyserial (3.4)
  - Installing python-dateutil (2.8.0)
  - Installing pyyaml (3.13)
  - Installing pyzmq (17.1.2)
  - Installing scp (0.13.0)
  - Installing selectors2 (2.0.1)
  - Installing textfsm (0.4.1)
  - Installing tornado (5.1.1)
  - Installing atomicwrites (1.3.0)
  - Installing future (0.17.1)
  - Installing ipython (7.2.0)
  - Installing jinja2 (2.10)
  - Updating jsonschema (3.0.0 -> 2.6.0)
  - Installing jupyter-client (5.2.4)
  - Installing more-itertools (6.0.0)
  - Installing ncclient (0.6.3)
  - Installing netaddr (0.7.19)
  - Installing netmiko (2.3.1 9c0996d)
  - Installing pluggy (0.8.1)
  - Installing py (1.7.0)
  - Installing snowballstemmer (1.2.1)
  - Installing appdirs (1.4.3)
  - Installing click (7.0)
  - Installing coverage (4.5.2)
  - Installing dataclasses (0.6)
  - Installing ipykernel (5.1.0)
  - Installing junos-eznc (2.2.0)
  - Installing mccabe (0.6.1)
  - Installing mypy-extensions (0.4.1)
  - Installing nbformat (4.4.0)
  - Installing nxapi-plumbing (0.5.2)
  - Installing pycodestyle (2.5.0)
  - Installing pydocstyle (3.0.0)
  - Installing pyeapi (0.8.2)
  - Installing pyflakes (2.1.0)
  - Installing pyiosxr (0.53)
  - Installing pytest (4.2.1)
  - Installing toml (0.10.0)
  - Installing typed-ast (1.3.1)
  - Installing black (18.9b0)
  - Installing flake8-import-order (0.18)
  - Installing mypy (0.670)
  - Installing napalm (2.4.0)
  - Installing nbval (0.9.1)
  - Installing pydantic (0.17)
  - Installing pylama (7.6.6)
  - Installing pytest-cov (2.6.1)
  - Installing requests-mock (1.5.2)
  - Installing ruamel.yaml (0.15.88)
  - Installing nornir (2.0.0)
```

Now, any subsequent poetry commands yield the following:
```
[~/git/nornir] [.env3] develop+ ± poetry install

[AttributeError]
module 'jsonschema' has no attribute 'Draft7Validator'

install [--no-dev] [--dry-run] [-E|--extras EXTRAS] [--develop DEVELOP]

[~/git/nornir] [.env3] develop+ 1 ± poetry show

[AttributeError]
module 'jsonschema' has no attribute 'Draft7Validator'

show [--no-dev] [-t|--tree] [-l|--latest] [-o|--outdated] [-a|--all] [--] [<package>]

[~/git/nornir] [.env3] develop+ 1 ± poetry add something

[AttributeError]
module 'jsonschema' has no attribute 'Draft7Validator'

add [-D|--dev] [--git GIT] [--path PATH] [-E|--extras EXTRAS] [--optional] [--python PYTHON] [--platform PLATFORM] [--allow-prereleases] [--dry-run] [--] <name> (<name>)...
```

Upgrading jsonschema fixes this, but of course later running poetry install command downgrades jsonschema to match the lock file, which breaks it again.
```
pip install --upgrade jsonschema
[~/git/nornir] [.env3] develop+ ± poetry debug:info

Poetry
======

 * Version: 0.12.11
 * Python:  3.6.5


Virtualenv
==========

 * Python:         3.6.5
 * Implementation: CPython
 * Path:           /data/home/brobert/git/nornir/.env3
 * Valid:          True


System
======

 * Platform: linux
 * OS:       posix
 * Python:   /usr

[~/git/nornir] [.env3] develop+ ± poetry install
Installing dependencies from lock file


Package operations: 0 installs, 1 update, 0 removals

  - Updating jsonschema (3.0.0 -> 2.6.0)
  - Installing nornir (2.0.0)
[~/git/nornir] [.env3] develop+ ± poetry install

[AttributeError]
module 'jsonschema' has no attribute 'Draft7Validator'

install [--no-dev] [--dry-run] [-E|--extras EXTRAS] [--develop DEVELOP]
```

`poetry debug:resolve -vvv` suggests the dependency might be coming from `nbformat`: 
``` nbformat (4.4.0) depends on jsonschema (>=2.4,<2.5.0 || >2.5.0)```

Seems related to https://github.com/sdispater/poetry/issues/532 (unresolved at time of writing)

So where should this be solved?
nbformat doesn't depend on 3.0.0. Poetry doesn't depend on jsonschema. The only place that uses both nbformat/jsonschema/poetry in combination is nornir, so nornir is the only place that can realistically set this dependency on ^3. Open to counter arguments here...

As an aside, is it right that `poetry install` seems to be trying to install `nornir==2.0.0` every time it runs?

**Comment by ktbyers** [comment]
@optiz0r Do you run into the issue if you install Poetry without using pip?

```
curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python
```

The above is the canonical way for installing poetry.

You can get into conflicts between Poetry's dependencies and the dependencies of the libraries Poetry installs if you use Pip. I had this same 'Draft7Validator/jsonschema' error at one point.

**Comment by optiz0r** [comment]
Issue is not present if poetry is installed that way. That said, eww to the `curl | interpreter` pattern in 2019.

**Comment by ktbyers** [comment]
Okay, so we can close this issue correct?

Once Poetry is installed in the proper way (i.e. not using pip), then jsonschema 3.0.0 is no longer required.

**Comment by ktbyers** [comment]
For reference, here is what the Poetry readme says on this:

```
Using pip to install poetry is also possible.

pip install --user poetry

Be aware, however, that it will also install poetry's dependencies which 
might cause conflicts.
```

From https://github.com/sdispater/poetry#installation

**Comment by dbarrosop** [comment]
> eww to the `curl | interpreter` pattern in 2019

well, it makes sense IMO in this case as you are installing a tool to manage python. You probably installed pip via apt or yum or similar (assuming a linux machine), poetry falls under the same category as pip.

**Comment by optiz0r** [comment]
I would still like to vote for inclusion of this change on the basis that:
- get-poetry.py is not suitable for installation on our production systems without a whole lot of extra effort
- jsonschema is already an implicit dependency, the issue is a combination of versions of software tools so a fix cannot be made anywhere but this project. If there's a time in the future when a nornir dependency implicitly depends on jsonschema 3+, it could be removed.
- It costs this project _very_ little to include this change, but brings benefit, and the work has already been done

**Comment by dbarrosop** [comment]
I'd rather fix the root cause of the problem rather than manage dependencies we don't care about directly. I have created #332 to discuss this. Let me know if I understood your scenario correctly and if you agree that's a better solution.

**Comment by optiz0r** [comment]
What you propose in #332 WFM. Feel free to leave this closed. I'll rebase this commit out of #329 so that can be merged independently.

**Comment by dbarrosop** [comment]
thanks!

**Comment by optiz0r** [comment]
For the sake of others finding this issue, get-poetry.py doesn't work on systems where `python` resolves to python2, and python3.6+ is a separate binary (see https://github.com/sdispater/poetry/issues/721, unresolved at time of writing). In order to use poetry on such systems, must install the following way:

```bash
curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python3
```

And then use it as follows:
```bash
python3 ~/.poetry/bin/poetry [commands]
```

Their suggested workaround is to install using pip, which doesn't work in this case because of the jsonschema conflict. Packaging is hard.

dmfigol:
@optiz0r Another way is to use `python -m pip install --user poetry` outside of virtual environment. This will install poetry dependencies in your home, but not inside virtual environment.

optiz0r:
@dmfigol I concede, `pip3 install --user poetry` is doing the job.

PR #327, opened 2019-02-24
Add deepmerge

optiz0r:
Required for upcoming PR to support merging config files, env
vars and passed parameters when initialising Nornir

Depends on PR #326 
Relates to issue #291


dbarrosop:
If a PR needs dependencies add them there, please.

PR #328, opened 2019-02-24
Update pydantic

optiz0r:
Required for upcoming PR. Bumps pydantic to a version which includes
the settings priority changes required for deep merging configurations.

Depends on PR #327 
Relates to issue #291


pyproject.toml
```diff
@@ -31,6 +32,7 @@ flake8-import-order = "*"
 requests-mock = "*"
 black = {version = "^18.3-alpha.0",allows-prereleases = true}
 mypy = "*"
+jsonschema = "^3"
```
dmfigol: the same comment as in #326, I don't think `jsonschema` is one of our direct dependencies

optiz0r: This PR is sequenced behind #326 and #327, but github doesn't give me a way to show that. The only commit worthy of review in this PR is 2c65b35.

#327, #328, #329 are a set of three, but your contributing guidance says I can't update dependencies in the same PR, hence breaking them up into different ones. #326 is independent and not a dependency of these, but I didn't fancy having divergent poetry.lock content in different PRs. if you decline to merge #326, I'll rebase the other PRs to remove it.

dbarrosop: Updating dependencies should be done in a different PR, that's correct. Adding new ones should be handled in the PR that requires it so the dependency can be evaluated in the given context.

dbarrosop:
I am closing this PR, I think having this in #329 is more valuable. I know it goes against the CONTRIBUTING guidelines but I think this should be an exception. Will try to find some time to update them to reflect this case. If someone beats me to it I will appreaciate it :)

Sorry for the confusion and the extra work.

PR #329, opened 2019-02-24
Fix291 deep merge

optiz0r:
This commit makes it possible to have non-overlapping and overlapping
settings present in all of configuration file, environment variables
and passed to `InitNornir` as kwargs, with all three merged with
respect to the precedence order of:
`hardcoded defaults < config file < env vars < kwargs`.

Previous to this commit:
- Env vars would only be considered if the setting was not present
  in the config file at all
- If any kwarg was set, providing a top-level category of configuration
  (e.g. `core`, `inventory`, `logging`, etc) then that kwarg would be
  taken exclusively, and any env vars or config file settings would
  be disregarded.

To maintain backwards compatibility with the 2.0 behaviour, this new
functionality is opt-in by passing `deep_merge=True` to `InitNornir`
or to the `Config.load_from_file` functions. The default behaviour is
to use the kwarg exclusively if provided, and take the merger of
config file and env var settings otherwise.

Includes tests and documentation.

Depends on PR #328 
Fixes for #291


optiz0r:
Rebased/squashed to include the changes in #327 and #328.

optiz0r:
Should be good to go now, finally...

dbarrosop:
Awesome, will try to find some time during the weekend to play with this. I mostly want to abuse it a bit to make sure we are not changing the default behavior.

optiz0r:
Thanks! I tried to capture what I thought were all the cases in the tests, but if you find a combination which breaks it, let me know and I'll add more tests for them.

dbarrosop:
I opened #336 as merging #309 showed this PR required some more work.

I think I need @ktbyers and @dmfigol to help out with this. I am a bit concerned about the use case and the added complexity.

Let me try to recap what we we can do today and what this PR is trying to do.

### today

The configuration has several sections:

* logging
* inventory
* jinja2
* core
* ssh

Those sections have different paremters you can set. For instance, the inventory configuration has:

* plugin
* options:
    * `**kwargs` (will depend on the plugin)
* transform_function
* transform_function_options

You can configure each one of those options via either a yaml file, via env variables or directly with code. The order of preference is "yaml < env < code" and you have some leeway to mix those methods. For instance, you could do something like this:

``` yaml
# config.yaml
inventory:
  plugin: path.to.a_plugin_for_dev
  options:
       some_option: some_value
  transform_function: path.to.some_function_goes_here
```

```
$ NORNIR_INVENTORY_PLUGIN=path.to.a_plugin_for_prod NORNIR_INVENTORY_PLUGIN={} ./my_script.py
```

And this would result in:

```
# result after configuration is loaded from both the file and the env
inventory:
  plugin: path.to.a_plugin_for_prod
  options: {}
  transform_function: path.to.some_function_goes_here
```

This gives a lot of flexibility already to work with multiple environments without too much effort.

### This PR

This PR adds a mechanism to change this behavior and, instead of replacing the variables, merge them. Using the previous example, the result would be:

```
# result after configuration is loaded from both the file and the env with this PR
inventory:
  plugin: path.to.a_plugin_for_prod
  options:
       some_option: some_value
  transform_function: path.to.some_function_goes_here
```

Looking into other tools I use with similar functionality (i.e. golang's viper or django settings), none allow you to do that.

The problem is basically that:

1. I think it adds a lot of complexity on something that is already more complex than I'd like to.
2. It adds an external dependency.
3. It feels to me this is very specific to your environment and I guess you are using this for your own user settings as there is very little need (I think) to do this on any of the nornir's settings (inventory.options is the only one that is not a scalar)

I hate doing this but due to these reasons I am leaning towards not accepting this PR but I'd like to hear from @dmfigol and @ktbyers first before deciding.

### Alternative

As an alternative to this PR, you could basically do some monkey-patching. I haven't tested it but this should work:

```
def my_custom_build_values(self, init_kwargs):
    ...

from nornir.core.deserializer.configuration import BaseNornirSettings
BaseNornirSettings._build_values = my_custom_build_values

nr = InitNornir(...)
```

Anyway, just wanted to present you with some alternatives to test pending further discussion.

optiz0r:
The use case for this is that to run nornir-based tools, the configuration data can come from multiple locations:
- fairly static data, like URL for inventory system backend
- execution-time data (env vars, cli options)
- runtime data (e.g. device connection secrets, retrieved from a secrets management system like vault), which cannot be written to a config file or env vars.

It used to be possible to use config files for the static data, and pass in `transform_function` at runtime (because it was a top-level setting). From v2, transform_function now moved, and there's no way to independently set this at runtime without masking the config file settings.

Monkey Patching is far from ideal because it's an inherently undocumented use of nornir and subject to breaking change with future nornir changes.

There are at present no hook/extension points within the configuration loading code paths to change the behaviour outside of the nornir core, hence this PR to do it inside the nornir core. Perhaps adding suitable extension points could be a viable alternative here?

dbarrosop:
> and pass in transform_function at runtime (because it was a top-level setting)

Overriding top-level options of each section is supported so this should work as well. If it doesn't let me know because we need to look into that (i.e. inventory.plugins, inventory.transform_function).

> Perhaps adding suitable extension points could be a viable alternative here?

That sounds good to me. Let's wait to see if dmfigol and ktbyers have anything to say first but if we decide to not move forward with this we can certainly support somehow overriding that method so people can tweak building the configuration.

Thanks for your patience with this :)

optiz0r:
>Overriding top-level options of each section is supported so this should work as well. If it doesn't let me know because we need to look into that (i.e. inventory.plugins, inventory.transform_function).

Overriding top-level sections is functional, not saying it isn't. `transform_function` *used* to be top-level, and it therefore it used to be possible to set that in code without affecting inventory options. nornir v2 moved it down one level, now it is *no longer possible* to set `transform_function` in code without masking the config file settings for the inventory plugin (or more specifically, `transform_function_options` which I need to be able to pass in from code because it contains secrets which may not be in the config file, or in env vars).

dbarrosop:
> or more specifically, transform_function_options which I need to be able to pass in from code because it contains secrets which may not be in the config file, or in env vars

This should work. For instance:

```
---
#config.yaml
inventory:
   plugin: path.to.my_plugin
   options:
    option_a: 123
    option_b: asd
  transform_function: path.to.my_function
```
Code:
```
nr = InitNornir(config_file="config.yaml", inventoy={"transform_function_option": {"secret": "asdasd"}})
```

Is that what you mean? Isn't that working?

optiz0r:
>This should work. For instance:

Yes, this is what I mean. This does not work work, and is what I want to make work with this PR. Perhaps we've hit the crux of the issue; you think the functionality I'm trying to enable with these changes is already present?

In your example, the `inventory.plugin`, `inventory.options.option_a`, `inventory.options.option_b` and `inventory.transform_function` config file settings (plus any `NORNIR_INVENTORY_*` env vars which might have been set) are all masked by passing the `inventory` kwarg to `InitNornir`, which only contains the one `transform_function_options` setting. That is what this PR aims to redress, by merging the three sources of `inventory` data (config, env, kwargs), so that the flattened view is what is used to construct the inventory plugin.

Here's why this cannot work with the code in the `develop` branch as it stands:

```python
    @classmethod
    def load_from_file(cls, config_file: str, **kwargs) -> configuration.Config:
        config_dict: Dict[str, Any] = {}
        if config_file:
            yml = ruamel.yaml.YAML(typ="safe")
            with open(config_file, "r") as f:
                config_dict = yml.load(f) or {}
        return Config.deserialize(**{**config_dict, **kwargs})
```

The very last line constructs a new dict, from the config file sections, and the kwargs. It only considers the top-level dict keys (so `core`, `inventory` etc), and if any key is found in `kwargs`, it completely replaces whatever was in `config_dict` like the settings there never existed.

Then, each of the keys in that reconstructed dict are passed in to a separate Pydantic model instance for processing, and the results are copied into the final configuration object.

```python
    @classmethod
    def deserialize(cls, **kwargs) -> configuration.Config:
        c = Config(
            core=CoreConfig(**kwargs.pop("core", {})),
            ssh=SSHConfig(**kwargs.pop("ssh", {})),
            inventory=InventoryConfig(**kwargs.pop("inventory", {})),
            logging=LoggingConfig(**kwargs.pop("logging", {})),
            jinja2=Jinja2Config(**kwargs.pop("jinja2", {})),
            **kwargs,
        )
        return configuration.Config(
            core=CoreConfig.deserialize(**c.core.dict()),
            inventory=InventoryConfig.deserialize(**c.inventory.dict()),
            ssh=SSHConfig.deserialize(**c.ssh.dict()),
            logging=LoggingConfig.deserialize(**c.logging.dict()),
            jinja2=Jinja2Config.deserialize(**c.jinja2.dict()),
            user_defined=c.user_defined,
        )
```

From pydantic's point of view, passing in a kwarg to the model's constructor means that value is taken as authoritative, and so when provided, env vars for that model are not considered either. Even if an env var is set which does not overlap with a kwarg.

The changes the pydantic developer made just for me/us when I first raised that behaviour in December, were to enable passing in the `config_dict` separately from the `kwargs`, and keep them separate right up until `_build_values` is called; at which point a Pydantic model subclass can read both sources of data and chose which ones to use for the final model instantiation. In my code I run a deep merge of the two sources, keeping the non-overlapping settings from both, and using the correct precedence order for the overlapping settings.

dbarrosop:
> you think the functionality I'm trying to enable with these changes is already present?

Yes, and I just tested the example above and it works as I described it. You can't merge what's inside `options` but you can override options (or any other) without affecting the rest. For instance, I added a test (because it wasn't being actively tested), here:

https://github.com/nornir-automation/nornir/pull/337

* `nr.config.core.num_workers` is coming from the config file
* `nr.config.core.raise_on_error` is coming from the code

optiz0r:
The example you've just demonstrated successfully has worked with the code currently in develop for a single day, since #309 was merged in, which contains a subset of the code in this PR. This code was submitted for review prior to #309 existing.  The difference between #309 and this PR is optional full deep merging of the config, which is what we discussed implementing in slack back in December, and is what I produced at the time, albeit without documentation and tests. It's sat in limbo for 6 weeks while I was busy with the day job.

Since the main thing I cared about was first-level section merging, rather than full depth merging, what's in develop as of today will suffice. Feel free to close this; I can live without full merge, and if you don't want the complexity anyway, then this is a happy middle ground.

dbarrosop:
Yeah, sorry, that PR was opened two months ago and fell through the cracks... Ok, let's close this one for now and if we need to revisit this we can look into the hook for overfiding the build_values function.

Thanks again for your patience.

PR #334, opened 2019-02-27
Improve build

dmfigol:
* Completely rework `Dockerfile`. The image size reduced from 1.3 Gb to 450 MB
* Tests which require additional containers are automatically skipped when run locally
* Rework `Makefile` to target specific directories for different tests
* Extract the library version using `pkg_resources`
   It is now available using `import nornir ; nornir.__version__`
* Use `pyproject.toml` for docs generation dependencies, remove `requirements.txt`
* Add the changelog to RTD and slightly update contributing guide

Note: Because I was changing `pyproject.toml` sections (in particular, I added the section `extras`), I had to remove `poetry.lock` to make it work.  
This means that all dev and non-dev dependencies were updated according to the constraints in `pyproject.toml`


dmfigol:
This is ready for review. The build fails because of the problem with `cryptography` 2.6 release today:
https://github.com/pyca/cryptography/issues/4789

dbarrosop:
Ok, this is awesome work. I only have two comments:

1. The Dockerfile feels a bit complex. What is the reasoning to use a multi-stage build? From what I can see you are only saving the need to have poetry in the final image. I also think having all those `COPY` is unnecessary. Ideally, layers should be reusable, and these aren't so caching those layers feels unnecessary (`ADD . /nornir` doesn't take long anyway)
2. Minor comment, not a fan of doing things like `black nornir tests` vs `black .` as the former requires keeping track of new things being added (albeit this should happen rerely)

Again, awesome work.

Dockerfile
```diff
@@ -1,8 +1,53 @@
...
+
+COPY pyproject.toml .
+COPY poetry.lock .
+
+# Dependencies change more often, so we break RUN to cache the previous layer
+RUN ${POETRY_PATH} run pip install --no-cache-dir -U pip setuptools \
+    && ${POETRY_PATH} install --no-interaction
+
+# Source code changes even more often, so we cache another layer
+COPY nornir nornir
```
dbarrosop: to make better use of the cache you should invert the order, that way you'd save copying the README file because nornir folder changed.

dmfigol:
@dbarrosop 
1) It's not only `poetry` but `curl` and `git` which take 100 Mb and are not needed for running our tests.
 
Another thing which addresses both 1) and 2) is that on my setup I have extra folders with some test scripts (which use existing nornir virtual environment). I have them in `.git/info/exclude` so they will never get to the repo by accident, but any time I run `black`, `pylama` or especially `make tests` locally with the current develop branch, the tests don't pass because of these extra Python scripts I have.

dmfigol:
@dbarrosop I rebased and changed order in Dockerfile based on your recommendation.

dbarrosop:
> on my setup I have extra folders

Why do you need them? Or rather, if you think you need them, why not add them to the tests? I don't think we should change things and make them more complex just to accommodate individual environments. It's the same reason I guess why the dockerfile is so massive and complex. Instead of just a single line adding the repo you need to manage each single file and folder individual and I that just makes things more complex and error-prone.

dmfigol:
The most important thing is that my `.venv` is in the `nornir` folder. It will not work, because I am developing on macOS. This can be partially solved with `.dockerignore`
A little bit less important, but still a problem to me: I have folder with some scripts which use editable installation of nornir. They help me verify that new functionality I write, results in the desired outcome, like when I was working on logging, netconf, etc.
I can make COPY in `Dockerfile` easier, but I would ask if we could keep specific things in `Makefile` instead of `.`.
Is this acceptable?

CHANGELOG.rst
```diff
@@ -38,42 +41,43 @@ Thanks to the following people for their contributions:
 * `ktbyers <https://github.com/ktbyers>`_
 * `dbarrosop <https://github.com/dbarrosop>`_
 
+
 2.0.0 - December 17 2018
-========================
+------------------------
 
-For details about upgrading to 2.0.0 see the `notes <https://nornir.readthedocs.io/en/2.0.0-beta/upgrading/1_to_2.html>`_.
+For details about upgrading to 2.0.0 see the `notes <https://nornir.readthedocs.io/en/develop/upgrading/1_to_2.html>`__.
```
dbarrosop: this should all point to stable

Dockerfile
```diff
@@ -1,8 +1,42 @@
...
+FROM python:${PYTHON}-slim-stretch AS builder
+
+WORKDIR /nornir
+ENV PATH="/root/.poetry/bin:$PATH"
+
+RUN apt-get update \
+    && apt-get install -yq curl git \
+    && curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python \
+    && python -m venv .venv \
+    && poetry config settings.virtualenvs.in-project true \
```
dbarrosop: shouldn't this be `settings.virtualenvs.create false`?

docs/conf.py
```diff
@@ -22,7 +22,7 @@
 
 from jinja2 import Environment, FileSystemLoader
 
-import toml
+from nornir import __version__
```
dbarrosop: I am pretty confident this is going to break readthedocs

pyproject.toml
```diff
@@ -15,15 +19,24 @@ python = "^3.6"
...
-netmiko = ">=2.3.3, <3"
-paramiko = ">=2.1.1, <3"
+netmiko = "^2.3.3"
+paramiko = "^2.1.1"
 requests = "^2"
 "ruamel.yaml" = "^0.15.85"
 mypy_extensions = "^0.4.1"
 pydantic = "^0.18.2"
+# The following dependencies are used for docs generation when installed as extras
+# (e.g. pip install nornir[docs])
```
dbarrosop: same here, I am not sure this is going to work in RTD

dbarrosop:
I have a few concerns about RTD so I'd like to merge this into a branch first and enable it in RTD to test before merge.

Could you also rebase? I am afraidthe CHANGELOG and a few others might be outdated by now.

In the future we should break PRs like this into smaller ones as this includes 5 different unrelated changes.

Dockerfile
dmfigol: `settings.virtualenvs.create false` this will use system python
`settings.virtualenvs.in-project true` this will use `.venv` inside project dir. The beauty of the latter is that we can now move the whole project folder around (e.g. between multistage)

dmfigol:
@dbarrosop 
I have updated links in the Changelog. `__version__` trick works locally for me in sphinx, I am not 100% sure `extras` will work in RTD, because I can't access builds there. I was using this as a reference
https://github.com/rtfd/readthedocs.org/issues/4912#issuecomment-457219385

Dockerfile
dbarrosop: Why do you want to create a virtualenv inside a container?

dmfigol: I guess it's my personal preference to keep venvs in docker, do you want me to change it?

dbarrosop: Yeah, I'd like to remove things that are not necessary to keep it simple

dmfigol: I've removed multi-stage build as agreed

dmfigol:
@dbarrosop do you need me to fix something else or we can merge this into a separate branch? If the latter, could you also please enable RTD builds for this branch?
Thanks!

dbarrosop:
Sorry for the delay, I just merged the branch and enabled the build:

https://readthedocs.org/projects/nornir/builds/8865595/

PR #340, opened 2019-03-06
Update the 1-to-2 upgrade guide for custom inventory data

optiz0r:
Highlights the requirement to move custom inventory data under
a `data` subkey when upgrading to nornir 2.x


dbarrosop:
awesome, thanks :)

PR #342, opened 2019-03-06
prepare 2.1 release

dbarrosop:


dmfigol:
In terms of changes of this PR, it looks good to me.
However, I think we shouldn't release 2.1 until the next release of `netmiko` with correct env markers. Currently we use `netmiko` git branch.
https://github.com/nornir-automation/nornir/blob/develop/pyproject.toml#L18
Pinging @ktbyers

ktbyers:
I will see if I can get this done by the end of the weekend (running unit test suite against real devices right now).

ktbyers:
Okay, I should be able to get this done by the end of the weekend. Tests are looking pretty good on it.

Issue #347, opened 2019-03-07
Allow Inventory Data To be Passed From A Dictionary


ipv6freely:
It would be really helpful to be able to pass inventory information from a (correctly structured) host and groups dictionary. How the dictionary was generated should be irrelevant. 

In my particular use case, I'm building the host dictionary simply based on a list of hosts from an NMS or a manual list of IPs. 

In prior versions of Nornir, the following worked correctly:
```isisdb = ['192.168.1.1', '192.168.1.2', '192.168.1.3']

hosts = {
    router: {
        'ip': router,
        'groups': ['wan-routers'],
        'role': 'wan-router'
    } for router in isisdb
}

groups = {
    'wan-routers': {
        'nornir_nos': 'junos'
    }
}

inv = Inventory(hosts=hosts, groups=groups)
```

This doesn't seem to work even more, even when the dictionaries are formatted correctly with correct keys. Restoring this function would be really nice to be able to build an inventory from any user-generated and correctly structured data.

dbarrosop:
I think this is fixed now so I am closing this issue. Feel free to reopen otherwise.

PR #348, opened 2019-03-08
added support to initialize SimpleInventory with native objects

dbarrosop:
Solves #347 

This allows you to use either native objects or yaml files to load the SimpleInventory. For instance:

```
nr = InitNornir(inventory={
    "plugin": "nornir.plugins.inventory.SimpleInventory",
    "options": {
        "host_file": "my_hosts.yaml",
        "group_file": "my_groups.yaml",
        "defaults_file": "my_defaults.yaml",
    }
})
```

or:

```
nr = InitNornir(inventory={
    "plugin": "nornir.plugins.inventory.SimpleInventory",
    "options": {
        "hosts": {
            "host1": {
                "username": "user",
                "groups": ["group_a"],
                "data": {"a": 1, "b": 2},
            },
            "host2": {"username": "user2", "data": {"a": 1, "b": 2}},
        }
        "groups": {"group_a": {"platform": "linux"}},
        "defaults": {"data": {"a_default": "asd"}},
    }
})
```

or a combination:

```
nr = InitNornir(inventory={
    "plugin": "nornir.plugins.inventory.SimpleInventory",
    "options": {
        "hosts_file": "my_hosts.yaml",
        "groups": {"group_a": {"platform": "linux"}},
        "defaults": {"data": {"a_default": "asd"}},
    }
})
```

### DIFF
--- a/nornir/plugins/inventory/simple.py
+++ b/nornir/plugins/inventory/simple.py
@@ -2,7 +2,12 @@ import logging
 import os
 from typing import Any
 
-from nornir.core.deserializer.inventory import GroupsDict, Inventory, VarsDict
+from nornir.core.deserializer.inventory import (
+    HostsDict,
+    GroupsDict,
+    Inventory,
+    VarsDict,
+)
 
 import ruamel.yaml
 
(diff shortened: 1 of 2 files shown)


dbarrosop:
I think we actually need to generalize this pattern so it works with all the plugins as some backends don't have the concept of "defaults" or "groups"

nornir/plugins/inventory/simple.py
```diff
@@ -15,28 +20,34 @@ def __init__(
...
-        if group_file:
-            if os.path.exists(group_file):
-                with open(group_file, "r") as f:
-                    groups = yml.load(f) or {}
-            else:
-                logger.debug("File %r was not found", group_file)
-                groups = {}
+        if groups is None:
+            groups = {}
+            if group_file:
```
ogenstad: Won't this and the defaults_file always be set to groups.yaml or defaults.yaml?

bdlamprecht:
I'd like to see this for filling in values of `connection_options > napalm > extras`.

This would enable setting values for `secret` when connecting to `ios` devices which still require the `enable` command.  (See issue #364).

dmfigol:
@dbarrosop to be honest, I think it is better to have a generic way to do this applicable for any plugin (as you mentioned), than merging this PR which will work only for simple inventory.

dbarrosop:
What would you generalize here? There isn't much to do generalize. If you wanted to implement this same exact thing in netbox it'd be just a matter of:

```
class NetboxInventory(Inventory):
    def __init__(..., groups=None, defaults=None):
        ...
        super.__init__(..., groups=groups or {}, defaults=defaults or {})
```

nornir/plugins/inventory/simple.py
dbarrosop: Not necessarily, someone could pass an empty string or None.

kyoungwookmin:
Does support to initialize SimpleInventory with native objects still work? 
nornir.__version__ is 3.3.0 

I'm trying various things, but I'm getting an error.

**1.  TypeError: SimpleInventory.__init__() got an unexpected keyword argument 'hosts'**

from nornir import InitNornir
nr = InitNornir(inventory={
    "plugin": "SimpleInventory",
    "options": {
        "hosts": {
            "host1": {
                "username": "user",
                "groups": ["group_a"],
                "data": {"a": 1, "b": 2},
            },
        },
        "groups": {"group_a": {"platform": "linux"}},
        "defaults": {"data": {"a_default": "asd"}},
    }
})

**2. PluginNotRegistered: plugin 'nornir.plugins.inventory.SimpleInventory' is not registered**

from nornir import InitNornir
from nornir.plugins.inventory import SimpleInventory
from nornir.core.plugins.inventory import InventoryPluginRegister
InventoryPluginRegister.register("SimpleInventory", SimpleInventory)

nr = InitNornir(inventory={
    "plugin": "nornir.plugins.inventory.SimpleInventory",
    "options": {
        "hosts": {
            "host1": {
                "username": "user",
                "groups": ["group_a"],
                "data": {"a": 1, "b": 2},
            },
        },
        "groups": {"group_a": {"platform": "linux"}},
        "defaults": {"data": {"a_default": "asd"}},
    }
})

davama:
Get the same error:

TypeError: __init__() got an unexpected keyword argument 'hosts'

:(

ktbyers:
@davama Yes, that doesn't exist any longer:

https://github.com/nornir-automation/nornir/blob/main/nornir/plugins/inventory/simple.py#L70-L73

PR #361, opened 2019-03-20
Updated NetBox inventory plugin to support pagination

bdlamprecht:
Using code _borrowed_ from `pynetbox` to accomplish, pagination was implemented.

I attempted to run `make tests` locally before submitting this PR, but there were many files with errors that I didn't modify or touch.

This works correctly on my production version of NetBox using the following `config.yaml`:
```
---
core:
  num_workers: 20

inventory:
  plugin: nornir.plugins.inventory.netbox.NBInventory
  options:
    nb_url: http://netbox.local
    nb_token: CUSTOM_TOKEN
```


bdlamprecht:
After several attempts, all the formatting errors are fixed, now the only thing that remains (which I need help with) is to update the `requests_mock` accordingly.

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
         )
         headers = {"Authorization": "Token {}".format(nb_token)}
 
+        def make_request(url: str) -> Any:
```
dbarrosop: This should return `Dict[str, Any]`

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
...
+            req = requests.get(url=url, headers=headers, params=filter_parameters)
+            req.raise_for_status()
+
+            if req.ok:
```
dbarrosop: Do you need this given you did raise_for_status?

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
...
+                return req.json()
+
+        def req_all(url: str) -> Any:
```
dbarrosop: `Dict[str, Any]`

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
...
+            req = make_request(url)
+            if isinstance(req, dict) and req.get("results") is not None:
```
dbarrosop: Fixing the `make_request` function will save you this check

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
...
+                ret = req["results"]
+                first_run = True
+                while req["next"]:
+                    next_url = (
+                        "{}{}limit={}&offset={}".format(
```
dbarrosop: Use f-strings

nornir/plugins/inventory/netbox.py
```diff
@@ -36,17 +36,41 @@ def __init__(
...
+                            url,
+                            "&" if url[-1] != "/" else "?",
```
dbarrosop: I guess this is because the first run is different from the previous one? If yes, using the same condition would make this more readable

dbarrosop:
To mock tests you can start by adding a function like this that uses `requests_mock` to mock your calls:

https://github.com/nornir-automation/nornir/blob/develop/tests/plugins/inventory/test_netbox.py#L14

(maybe a `get_inv_multiple_calls` function?)

Then you could add a new mocked folder like this with your data:

https://github.com/nornir-automation/nornir/tree/develop/tests/plugins/inventory/netbox/2.3.5

maybe 2.3.5_multiple_calls? (assuming it's the same netbox version)

In your case you will need to patch a few calls to test this but shouldn't be too different from the existing test

nornir/plugins/inventory/netbox.py
bdlamprecht: Done

nornir/plugins/inventory/netbox.py
bdlamprecht: Not sure about that as this also checks for invalid `token` as well.

nornir/plugins/inventory/netbox.py
bdlamprecht: Done.
Since it was originally formatted across multiple lines, I had to use multi-line strings.

nornir/plugins/inventory/netbox.py
bdlamprecht: I'm not sure I understand what you're saying here.
Yes, the first run is different, that is why it is separated under a `while` statement.

bdlamprecht:
The current version of NetBox is `2.5.8` .
A significant amount has changed between now and release `2.3.5`.

I'll **try** to work on getting this updated based off of your suggestions.
No guarantees that it will be correct though. :smile:

nornir/plugins/inventory/netbox.py
bdlamprecht: Pretty sure `raise_for_status()` doesn't catch everything.

dbarrosop: Ok, in that case I think you should raise an exception if `req.ok` returns `False`.

nornir/plugins/inventory/netbox.py
dbarrosop: an error shouldn't be signalled by returning `None` so `make_requests` should either succeed or return an exception. Also, this function doesn't seem to propagate the "error" if `req.get("results")` is `None` so at this point the issue seems to fail silently. If that's the desired behavior it should be correctly modeled by catching the expected error.

tests/plugins/inventory/test_netbox.py
```diff
@@ -12,37 +12,23 @@
...
 
-def transform_function(host):
-    vendor_map = {"Cisco": "ios", "Juniper": "junos"}
-    host["platform"] = vendor_map[host["vendor"]]
+    return netbox.NBInventory.deserialize(**kwargs)
 
 
 class Test(object):
     def test_inventory(self, requests_mock):
-        inv = get_inv(requests_mock, "2.3.5")
```
dbarrosop: don't remove the old test, the old test should still pass, it's how we ensure we don't break things for existing users.

dbarrosop:
Good work, I think we are on the right track, keep working on it addressing the comments and feel free to ping me when you are ready for another review :)

dbarrosop:
To avoid removing the tests don't rewrite current `get_inv` and `test_*` methods. Instead, create new ones; `get_inv_paging` and `test_*_paging`.

If newer versions will include API-breaking changes that will require a new codebase (even if it's hidden behind the same plugin with an selected with an option) and new tests.

bdlamprecht:
Closing due to PR #392 from @wvandeun

Issue #207, opened 2018-07-27
Support for several connections using the same ConnectionPlugin


dmfigol:
For some of my problems I need to be able to connect to the same device using SSH and terminal server at the same time. Both of these connection types are supported by Netmiko, for example.
To be able to support this, there are a couple of things that needs to be implemented:
1) Support from inventory perspective. Potentially something like `serial_options`, but I need to investigate this more
2) Support from Host and Task classes, because currently the task using a specific plugin will use registered name like `napalm`, meaning that only one connection using that ConnectionPlugin can be opened for one host
For this part I am thinking about creating a way to specify a custom name to be stored in `host.connections` dictionary

I am still looking for a good API design, so feedback is welcome.
I will work on this once #195 is merged.

ktbyers:
What if we implemented a second Netmiko connection type `netmiko_terminal_server` or `netmiko_serial`? I say this as then you could have one connection of each type for a given host.

You could also feed in different inventory arguments for the same host:

```
  netmiko_options:
    device_type: cisco_ios_telnet
    port: 23

netmiko_serial_options:
    device_type: terminal_server
    port: 22
```

dbarrosop:
A more generic solution, instead of hardcoding connection types, solution is to do something like this (this is how it would work approx from an user perspective):

```
def my_task(task):
     task.host.open_connection(connection_name="netmiko_ssh", connection="netmiko", **task.host["netmiko_options"])
     task.run(task=netmiko_send_config, connection="netmiko_ssh")
     task.host.close_connection(connection_name="netmiko_ssh")

     task.host.open_connection(connection_name="netmiko_serial", connection="netmiko", **task.host["netmiko_serial"])
     task.run(task=netmiko_send_config, connection="netmiko_serial")
     task.host.close_connection(connection_name="netmiko_serial")
```

And if `connection_name` is not specified default to the canonical name of the connection plugin as we do already today.

Basically, the main reason why I implemented `Connections` as a subclass of dict is so we could implement something like this at some point.

PR #367, opened 2019-03-27
Improve connection handling

### LINKED ISSUE #350: Connection is stored even when not established successfully
Example:
```
def main() -> None:
    with InitNornir("config.yaml") as nr:
        csr_dev = nr.filter(name="csr-dev")
        result = csr_dev.run(task=netmiko_send_command, command_string="sh ip int br")
        print(nr.inventory.hosts['csr-dev'].connections)
        # print_result(result)
    logger.info("END")
```
I deliberately put incorrect credentials in the inventory.
Here is the result:
```
-> % poetry run python sandbox2.py
[2019-03-11 15:45:10,667] INFO     {nornir.core:125} Running task 'netmiko_send_command' with args {'command_string': 'sh ip int br'} on 1 hosts
[2019-03-11 15:45:10,775] INFO     {paramiko.transport:1746} Connected (version 2.0, client Cisco-1.25)
[2019-03-11 15:45:13,728] INFO     {paramiko.transport:1746} Authentication (password) failed.
[2019-03-11 15:45:13,737] ERROR    {nornir.core.task:87} Host 'csr-dev': task 'netmiko_send_command' failed with traceback:
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/src/netmiko/netmiko/base_connection.py", line 858, in establish_connection
    self.remote_conn_pre.connect(**ssh_connect_params)
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/lib/python3.6/site-packages/paramiko/client.py", line 437, in connect
    passphrase,
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/lib/python3.6/site-packages/paramiko/client.py", line 749, in _auth
    raise saved_exception
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/lib/python3.6/site-packages/paramiko/client.py", line 736, in _auth
    self._transport.auth_password(username, password)
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/lib/python3.6/site-packages/paramiko/transport.py", line 1436, in auth_password
    return self.auth_handler.wait_for_response(my_event)
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/lib/python3.6/site-packages/paramiko/auth_handler.py", line 250, in wait_for_response
    raise e
paramiko.ssh_exception.AuthenticationException: Authentication failed.

During handling of the above exception, another exception occurred:

  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/task.py", line 67, in start
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/plugins/tasks/networking/netmiko_send_command.py", line 26, in netmiko_send_command
    net_connect = task.host.get_connection("netmiko", task.nornir.config)
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/inventory.py", line 302, in get_connection
    extras=conn.extras,
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/inventory.py", line 352, in open_connection
    configuration=configuration,
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/plugins/connections/netmiko.py", line 58, in open
    self.connection = ConnectHandler(**parameters)
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/src/netmiko/netmiko/ssh_dispatcher.py", line 228, in ConnectHandler
    return ConnectionClass(*args, **kwargs)
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/src/netmiko/netmiko/base_connection.py", line 312, in __init__
    self.establish_connection()
  File "/Users/dmfigol/projects/my/forks/nornir/.venv/src/netmiko/netmiko/base_connection.py", line 871, in establish_connection
    raise NetMikoAuthenticationException(msg)
netmiko.ssh_exception.NetMikoAuthenticationException: Authentication failure: unable to connect cisco_ios 10.48.18.30:22
Authentication failed.

{'netmiko': <nornir.plugins.connections.netmiko.Netmiko object at 0x11052cb40>}
[2019-03-11 15:45:13,795] INFO     {nornir.core:125} Running task 'close_connections_task' with args {} on 4 hosts
[2019-03-11 15:45:13,799] ERROR    {nornir.core.task:87} Host 'csr-dev': task 'close_connections_task' failed with traceback:
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/task.py", line 67, in start
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/__init__.py", line 152, in close_connections_task
    task.host.close_connections()
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/inventory.py", line 377, in close_connections
    self.close_connection(connection)
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/inventory.py", line 371, in close_connection
    self.connections.pop(connection).close()
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/plugins/connections/netmiko.py", line 61, in close
    self.connection.disconnect()
  File "/Users/dmfigol/projects/my/forks/nornir/nornir/core/connections.py", line 54, in close
    raise ValueError("Connection not established")
ValueError: Connection not established

[2019-03-11 15:45:13,901] INFO     {__main__:59} END
```
You can see that the connection object is stored anyway and then there is another exception when we are trying to close it. We shouldn't have tried closing it in my opinion.

dmfigol:
This is reworked #246

* Add support for several connections using the same plugin #207
* Don't store unestablished connection #350 
* Slight rework of Connection plugin:
  * added class variable "name"
  * slightly change connection plugin registration function signature
  * `ConnectionPlugin.open()` function now returns the connection object instead of modifying the state directly
  * add `task.get_connection` function, which only needs connection name as an argument
  * add `conn_name` and `autoconnect` arguments to `Task.__init__()`:
    * `conn_name: str` allows you to select the connection if you use a connection with a non-default name
    * `autoconnect: bool` is an improvement for already existing manual connection handling. When this argument is set to False, Nornir will not try opening a new connection during task execution and will raise an exception if the connection is not already open.
* completely rewritten howto for connection handling

Some notes from me:
1) Rework of connection plugin registration contains breaking changes. Some of them are aesthetic (e.g. you need only connection plugin class to register it, so it's better if it is the first argument), some of them are both necessity and aesthetic (for example, the most elegant way to fix #350 is to make sure our code crashes on the Connection object creation, instead of initializing object first and then opening connection).
2) Another interesting question/discussion: what is really a breaking change? For example, currently connection plugins are completely undocumented. Then does it count as one? Also, does the same apply for every single function in the core? We don't leverage `_private_function` notation in the codebase
3) Something else that bothers me in my implementation: I've added two new arguments to `Task.__init__()`, while it seems fine from UX standpoint, however this also means that users can't use these variable names `conn_name` & `autoconnect` together with already existing `name` `severity_level` in their tasks. I think we could/should at least prefix them with an underscore? Or may be you have some other suggestion?

Just to clarify, I am not going to push to accept breaking change, but I've decided first to commit what I had in mind, so that we can decide what is the best way: keep them, introduce them in 3.0, rework to support in 2.0 (though probably with uglier implementation).


dbarrosop:
> Then does it count as one

Yes

> We don't leverage _private_function notation in the codebase

Yes, we do. Plenty of functions marked as private in the core.

nornir/core/deserializer/inventory.py
```diff
@@ -53,7 +53,7 @@ def deserialize(
         platform: Optional[str] = None,
         groups: Optional[List[str]] = None,
         data: Optional[Dict[str, Any]] = None,
-        connection_options: Optional[Dict[str, ConnectionOptions]] = None,
+        connection_options: Optional[Dict[str, Any]] = None,
```
dbarrosop: Why?

nornir/core/exceptions.py
```diff
@@ -1,45 +1,66 @@
...
 
 
 class ConnectionException(Exception):
     """
     Superclass for all the Connection* Exceptions
     """
 
-    def __init__(self, connection: "Connection") -> None:
-        self.connection = connection
+    def __init__(self, host: "Host", name: str, details: str = "") -> None:
```
dbarrosop: Breaking change

nornir/core/connections.py
```diff
@@ -21,10 +21,22 @@ class ConnectionPlugin(ABC):
     """
 
     __slots__ = ("connection", "state")
+    name: ClassVar[str] = ""
 
-    def __init__(self) -> None:
-        self.connection: Any = UnestablishedConnection()
+    def __init__(
```
dbarrosop: Breaking change

nornir/core/connections.py
```diff
@@ -50,40 +62,41 @@ def close(self) -> None:
...
-
-    disconnect = close
+    pass
 
 
 class Connections(Dict[str, ConnectionPlugin]):
     available: Dict[str, Type[ConnectionPlugin]] = {}
 
     @classmethod
-    def register(cls, name: str, plugin: Type[ConnectionPlugin]) -> None:
```
dbarrosop: breaking change

nornir/core/exceptions.py
```diff
@@ -1,45 +1,66 @@
...
+        self.host = host
 
 
 class ConnectionAlreadyOpen(ConnectionException):
     """
     Raised when opening an already opened connection
     """
 
-    pass
+    def __init__(
```
dbarrosop: Breaking change

nornir/core/exceptions.py
```diff
@@ -1,45 +1,66 @@
...
+        super().__init__(host, name, details)
 
 
 class ConnectionNotOpen(ConnectionException):
     """
-    Raised when trying to close a connection that isn't open
+    Raised when the connection is supposed to be open but it isn't
     """
 
+    def __init__(self, host: "Host", name: str, details: str = "is not open") -> None:
```
dbarrosop: Breaking change

nornir/core/inventory.py
```diff
@@ -212,9 +212,9 @@ def get(self, item, default=None):
             return default
 
     def get_connection_parameters(
-        self, connection: Optional[str] = None
```
dbarrosop: Breaking change

nornir/core/inventory.py
```diff
@@ -272,50 +272,60 @@ def _get_connection_options_recursively(
             p.extras = p.extras if p.extras is not None else sp.extras
         return p
 
-    def get_connection(self, connection: str, configuration: Config) -> Any:
+    def get_connection(
```
dbarrosop: breaking change

nornir/core/inventory.py
```diff
@@ -272,50 +272,60 @@ def _get_connection_options_recursively(
...
+                    platform=conn_params.platform,
+                    extras=conn_params.extras,
+                )
+            else:
+                details = "is not open and automatic connection is disabled"
+                raise ConnectionNotOpen(self, name, details)
+        return self.connections[name].connection
 
-    def get_connection_state(self, connection: str) -> Dict[str, Any]:
+    def get_connection_state(self, name: str) -> Dict[str, Any]:
```
dbarrosop: breaking change

nornir/core/inventory.py
```diff
@@ -272,50 +272,60 @@ def _get_connection_options_recursively(
...
         """
-        if connection not in self.connections:
-            raise ConnectionNotOpen(connection)
-
-        return self.connections[connection].state
+        if name not in self.connections:
+            raise ConnectionNotOpen(self, name)
+        return self.connections[name].state
 
     def open_connection(
```
dbarrosop: breaking change

nornir/core/task.py
```diff
@@ -36,9 +38,19 @@ class Task(object):
         severity_level (logging.LEVEL): Severity level associated to the task
     """
 
-    def __init__(self, task, name=None, severity_level=logging.INFO, **kwargs):
+    def __init__(
```
dbarrosop: breaking change

nornir/core/task.py
```diff
@@ -131,6 +143,33 @@ def is_dry_run(self, override: bool = None) -> bool:
...
+        """
+        conn_name = self.conn_name or conn_plugin.name
+        return self.host.get_connection(
+            name=conn_name,
+            configuration=self.nornir.config,
+            plugin_name=conn_plugin.name,
+            autoconnect=self.autoconnect,
+        )
+
+    def get_connection_state(self, conn_plugin: Type["ConnectionPlugin"]):
```
dbarrosop: what is connection state?

nornir/core/task.py
```diff
@@ -131,6 +143,33 @@ def is_dry_run(self, override: bool = None) -> bool:
         """
         return override if override is not None else self.nornir.data.dry_run
 
+    def get_connection(self, conn_plugin: Type["ConnectionPlugin"]):
```
dbarrosop: I don't think this belongs here. The task is not connected to a device, a `Host` is connected to a device.

nornir/core/connections.py
```diff
@@ -36,12 +48,12 @@ def open(
         platform: Optional[str],
         extras: Optional[Dict[str, Any]] = None,
         configuration: Optional[Config] = None,
-    ) -> None:
+    ) -> Any:
```
dbarrosop: If we are going to return something let's alias it to something like `DeviceConnection` or something like that.

docs/howto/handling_connections.ipynb
```diff
@@ -4,16 +4,176 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# How to handle connections to devices\n",
+    "# How to handle connections to devices"
```
dbarrosop: This should have a dedicated howto.

tests/core/test_connections.py
```diff
@@ -83,31 +137,56 @@ class Test(object):
...
         nr = nornir.filter(name="dev2.group_1")
         r = nr.run(task=open_connection_twice, num_workers=1)
         assert len(r) == 1
         assert not r.failed
 
+    def test_auto_two_connections_dummy_plugin(self, nornir):
+        with nornir.filter(name="dev2.group_1") as nr:
+            r1 = nr.run(task=dummy_task)
+            assert r1["dev2.group_1"].result == "ok"
+            r2 = nr.run(task=dummy_task, conn_name="secondary_dummy")
```
dbarrosop: We need to find a better way. An alternative would be to have a task_configuration parameter and a TaskConfiguration object where we allow tweaking the behavior.

docs/howto/handling_connections.ipynb
```diff
@@ -4,16 +4,176 @@
...
+       "          extras:\n",
+       "              # non-standard library-specific options (check backend library docs)\n",
+       "              optional_args:\n",
+       "                  eos_autoComplete: true\n",
+       "        netmiko:\n",
+       "            platform: arista_eos\n",
+       "            extras:\n",
+       "                secret: \"my-enable-password\"\n",
+       "                global_delay_factor: 2\n",
+       "        netmiko_telnet:  # non-default connection name\n",
```
dbarrosop: how do you know this is a netmiko connection?

dbarrosop:
I really like the functionality but I think the changes are too big for a minor release. The way I see this all we need is:

1. To have a mechanism to express different connections of the same type (which you mostly have already)
2. A way to store multiple connections (which we have already)
3. A way to signal a task we want an alternate connection name

docs/howto/handling_connections.ipynb
dmfigol: Why? All of this refers to connection handling.

dmfigol:
@dbarrosop do you propose adapting changes to 2.x or adapt them for 3.x?

docs/howto/handling_connections.ipynb
dmfigol: I don't know beforehand, but I deduce it using the task. E.g. `netmiko_send_command` uses plugin netmiko so if we use a connection named `netmiko_telnet` for `netmiko_send_command` task, the connection should be created using netmiko plugin.
Obviously, if you open connection manually, you need to specify it yourself.

docs/howto/handling_connections.ipynb
dbarrosop: Because it's a different and more advanced use case. How tos are supposed to be short, concise and target a specific question and that how to already covers too many things. So this should be something like "Handling multiple connections with the same backend to the same device"

dbarrosop:
I don't know when 3.x is going to happen, not anytime soon as people is still adapting to 2.x. Unless something major happens I'd say we are at least 1 year out to consider 3.x so I'd suggest reducing the scope of the current change. When we decide to work on 3.x because we have enough data/experience to consider reevaluating certain parts of the API we can then revisit this. The benefit of this approach is that we can get experience now and do a complete forklift without having to consider current implementation (as opposed to your current changes which are dancing around some things we may want to change in the future).

dbarrosop:
Are there plans to keep working on this?

dmfigol:
Yes, haven't had time recently. However I still don't know about the best interface to add different arguments like connection name (and to be fair, all other non-task arguments should go there too). May be it is a Task runner which I haven't checked yet.

dbarrosop:
I think there are two main changes that needs to be done:

In the inventory, the key of connection_options is basically the connection name. Then, we add `plugin` key underneath that will be "napalm", "netmiko", "paramiko", etc... if there is not `plugin` key we assume plugin is the same as the name. Now, `task.get_connection("napalm")` should probably require no change as you are going to be retrieving the connection by name instead of by plugin (which should basically pose no change to the API as today name == plugin).

This is going to lead to a few internal changes but I think this is the minimum we should do for nonir 2.0 to make this work. We can explore other internal changes like the one you proposed in the beta idea that I am toying with.

FWIW I think the `autoconnect` should be a global option for nornir. Specially as I think this is going to be a matter of personal preference (like raise_on_error).

dbarrosop:

Issue #346, opened 2019-03-07
Path to files in `config.yaml` depends on the working directory


dmfigol:
Personally, I find it confusing that when you specify paths to `host_file`, `group_file` in `config.yaml` they are treated as relative path from the script execution directory, instead of relative to `config.yaml` itself.
What do you think?

dbarrosop:
Paths are always relative to the working dir. if you find it confusing use absolute paths :D Something we can do though is expand env variables so you can do something like:

```
hosts_file: $MY_CUSTOM_ENV/inventory/hosts.yaml
```

dmfigol:
When it is Python code, it is easy to hack something like `Path(__file__).parent / "inventory/hosts.yaml"`, but with a config file `config.yaml` it is harder to do: either you use absolute paths (which make it problematic to port between dev / prod), or you always need to be extra careful where you call the script.
May be I am just imagining a problem which does not exist.
That said I like the idea of automatic env vars expansion and it is really easy to implement.

dmfigol:
I will close this, not really relevant anymore. though env var expand is still an interesting feature.

PR #368, opened 2019-03-29
Issue #346 env var expand

eakman:
Implementation of suggestion by @dbarrosop in issue #346. These changes add environment variable expansion to Config#load_from_file method so that environment variables can be used in configuration files or as `kwargs` to `InitNornir`.

For example a configuration file with the following format could be used:

```
# config_file.yaml
---
core:
    num_workers: 100
inventory:
    plugin: nornir.plugins.inventory.simple.SimpleInventory
    options:
        host_file: $HOSTS
        group_file: $GROUPS
```
Or `InitNornir` could be invoked with the following `kwargs`:

```
 nr = InitNornir(
    core={"num_workers": 100},
    inventory={
        "plugin": "nornir.plugins.inventory.simple.SimpleInventory",
        "options": {"host_file": "$HOSTS", "group_file": "$GROUPS"},
    },
    user_defined={"data": ["$RANDOM_VAR"]}
)
```

The PR consists of the following changes:
- _deep_expand helper function added to `nornir/core/deserializer/configuration.py`
- Slight modification to `Config#load_from_file` method in `nornir/core/deserializer/configuration.py`
- A test config file added to `tests/core/deserializer/test_configuration/`
- Two unit tests added to `tests/core/deserializer/test_configuration.py`
- A test config file added to `tests/core/test_InitNornir`
- Two unit tests added to `tests/core/test_InitNornir.py`

### DIFF
--- a/nornir/core/deserializer/configuration.py
+++ b/nornir/core/deserializer/configuration.py
@@ -165,32 +165,32 @@ class Config(BaseNornirSettings):
     ) -> configuration.Config:
         __config_settings__ = __config_settings__ or {}
 
-        expanded_config = _deep_expand(__config_settings__)
-        expanded_kwargs = _deep_expand(kwargs)
+        # expanded_config = _deep_expand(__config_settings__)
+        # expanded_kwargs = _deep_expand(kwargs)
 
         c = Config(
             core=CoreConfig(
-                __config_settings__=expanded_config.pop("core", {}),
-                **expanded_kwargs.pop("core", {}),
+                __config_settings__=__config_settings__.pop("core", {}),
+                **kwargs.pop("core", {}),
             ),
             ssh=SSHConfig(
-                __config_settings__=expanded_config.pop("ssh", {}),
-                **expanded_kwargs.pop("ssh", {}),
+                __config_settings__=__config_settings__.pop("ssh", {}),
+                **kwargs.pop("ssh", {}),
             ),
             inventory=InventoryConfig(
-                __config_settings__=expanded_config.pop("inventory", {}),
-                **expanded_kwargs.pop("inventory", {}),
+                __config_settings__=__config_settings__.pop("inventory", {}),
+                **kwargs.pop("inventory", {}),
             ),
             logging=LoggingConfig(
-                __config_settings__=expanded_config.pop("logging", {}),
-                **expanded_kwargs.pop("logging", {}),
+                __config_settings__=__config_settings__.pop("logging", {}),
+                **kwargs.pop("logging", {}),
             ),
             jinja2=Jinja2Config(
-                __config_settings__=expanded_config.pop("jinja2", {}),
-                **expanded_kwargs.pop("jinja2", {}),
+                __config_settings__=__config_settings__.pop("jinja2", {}),
+                **kwargs.pop("jinja2", {}),
             ),
-            __config_settings__=expanded_config,
-            **expanded_kwargs,
+            __config_settings__=__config_settings__,
+            **kwargs,
         )
         return configuration.Config(
             core=CoreConfig.deserialize(**c.core.dict()),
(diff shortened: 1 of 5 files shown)


dbarrosop:
Awesome work. From an implementation point of view, I think it'd be cleaner to use a custom type instead of the function `_deep_expand`; https://pydantic-docs.helpmanual.io/#custom-data-types

The `validate` function would both validate it's a string and do the expansion. Thoughts?

eakman:
Interesting idea! Truth be told, I'm a little new to Pydantic, but let me do a bit of a reading and see if I can fix this up.

eakman:
So how do you envision this custom type being used? Do you mean rather than traversing the data from the config file, use a custom type for select properties of the various config types and have the traversal/expansion logic live in the custom type's validators?

i.e.

```
class InventoryConfig(BaseNornirSettings):
    plugin: str = Schema(
        default="nornir.plugins.inventory.simple.SimpleInventory",
        description="Import path to inventory plugin",
    )
    options: Dict[str, CustomTypeWhichExpandsWhenNeeded] = Schema(
        default={}, description="kwargs to pass to the inventory plugin"
    )
    ...
```

Sorry if I'm way off here lol.

Best,
Eitan

eakman:
Or alternatively, are you looking for something like this.

```
class RawConfigData(object):
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, conf):
        if isinstance(conf, str):
            return os.path.expandvars(conf)
        if isinstance(conf, dict):
            for key, value in conf.items():
                conf[key] = RawConfigData.validate(value)
        elif isinstance(conf, list):
            return list(map(lambda x: RawConfigData.validate(x), conf))
        return conf

class ConfigData(BaseModel):
    data: RawConfigData

os.environ["MYVAR"] = "this is my var"
os.environ["MYVAR2"] = "this is another var"

mock = {"hello": [{"hello": "$MYVAR"}, "$MYVAR2"]}
print (ConfigData(data=mock).data) # --> {'hello': [{'hello': 'this is my var'}, 'this is another var']}
```

dbarrosop:
Similar to the second option but only for strings:

```
class StringExpVars(object):
    @classmethod
    def __get_validators__(cls):
        yield cls.validate

    @classmethod
    def validate(cls, conf):
        if isinstance(conf, str):
            return os.path.expandvars(conf)
        else:
            raise ValueError("only strings supported here")
```

And then just replace `str` with this in the places where we want vars to be expanded. Which may be all of the places tbh.

dbarrosop:
An alternative might be to treat the configuration file as a jinja2 template, method that could be extended to the inventory files (sorry this is derailing).

eakman:
Haha! It's all good!

My concern is that if we expand strings on a case by case basis, users might find the variable expansion behavior difficult to predict. They may expect it to work in places where it does not.

The jinja2 suggestion gave me an idea. We could treat the entire config file content as a string and then just expand that.

i.e.

```
...
    @classmethod
    def load_from_file(cls, config_file: str, **kwargs: Any) -> configuration.Config:
        config_dict: Dict[str, Any] = {}
        if config_file:
            yml = ruamel.yaml.YAML(typ="safe")
            with open(config_file, "r") as f:
                config_dict = yml.load(os.path.expandvars(f.read())) or {}
        return Config.deserialize(
            __config_settings__=config_dict, **_deep_expand(kwargs)
        )
...
```

We could also add a flag to the load_from_file to disable/enable expansion. Your call! Alternatively, we can just table this for now and let it brew a for a bit until we think of the best solution.

Best,
Eitan

dbarrosop:
No need to table this, your suggestion is cleaner and simpler so let's go with it. I'd like to support this in the `SimpleInventory` too. I wouldn't bother using a flag to toggle this for now. We can look at it later on if needed.

eakman:
Ok sounds good! I'll get to work on this and re-submit.

dbarrosop:
Hi, just following up on this, are you planning to keep working on this?

dbarrosop:

Issue #370, opened 2019-04-02
Dynamically adding hosts to inventory does not populate defaults to the hosts


brandomando:
As I discover hosts using a crawler, `nornir.core.inventory.Host` objects are created, and the inventory is updated with the new host.  When attempting to connect to the host using the `username` and `password` attributes of `nornir.inventory.defaults`, it raises an exception that "No authentication methods available", indicating that the defaults are not being mapped to the dynamically created hosts.

As a workaround, I mapped the inventory defaults to the host defaults as I created the object:
```
host = Host(
    name=name,
    platform=platform,
    hostname=hostname,
    defaults=nr.inventory.defaults
    )
```
Once the defaults were mapped to the inventory defaults, the auth error subsided and I was able to successfullly connect to the new host.

PR #372, opened 2019-04-03
Added add_host and add_group functions to nornir.core.inventory.Inventory class

brandomando:
Added add_host and add_group functions to nornir.core.inventory.Inventory class.  Automatically maps the existing inventory defaults to the new host or group as they are dynamically created after initialization. Added unit tests to test_inventory module.

Ref: [nornir-automation/nornir Issue# 370](https://github.com/nornir-automation/nornir/issues/370)

### DIFF
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -438,3 +438,19 @@ class Inventory(object):
             if host.has_parent_group(group):
                 hosts.add(host)
         return hosts
+
+    def add_host(self, name: str, defaults: Optional[Defaults] = None,
+                **kwargs) -> None:
+        """
+        Add a host to the inventory after initialization
+        """
+        host = {name: Host(name, defaults=self.defaults, **kwargs)}
+        self.hosts.update(host)
+
+    def add_group(self, name: str, defaults: Optional[Defaults] = None,
+                **kwargs) -> None:
+        """
+        Add a group to the inventory after initialization
+        """
+        group = {name: Group(name, defaults=self.defaults, **kwargs)}
+        self.groups.update(group)
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -439,18 +439,24 @@ class Inventory(object):
                 hosts.add(host)
         return hosts
 
-    def add_host(self, name: str, defaults: Optional[Defaults] = None,
-                **kwargs) -> None:
+    def add_host(self, name: str, **kwargs) -> None:
         """
         Add a host to the inventory after initialization
         """
-        host = {name: Host(name, defaults=self.defaults, **kwargs)}
+        host = {
+            name: deserializer.inventory.InventoryElement.deserialize_host(
+                name=name, defaults=self.defaults, **kwargs
+            )
+        }
         self.hosts.update(host)
 
-    def add_group(self, name: str, defaults: Optional[Defaults] = None,
-                **kwargs) -> None:
+    def add_group(self, name: str, **kwargs) -> None:
         """
         Add a group to the inventory after initialization
         """
-        group = {name: Group(name, defaults=self.defaults, **kwargs)}
+        group = {
+            name: deserializer.inventory.InventoryElement.deserialize_group(
+                name=name, defaults=self.defaults, **kwargs
+            )
+        }
         self.groups.update(group)
--- a/tests/core/test_inventory.py
+++ b/tests/core/test_inventory.py
@@ -236,3 +236,43 @@ class Test(object):
             inv.hosts["dev4.group_2"],
             inv.hosts["dev3.group_2"],
         }
+
+    def test_add_host(self):
+        data = {"test_var": "test_value"}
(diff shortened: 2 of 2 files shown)


dbarrosop:
You need to run black to format the code.

nornir/core/inventory.py
```diff
@@ -438,3 +438,19 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
             if host.has_parent_group(group):
                 hosts.add(host)
         return hosts
+
+    def add_host(self, name: str, defaults: Optional[Defaults] = None,
```
dbarrosop: You are accepting `defaults` as an argument but you are not really using it :) In any case, it's not really necessary so removing it should be ok.

nornir/core/inventory.py
```diff
@@ -438,3 +438,19 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
...
+                **kwargs) -> None:
+        """
+        Add a host to the inventory after initialization
+        """
+        host = {name: Host(name, defaults=self.defaults, **kwargs)}
```
dbarrosop: This is a really good start, however, in order to properly deserialize connection options you should use this method:

https://github.com/nornir-automation/nornir/blob/develop/nornir/core/deserializer/inventory.py#L78

So your code will be something like `host = {name: InventoryElement.deserialize_host(name=name, defaults=self.defaults, **kwargs)}`

nornir/core/inventory.py
```diff
@@ -438,3 +438,19 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
...
+        self.hosts.update(host)
+
+    def add_group(self, name: str, defaults: Optional[Defaults] = None,
+                **kwargs) -> None:
+        """
+        Add a group to the inventory after initialization
+        """
+        group = {name: Group(name, defaults=self.defaults, **kwargs)}
```
dbarrosop: Same comments as with `add_host` but using `deserialize_group` instead.

tests/core/test_inventory.py
```diff
@@ -236,3 +236,43 @@ def test_children_of_obj(self):
             inv.hosts["dev4.group_2"],
             inv.hosts["dev3.group_2"],
         }
+
+    def test_add_host(self):
```
dbarrosop: Nice! Would you mind adding also some connection_options  to h3 as that's the tricky part here?

dbarrosop:
Nice work, just a couple of minor comments

tests/core/test_inventory.py
brandomando: Thank you. I added the assertions to both add_host and add_group tests.

nornir/core/inventory.py
```diff
@@ -439,18 +439,24 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
...
         """
-        host = {name: Host(name, defaults=self.defaults, **kwargs)}
+        host = {
+            name: deserializer.inventory.InventoryElement.deserialize_host(
+                name=name, defaults=self.defaults, **kwargs
+            )
+        }
         self.hosts.update(host)
 
-    def add_group(self, name: str, defaults: Optional[Defaults] = None,
```
brandomando: Removed unused defaults from `add_group` function definition

nornir/core/inventory.py
```diff
@@ -439,18 +439,24 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
                 hosts.add(host)
         return hosts
 
-    def add_host(self, name: str, defaults: Optional[Defaults] = None,
```
brandomando: Removed unused defaults variable from `add_host` definition

nornir/core/inventory.py
```diff
@@ -439,18 +439,24 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
...
-                **kwargs) -> None:
+    def add_host(self, name: str, **kwargs) -> None:
         """
         Add a host to the inventory after initialization
         """
-        host = {name: Host(name, defaults=self.defaults, **kwargs)}
+        host = {
```
brandomando: Re-wrote `host` variable using the `deserialize_host` function rather than the direct `Host` class initialization

nornir/core/inventory.py
```diff
@@ -439,18 +439,24 @@ def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
...
         self.hosts.update(host)
 
-    def add_group(self, name: str, defaults: Optional[Defaults] = None,
-                **kwargs) -> None:
+    def add_group(self, name: str, **kwargs) -> None:
         """
         Add a group to the inventory after initialization
         """
-        group = {name: Group(name, defaults=self.defaults, **kwargs)}
+        group = {
```
brandomando: Re-wrote `host` variable using the `deserialize_group` function rather than the direct `Group` class initialization

dbarrosop:
Nice work! :)

PR #373, opened 2019-04-08
Improve build

dbarrosop:


dbarrosop:
@dmfigol opened this PR to keep with the discussion. I will try to do a last review now that we know that readthedocs is working and merge

pyproject.toml
```diff
@@ -15,15 +19,25 @@ python = "^3.6"
...
 napalm = "^2"
-netmiko = ">=2.3.3, <3"
-paramiko = ">=2.1.1, <3"
+netmiko = "^2.3.3"
+paramiko = "^2.1.1"
 requests = "^2"
 "ruamel.yaml" = "^0.15.85"
 mypy_extensions = "^0.4.1"
 pydantic = "^0.18.2"
+# The following dependencies are used for docs generation when installed as extras
```
dbarrosop: I think I'd rather add this to the dev dependencies.

Makefile
```diff
@@ -24,50 +25,44 @@ build_test_container:
 .PHONY: enter-container
 enter-container:
 	${DOCKER_COMPOSE} \
-		run nornir bash
+		run --rm nornir bash
```
dbarrosop: maybe we should add `run --rm` to the the DOCKER_COMPOSE variable

dbarrosop:
Ok, just left two minor comments, once they are addressed I will merge it.

Good job, btw, glad to see we got it done finally :)

PR #374, opened 2019-04-08
added external resources

dbarrosop:
@dmfigol @ogenstad @ktbyers any opinions/missing resources about this? Format could also be better, if you have suggestions let me know :)


ogenstad:
Let me update my blog post so that the example is in Nornir 2.x format to avoid confusion. Should perhaps also reach out to Cisco and see if I can change the one there.

PR #375, opened 2019-04-10
Get inventory dicts

brandomando:
Adding functions to nornir.core.inventory.Inventory to abstract serialization of the running inventory into dictionaries. Functions for: full inventory, defaults, groups, and hosts. This makes the inventory available in an exportable format that users could then build plugins to push out into various formats.

A simple future usecase would be to leverage the dicts to export to simple inventory files.  This would allow dynamically added inventory items to be captured and placed into a more permanent inventory.

### DIFF
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -471,4 +471,4 @@ class Inventory(object):
         return self.get_inventory_dict().get("groups")
 
     def get_hosts_dict(self):
-        raise NotImplementedError
+        return self.get_inventory_dict().get("hosts")
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -461,14 +461,26 @@ class Inventory(object):
         }
         self.groups.update(group)
 
-    def get_inventory_dict(self):
+    def get_inventory_dict(self) -> Dict:
+        """
+        Return serialized dictionary of inventory
+        """
         return deserializer.inventory.Inventory.serialize(self).dict()
 
-    def get_defaults_dict(self):
+    def get_defaults_dict(self) -> Dict:
+        """
+        Returns serialized dictionary of defaults from inventory
+        """
         return self.get_inventory_dict().get("defaults")
 
-    def get_groups_dict(self):
+    def get_groups_dict(self) -> Dict:
+        """
+        Returns serialized dictionary of groups from inventory
+        """
         return self.get_inventory_dict().get("groups")
 
-    def get_hosts_dict(self):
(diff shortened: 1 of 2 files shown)


nornir/core/inventory.py
```diff
@@ -460,3 +460,27 @@ def add_group(self, name: str, **kwargs) -> None:
...
+        """
+        Return serialized dictionary of inventory
+        """
+        return deserializer.inventory.Inventory.serialize(self).dict()
+
+    def get_defaults_dict(self) -> Dict:
+        """
+        Returns serialized dictionary of defaults from inventory
+        """
+        return self.get_inventory_dict().get("defaults")
```
dbarrosop: I think we should call instead:

```
         return deserializer.inventory.Defaults.serialize(self.defaults).dict()
```

nornir/core/inventory.py
```diff
@@ -460,3 +460,27 @@ def add_group(self, name: str, **kwargs) -> None:
...
+
+    def get_groups_dict(self) -> Dict:
+        """
+        Returns serialized dictionary of groups from inventory
+        """
+        return self.get_inventory_dict().get("groups")
```
dbarrosop: I think we should call instead:

```
         return {k: deserializer.inventory.InventoryElement.serialize(v).dict() for k, v in self.groups}
```

nornir/core/inventory.py
```diff
@@ -460,3 +460,27 @@ def add_group(self, name: str, **kwargs) -> None:
...
+
+    def get_hosts_dict(self) -> Dict:
+        """
+        Returns serialized dictionary of hosts from inventory
+        """
+        return self.get_inventory_dict().get("hosts")
```
dbarrosop: I think we should call instead:

```
         return {k: deserializer.inventory.InventoryElement.serialize(v).dict() for k, v in self.hosts}
```

dbarrosop:
Looks good, just a few minor comments to make the code more efficient.

nornir/core/inventory.py
brandomando: I see what you're saying on this and the groups. It avoids calling an additional function to perform the action. One minor correction however, just adding `.items()` on both this and groups
```
        return {k: deserializer.inventory.InventoryElement.serialize(v).dict() for k, v in self.hosts.items()}
```

brandomando:
@dbarrosop , I believe I have addressed your changes. Let me know if there's anything else.

Thanks

dbarrosop:
Awesome!

Issue #383, opened 2019-04-27
add_host, add_groups, and add_defaults methods should re-initialize the Inventory object


brandomando:
While working on some documentation for the `add_hosts` in `inventory.py`, I realized that it is possible now to add a host to the inventory with groups that are not defined in the `Nornir.inventory.groups` object because the only time the check is run is when `Inventory` is initialized.

To resolve this, I would propose adding `self.__init__()` to the end of the method to re-initialize the inventory after each host or group is added to the inventory using the `add_host` or `add_group` methods to check for groups' existence (also since groups can be nested).

brandomando:
Fixed by https://github.com/nornir-automation/nornir/issues/384

PR #384, opened 2019-04-27
Fix add_host and add_group methods to re-initialize inventory

brandomando:
Referencing Issue #383 

Added line on `add_host` and `add_group` methods to re-initialize inventory.  Fixes the issue of being able to add undefined parent groups to hosts and groups when they are added dynamically using those methods of the `Inventory` class.

### DIFF
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -428,6 +428,17 @@ class Inventory(object):
     def __len__(self):
         return self.hosts.__len__()
 
+    def _update_group_refs(self, inventory_element: InventoryElement) -> None:
+        """
+        Returns inventory_element with updated group references for the supplied
+        inventory element
+        """
+        if hasattr(inventory_element, "groups"):
+            inventory_element.groups.refs = [
+                self.groups[p] for p in inventory_element.groups
+            ]
+        return inventory_element
+
     def children_of_group(self, group: Union[str, Group]) -> Set[Host]:
         """
         Returns set of hosts that belongs to a group including those that belong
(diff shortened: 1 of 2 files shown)


brandomando:
@dbarrosop this fixes the recursion when adding hosts or groups to the inventory that we had originally discussed. Unless the `Inventory` object is re-initialized, the lower objects do not properly inherit the parents' values.

nornir/core/inventory.py
```diff
@@ -449,6 +449,7 @@ def add_host(self, name: str, **kwargs) -> None:
             )
         }
         self.hosts.update(host)
+        self.__init__(hosts=self.hosts, groups=self.groups, defaults=self.defaults)
```
dbarrosop: I am concerned about side effects if the __init__ function changes or if the inventory is too large and this becomes  a waste of CPU cycles. Maybe we can just create a `_add_group_refs` function and explicitly execute these two lines:

```
def _update_group_refs(self, inventory_element: InventoryElement) -> None:
        inventory_element.groups.refs = [self.groups[p] for g in inventory_element.groups]
```

Or something similar.

brandomando:
I agree. As I thought through it some more, if you have a large inventory
or if you are adding lots of hosts, it could become a very expensive
transaction. I will add those helpers for updating the refs instead.

Thanks for the feedback.

nornir/core/inventory.py
brandomando: I added the helper and associated tests. It seems to work well like this.

nornir/core/inventory.py
```diff
@@ -428,6 +428,17 @@ def filter(self, filter_obj=None, filter_func=None, *args, **kwargs):
     def __len__(self):
         return self.hosts.__len__()
 
+    def _update_group_refs(self, inventory_element: InventoryElement) -> None:
+        """
+        Returns inventory_element with updated group references for the supplied
+        inventory element
+        """
+        if hasattr(inventory_element, "groups"):
```
brandomando: I added the `if hasattr` because it was throwing an `AttributeError` if there were no pre-defined groups in the inventory. This addresses that without the need for implementing error handling.

dbarrosop:
Awesome! Merging :)

PR #385, opened 2019-04-29
Add Apache Kafka streaming function

bewing:


nornir/plugins/functions/stream/kafka.py
```diff
@@ -0,0 +1,78 @@
...
+        result: (:obj:`nornir.core.task.Result`)
+        topic (``str``): Topic to send on
+        bootstrap_servers (``str`` or ``[str, ...]``: Kafka brokers to bootstrap from
+        key: Key for the message, default the name of the :obj:`nornir.core.inventory.Host`
+        key_serializer: (``callable``): Function to serialize key to bytes.  Default is UTF-8
+        value_serializer (``callable``): Function to serialize result value to bytes.  Default is
+           a JSON dump of the result object, with objects de-referenced.
+        **kwargs: keyword args to pass into :obj:`kafka.producer.kafka.KafkaProducer`
+    """
+    with LOCK:
```
bewing: Not 100% sure this lock is necessary since we're not dealing with screen I/O, and we create a `Producer` object for each call.

nornir/core/task.py
```diff
@@ -174,6 +175,7 @@ def __init__(
         self.exception = exception
         self.name = None
         self.severity_level = severity_level
+        self.timestamp = datetime.datetime.now(datetime.timezone.utc)
```
dbarrosop: I am curious about this, you added it but don't seem to be using it anywhere

bewing: The primary use-case for me is "poor man's telemetry", meaning running getters on a schedule, timestamping them, and using a pipeline to eventually process them into other datastores (ELK, etc).  Having the timestamp of when the result was generated helps to order the results, in the case where a large number of them result in batching inside the producer.

If you think that's unnecessary, I'd be happy to insert the timestamp inside _`send_single_result`

dbarrosop: It's all good, just realized you added this to the `Result` class, I misread it yesterday and thought it was being added to the `Task` class.

nornir/plugins/functions/stream/kafka.py
dbarrosop: If we are unsure I'd suggest removing it and adding it later as the lock is going to create some contention

dbarrosop:
Not sure if you are familiar with jupyter but I'd like to have something like this:

https://github.com/nornir-automation/nornir/blob/develop/docs/plugins/tasks/version_control/gitlab.ipynb

which serves as both docs and tests:

https://nornir.readthedocs.io/en/stable/plugins/tasks/version_control/gitlab.html

dbarrosop:

Issue #386, opened 2019-04-30
Create napalm_rollback task.


jchristopher327:
Hey,

I wanted to utilize Napalm's rollback feature and didn't see it as part of Nornir. I have the code for this but wanted to see if it's worth having a function in Napalm that compares the rollback config to the running config. I think this would involve creating a new function in Napalm and then updating Nornir to use that function. The benefit of this is to have the user see what the rollback changes will be before executing the rollback.

If this is outside the scope of the request (very likely), let me know and I'll create the PR for `napalm_rollback`.

Thanks.

ktbyers:
That sounds reasonable to me. So have it as far as creating a PR.

ogenstad:
Closing as stale, please open a new request if this is still needed.

PR #387, opened 2019-04-30
add napalm_rollback

jchristopher327:
Implements #386

### DIFF
--- a/nornir/plugins/tasks/networking/napalm_rollback.py
+++ b/nornir/plugins/tasks/networking/napalm_rollback.py
@@ -1,16 +1,16 @@
 from nornir.core.task import Result, Task
 
 
-def napalm_rollback(task: Task, rollback: bool = False) -> Result:
+def napalm_rollback(task: Task, dry_run: bool = None) -> Result:
     """
     Rollback device configuration using napalm
 
     Arguments:
-        rollback: whether to rollback the configuration or not
+        dry_run: whether to rollback the configuration or not
     """
     device = task.host.get_connection("napalm", task.nornir.config)
 
-    if rollback:
+    if dry_run:
         device.rollback()
 
     return Result(host=task.host)
--- a/nornir/plugins/tasks/networking/napalm_rollback.py
+++ b/nornir/plugins/tasks/networking/napalm_rollback.py
@@ -1,0 +1,18 @@
+from typing import Optional
+
+from nornir.core.task import Result, Task
+
+
+def napalm_rollback(task: Task, rollback: bool = False) -> Result:
+    """
+    Rollback device configuration using napalm
+
+    Arguments:
+        rollback: whether to rollback the configuration or not
+    """
+    device = task.host.get_connection("napalm", task.nornir.config)
+
+    if rollback:
+        device.rollback()
+
+    return Result(host=task.host)
--- a/nornir/plugins/tasks/networking/napalm_rollback.py
+++ b/nornir/plugins/tasks/networking/napalm_rollback.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 from nornir.core.task import Result, Task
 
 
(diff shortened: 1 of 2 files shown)


nornir/plugins/tasks/networking/napalm_rollback.py
```diff
@@ -0,0 +1,16 @@
+from nornir.core.task import Result, Task
+
+
+def napalm_rollback(task: Task, rollback: bool = False) -> Result:
+    """
+    Rollback device configuration using napalm
+
+    Arguments:
+        rollback: whether to rollback the configuration or not
```
dbarrosop: Instead of a `rollback` parameter it'd be nice to have a `dry_run` instead, pretty much following the same pattern as `napalm_configure`: https://github.com/nornir-automation/nornir/blob/fee75ff2902e1a6af72a6981b2dbbff9be7e290d/nornir/plugins/tasks/networking/napalm_configure.py

jchristopher327: OK, will do. Any preference on what the default should be for `dry_run`? I'm thinking `True` since false won't doesn't provide any feedback to the user on what changes will be rolled back.

dbarrosop: Default should be `None` like with `napalm_configure`.

jchristopher327: Ok, all set.

dbarrosop:
It'd be also nice to create a jupyter notebook to document and test this. Not sure if you are familiar with it. This serves as both documentation (https://nornir.readthedocs.io/en/stable/plugins/tasks/version_control/gitlab.html) and also tests. I can help out with the details of dealing with jupyter notebooks and mocking napalm if you unfamiliar with that.

dbarrosop:
If you don't have the time for that let me know.

jchristopher327:
Unfortunately I'm not familiar with Jupyter, but have the time to get familiar. I was just looking at a few of the howto examples. Is there one that shows mocked configuration data? I see the handling connections howto showing mocked data but it's for facts and not config.

In any case, let me know how to best proceed.

dbarrosop:
For config it's pretty much the same way. Check here;

https://github.com/napalm-automation/napalm-ansible/tree/develop/tests/napalm_install_config/mocked/replace.dry_run.change

and

https://github.com/napalm-automation/napalm-ansible/tree/develop/tests/napalm_install_config/mocked/merge.error

The compare_config is the only method that will return something, the rest will either be just an empty dict or an error as in the second example.

jchristopher327:
I just took a look over the two links and unfortunately I can't make sense of it :(

dbarrosop:
Mmm, any questions or anything I can help with? Happy to help but I'd need to know what you did/tried or if there is anything in particular that is confusing.

jchristopher327:
This is one of those "I don't have any questions because I didn't understand it" situations as I'm not familiar with tests or mocking data. When I looked at the two links you sent, I couldn't figure out how that was able to test whether the rollback code works or not.

dbarrosop:
1. Here is an example of a jupyter notebook: https://github.com/nornir-automation/nornir/tree/develop/docs/plugins/tasks/version_control
2. To mock napalm you use the mock driver, specify the path to the mocked data in the inventory and then mock the data. For instance, in the inventory:
```
connection_options:
    napalm:
         platform: mock
         extras:
               optional_args:
                     path: path/to/folder/with/mocked/data
```

Then, in that path, data like this: 
https://github.com/napalm-automation/napalm-ansible/tree/develop/tests/napalm_install_config/mocked/replace.dry_run.change

When a function doesn't return anything the contents is `{}` unless you are emulating an error.

You have plenty of examples here:

https://github.com/napalm-automation/napalm-ansible/tree/develop/tests/napalm_install_config/mocked

dbarrosop:

PR #388, opened 2019-05-01
Fix markup for links to external resources

vincentbernat:
Text and links are swapped in the "external resources" section.


dmfigol:
Thank you for the contribution!

PR #390, opened 2019-05-04
experiment with asyncio

dbarrosop:
A few things this PR tackles:

1. I don't like the relation between `Nornir` <-> `Task` and <-> the functions that the task calls. There are is basically a circular dependendency so I tried to clean that bit while playing with this. To avoid breaking things and allowing us to experiment I have decided to start a `beta` submodule where we can experiment with new features with changes to the API without guaranteeing stability.
2. With this change the idea is that `Nornir` is responsible of gluing all the data together (config, state, inventory) but it's not responsible of implementing any logic to run anything, that's responsibility of the `TaskRunner`.
3. I added a decorator to extend the logic of how to run tasks (added a `with_retries` that basically retries tasks N times if an exception is returned)
4. Added support for async tasks

See [test.py](https://github.com/nornir-automation/nornir/pull/390/files#diff-b284a28710cce90d9d9be3a7f4cabc8e) for examples.

The `TaskRunner` object will get a bunch of parameters to the class itself (the function to run,  num_workers, hosts to run, etc.) and then its `run` method will receive another bunch of arguments that will basically be passed as they are to the function to run. For instance:

```
result = TaskRunner(
    nornir=nornir.filter(F(...)),
    task=napalm_get,
).run(
    filters=["interfaces"]
)
```

Note: This code is far from complete, we need to figure out which data to return and how.


dbarrosop:

PR #392, opened 2019-05-08
Fix netbox pagination

wvandeun:
Fixes netbox pagination.
This fixes the same issue as #361. The owner of that PR didn't have the time to continue to work on the PR & I needed a solution so I implemented a solution myself.
I have discussed this @bdlamprecht and he was ok with the approach.

### DIFF
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -13,6 +13,7 @@ class NBInventory(Inventory):
         nb_url: Optional[str] = None,
         nb_token: Optional[str] = None,
         use_slugs: bool = True,
+        ssl_verify: bool = True,
         flatten_custom_fields: bool = True,
         filter_parameters: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
@@ -37,6 +38,7 @@ class NBInventory(Inventory):
 
         session = requests.Session()
         session.headers.update({"Authorization": f"Token {nb_token}"})
+        session.verify = ssl_verify
 
         # Fetch all devices from Netbox
         # Since the api uses pagination we have to fetch until no next is provided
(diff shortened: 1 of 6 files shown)


dbarrosop:
Thanks for this work, can we create a new test without removing the old one? New code should make sure the old test also passes. Thanks again!

wvandeun:
The old test still passes. 
I just deleted some lines which were commented out and not useful in any way. 
The signature of the get_inv function changed: I added a parameter to enable/disable pagination.

If you want I can just create another get_inv_pagination function, so that the old test is still exactly the same.

nornir/plugins/inventory/netbox.py
```diff
@@ -34,19 +35,30 @@ def __init__(
...
-        r.raise_for_status()
-        nb_devices = r.json()
+        session = requests.Session()
+        session.headers.update({"Authorization": f"Token {nb_token}"})
+        session.verify = ssl_verify
+
+        # Fetch all devices from Netbox
+        # Since the api uses pagination we have to fetch until no next is provided
+
+        url = f"{nb_url}/api/dcim/devices/?limit=0"
```
dbarrosop: is `limit=0` needed?

dbarrosop:
You are right, sorry, the diff mislead me and made me think you had deleted things you hadn't. It's all good then, just a minor question due to my lack of operational knowledge with netbox

nornir/plugins/inventory/netbox.py
wvandeun: It is not strictly required. However if not specified then Netbox will return the number of records configured in PAGINATE_COUNT (50 by default). If limit is set to 0 then the Netbox will return the maximum number of records specified by MAX_PAGE_SIZE configuration.
50 is a bit low and would take quite some requests in a reasonable sized inventory.

wvandeun: https://netbox.readthedocs.io/en/latest/api/overview/#pagination

dbarrosop: ok, thanks for the clarification. Merging :)

bdlamprecht:
Nice, great work @wvandeun!
All it takes is someone who knows what their doing and has the time.

Issue #393, opened 2019-05-14
to_dict method of nornir.core.state.GlobalState class returns AttributeError


brandomando:
It appears that the attribute `__dict__` is not implemented in the class `GlobalState`:
```
In [43]: nr.data.to_dict()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-43-2a86f21844a7> in <module>
----> 1 nr.data.to_dict()

c:\networkdevops\nornir\nornir\core\state.py in to_dict(self)
     27     def to_dict(self) -> Dict[str, Any]:
     28         """ Return a dictionary representing the object. """
---> 29         return self.__dict__

AttributeError: 'GlobalState' object has no attribute '__dict__'
```
Looking at the source, this attribute is not defined

brandomando:
The `Nornir.dict()` method is also broken because it returns `return {"data": self.data.dict(), "inventory": self.inventory.dict()}`.  Additionally, `Nornir.core.Inventory` class does not have a `dict()` method.  However, the newly implemented `get_inventory_dict()` method of the `Inventory` class would return the dict that I assume the `dict()` method would have returned if implemented, so perhaps just changing the reference would fix the inventory dict issue.

dbarrosop:
Yeah, this is probably a regression and we broke it while refactoring the way the inventory was being serialized into python objects.

dbarrosop:
This is fixed in #394

PR #394, opened 2019-05-15
Fix Nornir dict method and dependencies

brandomando:
Ref https://github.com/nornir-automation/nornir/issues/393

* Add `dict()` method to `Inventory` class.
* Point `get_inventory_dict` to `dict()` method in Inventory
* Refactor `to_dict()` method of `GlobalState` and rename to `dict()` to match reference in `Nornir` class


dbarrosop:
Awesome, thanks!

PR #414, opened 2019-07-21
Add netmiko_commit task

jrokeach:
Add netmiko_commit task to support ability to write to two-stage config devices via netmiko.

### DIFF
--- a/nornir/plugins/tasks/networking/__init__.py
+++ b/nornir/plugins/tasks/networking/__init__.py
@@ -2,6 +2,7 @@ from .napalm_cli import napalm_cli
 from .napalm_configure import napalm_configure
 from .napalm_get import napalm_get
 from .napalm_validate import napalm_validate
+from .netmiko_commit import netmiko_commit
 from .netmiko_file_transfer import netmiko_file_transfer
 from .netmiko_send_command import netmiko_send_command
 from .netmiko_send_config import netmiko_send_config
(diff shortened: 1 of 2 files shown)


ktbyers:
I think this looks fine except that we should allow the passing of arguments to the commit method in a similar manner to what we did here:

https://github.com/nornir-automation/nornir/blob/develop/nornir/plugins/tasks/networking/netmiko_send_command.py#L11

jrokeach:
Thanks for the feedback @ktbyers. I'll provide an update.

dbarrosop:
what does this do? Shouldn't this be an option in `netmiko_send_config`?

jrokeach:
> what does this do?

Calls the commit() method of netmiko to apply staged configuration changes on platforms like XR, JunOS, PANOS, etc.

>Shouldn't this be an option of `netmiko_send_config`?

I think to the point of passing args, there are a few different options that could be passed to the netmiko commit() method and it could be used a few different ways. 

I'm thinking the most practical use case is with grouped tasks - if it's a two-stage commit platform, then group the netmiko_send_config and netmiko_commit tasks together. But if not we don't want to call netmiko_commit. Thoughts?

ktbyers:
It is a different method in Netmiko that is really specific to Juniper, IOS-XR, and I think one other platform. It would be hard to integrate it to the existing `netmiko_send_config` as it would be hard to determine which arguments were for which Netmiko method (i.e. which arguments were for `send_config_set` and which arguments were for `commit`). It might be a rare enough case though that you could just do it through a custom Nornir task.

dbarrosop:
Ok, thanks for the clarification. @ktbyers feel free to squash and merge when you feel this is good :)

PR #416, opened 2019-08-02
NETCONF prototype

dbarrosop:
This is a prototype of a NETCONF plugin. It'd be great if the community could provide feedback and help testing it.

At this point it is mostly a shim layer. It doesn't even provide tasks, just a connection plugin. The way it can be used now is by just getting the connection and calling manager's methods directly (https://ncclient.readthedocs.io/en/latest/manager.html#ncclient.manager.Manager)

There is a bit of code here in https://github.com/nornir-automation/nornir/pull/416/files#diff-b284a28710cce90d9d9be3a7f4cabc8e


ogenstad:
I think the connection plugin looks pretty good as is, it would be nice a container to run tests against. Perhaps this one or something similar (haven't tried it):
https://hub.docker.com/r/sysrepo/sysrepo-netopeer2/

In a way I don't know if individual plugins to run the actions such as .get, .get_config, .edit_config, .commit etc is strictly necessary as it's so easy to just have something connect to the netconf connection plugin. On the other hand it might make sense to have a few netconf task plugins from a documentation point of view as it might be easier for people to find out how to do something over netconf with Nornir if it ships with plugins.

I can do some tests the coming days with this.

ogenstad:
I've tested this now and it works but with two issues. When connecting to a new host it would just fail without any clear reason unless I add `"hostkey_verify": False,` to the parameters, if I add the device first to ~/.ssh/known_hosts the hostkey_verify=False isn't needed. I think ncclient should present some error message for this and it should be fixed on that side. All their examples seem to use hostkey_verify=False but that just seems wrong.

The other problem was that when I have a key in ~/.ssh/id_rsa the connection just failed which should also have been picked up by ncclient.

With this inventory I was able to connect without any issues.

```yaml
---
nso:
    hostname: 192.168.16.20
    username: admin
    password: admin
    port: 2022
    connection_options:
        netconf:
            extras:
                allow_agent: False
                hostkey_verify: False
```

I think some documentation that mentions this could be good.

PR #418, opened 2019-08-05
Enable mypy for nornir.core.task

ogenstad:
Added type checking for nornir.core.task. The changes to Host, with Union[Host, None] had to do with the http_method task.


dbarrosop:
awesome!

PR #437, opened 2019-09-28
Github Task Plugin

fach:
This PR adds support for github as a version control backend in Nornir as a task plugin. The following operations are supported against a repository:

- Creating files (w/ dry-run support)
- Updating files (w/ dry-run support)
- Getting files (w/ dry-run support)

Tests are also included.

### DIFF
--- a/docs/plugins/tasks/version_control/gitlab.ipynb
+++ b/docs/plugins/tasks/version_control/gitlab.ipynb
@@ -36,7 +36,7 @@
       "\n",
       "    Example:\n",
       "\n",
-      "        nornir.run(files.gitlab,\n",
+      "        nornir.run(version_control.gitlab,\n",
       "                   action=\"create\",\n",
       "                   url=\"https://gitlab.localhost.com\",\n",
       "                   token=\"ABCD1234\",\n",
--- a/nornir/__init__.py
+++ b/nornir/__init__.py
@@ -1,6 +1,7 @@
+from nornir.init_nornir import InitNornir
+
 import pkg_resources
 
-from nornir.init_nornir import InitNornir
 
 __version__ = pkg_resources.get_distribution("nornir").version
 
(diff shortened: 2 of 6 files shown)


fach:
It looks like CI is passing except for sphinx due to an issue during setup:

https://travis-ci.org/nornir-automation/nornir/jobs/590805998

Does someone mind triggering a restart for this PR? I don't think I can unless I push another commit.

wvandeun:
Hi,

I've worked on the gitlab plugin and was thinking about changing the API.
Basically I want to step away from having a single "factory" function/task with an action argument. 

I'm doing this mainly because depending on the action you specify, you have to provide certain arguments to the function. I think it would be better to have a task/function for each action, because it is clearer which arguments you would have to specify by just inspecting the function definition. 

 It is still work in progress, but maybe you should have a look and provide feedback. It would be best if both plugins use a similar approach.
https://github.com/wvandeun/nornir/commit/95f88724aefd67e78ded37854c5134598fa97b65

dbarrosop:
> I think it would be better to have a task/function for each action

I agree with that. I think we need to start by identifying what we want to support, what can be done with generic `git` tasks and what needs specific `git{hub, lab}` tasks

fach:
> Basically I want to step away from having a single "factory" function/task with an action argument
I support this. In hindsight I should have opened up a bug to discuss first because I considered going in this direction with the github plugin but decided against it in favor of velocity and symmetry with the gitlab plugin.

> I think we need to start by identifying what we want to support, what can be done with generic git tasks and what needs specific git{hub, lab} tasks

Agreed. The API between github and gitlab are extremely similar. It seems like we could deliver a common git* API for the most part.

wvandeun:
What I need, or what I had implemented was:
- creating a file
- updating a file
- get/download a file
- check if a file exists
Without having git toolset installed and without having the complete repository locally.
I think it would make sense to implement a task to delete or move a file.

dbarrosop:
> what I had implemented

That proposal looks good to me

> A BackupToGitlab processor

Yes, I agree this could be better implemented either as a processor or just as a function. If the function had a lock you could either call it within your own task or after it.

dbarrosop:

PR #438, opened 2019-10-04
Fix netbox unnamed devices

wvandeun:
Devices in Netbox don't need to have a name defined. This is usefull
to model for example switch stacks, for which individual members don't
neceseraly need a name.
In such cases we will define the Nornir host's name to the device id
in Netbox.


dbarrosop:
Awesome, thanks!

Issue #429, opened 2019-08-22
KeyError when setting a "False" Data value in defaults.yaml


aSauerwein:
Example:
defaults.yaml:
```
---
data:
    key3: value3
    key4: false
```
Python code: 
```
for host, host_obj in nr.inventory.hosts.items():
    print(f'{host}: {host_obj["key3"]}')
    print(f'{host}: {host_obj["key4"]}')
```

host_obj["key4"] will fail with "KeyError"

the reason for this is in inventory.py line 161:
https://github.com/nornir-automation/nornir/blob/develop/nornir/core/inventory.py
```
r = self.defaults.data.get(item)
    if r:
        return r
```
self.defaults.data.get(**KEY4**) = False
and therefore "if r" resolves to False => KeyError. 

potential fix:
```
if r != None:
    return r
```

dbarrosop:
Closed via #443

PR #443, opened 2019-10-18
Fix KeyError when setting a "False" val in defaults

k01ek:
https://github.com/nornir-automation/nornir/issues/429


dbarrosop:
Awesome, thanks!

PR #444, opened 2019-10-18
Add netconf plugins

ogenstad:
* netconf_edit_config
* netconf_get
* netconf_get_config


dbarrosop:
I still need to review but I think you need to add them to the docs.

ogenstad:
Currently, I’m pretty sure all of the networking plugins are imported at
the same time. Was thinking it might be a bit cluttered, but not related to
this PR.

dbarrosop:
Awesome, thanks!

PR #445, opened 2019-10-25
Update to support Netmiko 3.x.x

carlmontanari:
Updated netmiko version --> `>=2.4.2`
Updated paramiko version --> `^2.3.2`

Updating Poetry pinned version of netmiko in support of upcoming netmiko 3.x release which includes significant reliability improvements. Also have updated paramiko version as the previously pinned version was getting pretty old. 

I did notice that after doing `poetry update` the `mypy` version in the lock file went up to `.740` and running `make tests` threw a few mypy errors. So this PR does *not* have the lock file updated/I didn't run `poetry update`.

cc @ktbyers


carlmontanari:
Updated name to remove the POC Do Not Merge bits. As CI passed w/out the mypy issues I saw locally I suspect this should be good to go pending review!

dmfigol:
Are there any breaking changes in 3.x? We could just do `^3` if not.

ktbyers:
@dmfigol I wanted to avoid having to coordinate Nornir and Netmiko...right now Netmiko 3.0.0 is not released as I wanted this approved/merge before I released it (as current pyproject.toml pattern in Nornir disallows Netmiko 3.x.x).

The main change in Netmiko 3.0.0 is it no longer allows Python 2.7 (or Python 3.4/3.5). That is obviously a no-op for Nornir though.

We also made a lot of reliability improvements in Netmiko 3.0.0 so it is possible that those changes broke something on some Netmiko platforms. We have tested Netmiko 3.x.x a lot, but the set of platforms we can test against is relatively narrow.  Consequently, I wouldn't want to to disallow Netmiko 2.4.2 use with Nornir.

Understood on the lockfile. We were more concerned about end-users being able to install/use Netmiko 3.0.0 in conjunction with Nornir.

So we are fine either way with the lockfile. If you want it updated--no worries we can update it. Just let us know the preference. 

There was something in the docs indicating Poetry lockfile only got updated at release--so we intentionally didn't update it.

dmfigol:
@ktbyers thanks for detailed explanation. 
Yes, it seems that `netmiko>=2.3.3` will be the best in this situation.
I think we do need an update to the lockfile (especially since Carl noticed that mypy broke - that should be taken care of as a separate PR probably), because when a user does `pip install nornir`, it will install the latest versions of packages which satisfy the package metadata coming from `pyproject.toml` (after this PR - latest netmiko), BUT our CI will still use whatever is in the lockfile.
@dbarrosop thoughts?

ktbyers:
Sounds good. We'll update the lock file and see if we can fix the mypy issue in a separate PR.


Kirk

carlmontanari:
Ok, have updated the lock file here (though not really much changed since its still 2.4.2 installed). Also opened #448 for the mypy bits (and did not update mypy in this lock file).

Let me know if there is anything else that needs to get sorted!

carlmontanari:
I'm going to defer to @ktbyers :), but seems like it would be cool to wait so that the improvements of 3.x can be had in Nornir-land -- and at least w/ this we are ready to go minus just updating the lock file.

ktbyers:
I would vote to just merge it...I don't think we are going to catch much/anything in the Nornir tests (testing 2.4.2 versus 3.0.0). Carl already manually tested it and we are already using Netmiko 3.0.0 (develop branch) in our Nornir course.

And we ran the full test suite there (i.e. the full-test suite we have for our course using Nornir + Netmiko 3.0.0).

dmfigol:
Thanks @carlmontanari for the PR and @ktbyers for the input!

PR #448, opened 2019-10-28
Minor Typing updates to appease mypy .740

carlmontanari:
Related to #445 

Updates to mypy caused some typing complaints to crop up in two task plugins.

1) In `command.py` mypy did not like that we were going from bytes -> string without being more explicit about it. Modified this to have `bytes_stdout` and `bytes_stderr` come back from `cmd.communicate` and then assigned `stdout` and `stderr` the decoded value of this.

2) In `load_json.py` it looks like there was maybe an idea at some point to pass kwargs around, but since it seems that was not used (and was causing a mypy error) I just removed that bit.

I believe I've got poetry bits updated for just mypy (though there are some hashes updated too which I assume is ok), but let me know if I buggered it up.


dmfigol:
LGTM

PR #449, opened 2019-11-08
Update pydantic to v1

### LINKED ISSUE #447: Update pydantic to 1.0
Finally a major release of pydantic is out https://github.com/samuelcolvin/pydantic/releases
It would be great to finally pin the last dependency to a major version
Related to #428

AchilleAsh:
Fix #447 

The only issue (I hope) is in the configuration doc generation using sphinx. Pydantic changed its schema output to use _$ref_ and the 
```docs/_data_templates/configuration-parameters.j2``` template can't handle references to definitions.

Because the models schemas are overriden using ```Field```, the _$ref_ key is in an _allOf_ array ( see https://github.com/samuelcolvin/pydantic/blob/master/pydantic/schema.py#L723 ) so the Jinja template has to do extra work to extract the $ref:

```python 
{'core': {
          'title': 'Core', 
          'default': {'num_workers': 20, 'raise_on_error': False}, 
          'env_names': ['nornir_core'],
          'allOf': [{'$ref': '#/definitions/CoreConfig'}]
           }
```
This references a definition named _CoreConfig_ in the definitions dict. If the Jinja2 _with_ notation is too cryptic (extract the first _$ref_ in _allOf_ then split to only get _CoreConfig_ here) , It could be instead be done in pure python before rendering the template.

I've also replaced the deprecated _Schema_ by _Field_ as per Pydantic deprecated warning


dbarrosop:
Nice work, looks like there are a couple of things to fix for the CI to pass (mostly cosmetics), let me know if you need help.

Thanks a lot for the PR!

AchilleAsh:
Sorry, I forgot to run mypy and nbval tests. I've fixed a few things, everything is green now.

dbarrosop:
Awesome work, thanks!

PR #450, opened 2019-11-11
Update confusing tutorial ordering

JStation:
**Change**
Move inventory section one step earlier in front of initialization section

**Reasoning**
Initializing Nornir before creating an inventory file throws a FileNotFoundError for the missing inventory/hosts.yaml file. This was unexpected as a new user moving through the documentation linearly. An easy solution is to introduce the reader to the inventory section before the initialization section. Otherwise, a note on expecting the error would suffice.

### DIFF
--- a/docs/tutorials/intro/index.rst
+++ b/docs/tutorials/intro/index.rst
@@ -10,8 +10,8 @@ We're glad you made it here! This is a great place to learn the basics of Nornir
    Nornir at a glance <overview>
    100% Python <python>
    Installation guide <install>
-   initializing_nornir.ipynb
    inventory.ipynb
+   initializing_nornir.ipynb
    executing_tasks.ipynb
    grouping_tasks.ipynb
    task_results.ipynb


dbarrosop:
Thanks for your PR, looking at it I agree with the problem but I would prefer to keep the order because the inventory tutorial is a bit large and I'd rather start simple. The inventory tutorial also assumes you know how to initialize the inventory.

An alternative solution would be to add very simple instructions to create a few files for the inventory. What do you think?

JStation:
I see your point. The complete inventory tutorial page is a bit of a (necessarily) long read for someone just trying to get their feet wet. 

I like the idea of beginning the initialization page with very simple inventory files or even a single file, if possible. The best way I can think to implement that would be to split the inventory page into two sections, one of which could be moved to the initializing section.

I'm thinking a structure like this:
**Initializing Nornir**
- Basic Inventory (for a single host - without a groups file, if possible)
- Config File
- InitNornir

**Inventory**
- Overview w/ mention of SimpleInventory
- Additional Inventory (groups, data, inheritance, etc)
- Accessing Inventory (w/ nornir object) <- subsequently broken into accessing sections

I also noticed that the last line of the installation page sets the reader up for creating an inventory file as the next step. Depending on how this is resolved, this may need changed as well.
pic-selected-191112-0857-51

dbarrosop:
Ok, I have been looking into this a bit further and I think I know where the problem is and why you got the error you got. The tutorial is designed to work as it is by using the Jupyter notebooks provided. Alternatively, you could run things on your own following the instructions but you have to be in the right [folder](https://github.com/nornir-automation/nornir/tree/develop/docs/tutorials/intro). That's the reason why it's not going into details about the inventory just yet.

So I think the solution is to leave things as they are but adding instructions to either use the Jupyter notebooks or to change into that folder before starting.

PR #455, opened 2019-11-28
Update gitlab.py

### LINKED ISSUE #454: gitlab plugin _get_repository problem
The problem is function _get_repository search request. If repository path contains group like 'diaspora/diaspora-project-site' search does not work. 
If search only by name 'diaspora-project-site' result may contains list of repo: 
'test/diaspora-project-site'
'diaspora1/diaspora-project-site'
'blabla/diaspora-project-site'
and function return only first PID (project id). 
As said api docs https://docs.gitlab.com/ee/api/projects.html it is posible to use URL-encoded path of the project instead PID.
I will make PR for fix it.

k01ek:
Fix #454

### DIFF
--- a/nornir/plugins/tasks/version_control/gitlab.py
+++ b/nornir/plugins/tasks/version_control/gitlab.py
@@ -36,7 +36,7 @@ def _get_repository(session: requests.Session, url: str, repository: str) -> str
     for p in respjson:
         if p.get("name", "") == repository:
             found = True
-            pid = p.get("id", "0")
+            pid = str(p.get("id", "0"))
 
     if pid == "0" or not found:
         raise RuntimeError("Gitlab repository not found")
--- a/nornir/plugins/tasks/version_control/gitlab.py
+++ b/nornir/plugins/tasks/version_control/gitlab.py
@@ -1,7 +1,8 @@
 import base64
 import difflib
 import threading
-import urllib
+
+from urllib.parse import quote
 from pathlib import Path
 from typing import Tuple
 
@@ -22,7 +23,7 @@ def _generate_diff(original: str, fromfile: str, tofile: str, content: str) -> s
 
 def _get_repository(session: requests.Session, url: str, repository: str) -> str:
     if "/" in repository:
-        pid = urllib.parse.quote(repository, safe="")
+        pid = quote(repository, safe="")
         return pid
     resp = session.get(f"{url}/api/v4/projects?search={repository}")
     if resp.status_code != 200:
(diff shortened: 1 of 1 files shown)


k01ek:
@wvandeun can you review?

nornir/plugins/tasks/version_control/gitlab.py
```diff
@@ -19,15 +19,15 @@ def _generate_diff(original: str, fromfile: str, tofile: str, content: str) -> s
     return "\n".join(diff)
 
 
-def _get_repository(session: requests.Session, url: str, repository: str) -> int:
+def _get_repository(session: requests.Session, url: str, repository: str) -> str:
     if "/" in repository:
         pid = repository.replace("/", "%2F")
```
wvandeun: Can you use `urllib.parse.quote()` to the URL encoding, instead of just encoding a forward slash manually.
The same will need to happen for file path btw.

nornir/plugins/tasks/version_control/gitlab.py
```diff
@@ -36,16 +36,16 @@ def _get_repository(session: requests.Session, url: str, repository: str) -> int
     for p in respjson:
         if p.get("name", "") == repository:
             found = True
-            pid = p.get("id", 0)
+            pid = p.get("id", "0")
```
wvandeun: Why are you converting the default value to a str here? This doesn't make sense to me.

nornir/plugins/tasks/version_control/gitlab.py
```diff
@@ -36,7 +36,7 @@ def _get_repository(session: requests.Session, url: str, repository: str) -> str
     for p in respjson:
         if p.get("name", "") == repository:
             found = True
-            pid = p.get("id", "0")
+            pid = str(p.get("id", "0"))
```
wvandeun: Again, I don't see why you are converting pid to a str here?

nornir/plugins/tasks/version_control/gitlab.py
k01ek: Its make sense for type annotations...

nornir/plugins/tasks/version_control/gitlab.py
k01ek: Sure, i will change it also for the filename..

dbarrosop:
Nice work, I have a question (due to ignorance), I see you changed the type of `pid` from `int` to `str`, I don't have strong opinions about it other than to be consistent with gitlab's documentation. Do they refer to pid as a string or as an int?

wvandeun:
The API documents it as string or integer.
https://docs.gitlab.com/ee/api/projects.html#get-single-project

But the JSON payload always represents id as an integer.

I don't see the point of converting it to a string.

k01ek:
It doesn't matter for this code, cause under the hood we use pid as string in call:
f"{url}/api/v4/projects/{pid}/repository/files/{filename}?ref={ref}"
I think,  for order, if method annotate result as 'str', it is better to use 'str' as result..

k01ek:
Another option is to use pid as empty string if search does not get result.

dorko20:
Hi guys @k01ek @dbarrosop @wvandeun ,   
first of all, really great job, I really enjoy Nornir.   

I would like to ask you when this change will be merged into the latest nornir release?  
I am facing the same problem and @k01ek solution works ok for me.

Thanks!

k01ek:
Hi @dorko20 . @wvandeun has open new PR for fixing this problem.
I will close that

PR #456, opened 2019-12-11
add possibility to use netbox config context fields in nornir

smnmtzgr:
I think it is a correct solution to make it possible to use config_context fields which are defined in netbox. The netbox docu tells:

> Sometimes it is desirable to associate arbitrary data with a group of devices to aid in their configuration. For example, you might want to associate a set of syslog servers for all devices at a particular site. Context data enables the association of arbitrary data to devices and virtual machines grouped by region, site, role, platform, and/or tenant. Context data is arranged hierarchically, so that data with a higher weight can be entered to override more general lower-weight data. Multiple instances of data are automatically merged by NetBox to present a single dictionary for each object.
Devices and Virtual Machines may also have a local config context defined. This local context will always overwrite the rendered config context objects for the Device/VM. This is useful in situations were the device requires a one-off value different from the rest of the environment.

With my change the netbox plugin will map these variables directly to host["data"] (flatten_config_context_fields = True (default)) or to host["data"]["config_context"] (flatten_config_context_fields = False)

### DIFF
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -15,7 +15,7 @@ class NBInventory(Inventory):
         use_slugs: bool = True,
         ssl_verify: Union[bool, str] = True,
         flatten_custom_fields: bool = True,
-        flatten_config_context_fields: bool = True,
+        flatten_config_context_fields: bool = False,
         filter_parameters: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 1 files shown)


dbarrosop:
The CI is failing, please, review your code and make sure it works. A unit test testing your feature to avoid regressions would also be welcome.

nornir/plugins/inventory/netbox.py
```diff
@@ -15,6 +15,7 @@ def __init__(
         use_slugs: bool = True,
         ssl_verify: Union[bool, str] = True,
         flatten_custom_fields: bool = True,
+        flatten_config_context_fields: bool = True,
```
dbarrosop: default value should be false to avoid changing existing behavior

ktbyers:
Closing as this PR looks no longer applicable to nornir-core.

PR #459, opened 2019-12-29
AnsibleInventory Enhancements: Dynamic Inventory, Multiple Sources, Merge

carlmontanari:
Apologies in advance for a fairly heavy PR w/ no issue/notice; feel free to close or let me know if slicing this up into a few smaller chunks would be more desirable for the project.

General purpose of this PR is to add more functionality to the Ansible inventory plugin -- primarily to support dynamic inventory scripts. A few other minor things came up while working on this as well...

Minor changes:

- Added `.idea` to gitignore
- Modified Dockerfile slightly -- I'm not 100% on this, but pretty sure this was just a Poetry syntax change... modified `poetry config settings.virtualenvs.create false` --> `poetry config virtualenvs.create false`. I've not really used Poetry but all the tests and such seem to work as desired and I don't get whatever error complaining about invalid setting anymore!
- Modified pyrpoject.toml setting for Black `allows-prereleases` --> `allow-prereleases` -- this was another one that just popped up saying it was/is being deported (the "allows" bit)

Inventory/Ansible related changes:

- Created a `NornirNoValidInventoryError` Exception if no valid inventory sources could be parsed. Raising the `ruamel.yaml` exceptions didn't make sense after adding in the `ScriptParser` to the Ansible inventory bits -- so I figured something more descriptive would make sense/be useful.
- Added `ini` and `py` to `VARS_FILENAME_EXTENSIONS` for valid ansible inventory file extensions. Since there are plugins handy on vi/sublime/pycharm/etc. for INI files figured it would be nice to support that, and added .py of course for dynamic inventory scripts.
- Created a `ScriptParser` class that inherits from the AnsibleInventory class like the INI/YAML parsers. This was modeled off of Ansible core code but obviously shoehorned into Nornir a bit.
- Added two args to the `AnsibleInventory` class:
  - `inventory` -- hosts file felt wrong for dynamic inventory and also for multiple inventory sources (more on that in a sec); also wanted to ensure I didn't bugger up existing `hostsfile` default behavior
  - `hash_behavior` -- optional arg for how to handle combining inventories. defaults to `replace` as per normal Ansible and Nornir behavior, but supports `merge` as well
- Added functions to handle hash stuff as outlined above -- these were pretty much lovingly pulled right out of Ansible code (do we need some kind of attribution or since its just a tiny part its cool? I don't know how that works...)
- Pretty heavily changed `parse` function to support `hostsfile` and `inventory`, as well as `inventory` being a comma sep list of inventory locations to check. This also raises the new exception if no valid inventories existed, and deals with combining inventories based on the hash behavior setting.
- Added tests to validate dynamic scripts work, merge behavior works as desired, and multiple inventory sources work as desired

Outstanding Issue(s):

~~- Mypy is complaining about typing around the hash stuff -- MutableMapping has no attribute copy... thought typing things as a dict would be fine but clearly not, typing as MutableMapping gives the copy attribute complaint. I will look at this more tomorrow to try to resolve, if I don't figure it out I could use some help/explanation on how to solve this one if somebody would like to help me out here!~~ No more complaints here, but fix feels pretty hacky...

Other thoughts:

~~- Not sure if dynamic inventory scripts ever return things in a different format -- this is molded off of the scripts I have currently handy and seem to be pretty much standard.~~ Removed the is instance list check... pretty sure it'll always be lists in the two places I had this in the original commit. If its ever an issue its an easy fix at any rate.
- Would be good to support (if we don't already.... tbh didn't check but feel like this was a thing before) directories for each host/group within the host/group vars directories -- i.e for host `foo`; support `host_vars/foo/somefileshere.yaml`. Will check that out next weekend probably :) 

Carl

~~Edit: Looks like I missed a pylama complaint as well -- will update and fix tonight!~~ Sorted.


nornir/plugins/inventory/ansible.py
```diff
@@ -242,24 +253,178 @@ def load_hosts_file(self) -> None:
...
 
 
 class AnsibleInventory(Inventory):
-    def __init__(self, hostsfile: str = "hosts", *args: Any, **kwargs: Any) -> None:
-        host_vars, group_vars, defaults = parse(hostsfile)
+    def __init__(
+        self,
+        hostsfile: str = "hosts",
+        inventory: str = "",
+        hash_behavior: str = "replace",
```
dbarrosop: would you mind adding a docstring describing the options? Specially `hash_behavior`; what are the possibilities and what do they do?

nornir/plugins/inventory/ansible.py
```diff
@@ -242,24 +253,178 @@ def load_hosts_file(self) -> None:
...
+            new_v: Dict[str, Any] = dict(v)
+            result[k] = merge_hash(result[k], new_v)
+        else:
+            result[k] = v
+    return result
 
-    return parser.hosts, parser.groups, parser.defaults
+
+def replace_hash(a: Dict[str, Any], b: Dict[str, Any]) -> Dict[str, Any]:
+    result = a.copy()
```
dbarrosop: shouldn't we be using deepcopy to avoid copying references at the second level?

nornir/plugins/inventory/ansible.py
```diff
@@ -242,24 +253,178 @@ def load_hosts_file(self) -> None:
...
+def parse(
+    hostsfile: str, inventory: str = "", hash_behavior: str = "replace"
+) -> Tuple[HostsDict, GroupsDict, DefaultsDict]:
+    possible_sources: List[str] = []
+
+    # add hostsfile to to sources to preserve existing functionality
+    if Path(hostsfile).expanduser().is_file():
+        possible_sources.append(hostsfile)
+
+    inventory_locations = inventory.split(",") if inventory else []
```
dbarrosop: I am confused by this, why are you splitting this using the `,` as separator? What is the use case and reasoning? This wasn't supported before.

dbarrosop:
Outstanding job :)

I am in favor of supporting the dynamic inventory but main concern is that we are making this a bit too "flexible" which leads to (a) complexity in the code, (b) complexity to cover all the possible cases in testing due to its flexibility and (c) complexity even in the documentation.

Let me expand, the way I am interpreting the code (so correct me if I am wrong) you are adding support for:

1. Specifying multiple sources for the inventory, the source could even be a folder which means that all the files in the folder would even be included.
2. If the files are multiple formats they are combined, this includes even the newly added format (via the script)
3. There is also support for overlapping data with the possibility to merge/replace data in multiple places.
4. Everything is supported automagically without the user having to specify anything.

So my main concern is that this opens up a lot of scenarios and "what if" cases which makes it hard to reason about and test. Not sure if this functionality is supported in ansible and/or if we should care. I'd personally prefer something simpler that is easier to test and reason about unless there are very compelling reasons for them.

Do not take this comment as "let's scrap some of the functionality", I want to hear your thoughts about my comment first before deciding next steps.

Thanks again for the great job :D

nornir/plugins/inventory/ansible.py
carlmontanari: Done!

nornir/plugins/inventory/ansible.py
carlmontanari: In Ansible you are able to pass a list of inventories -- a very very handy feature. This is intended to mimic this. I did *not* do this on `hostsfile` to ensure I didn't break any current functionality/confuse people (that's basically the main reason for the second `inventory` argument). I think it'd be nice to just pass a list, but I assume that wouldn't work if this option was passed as an environment variable so I just rolled with comma sep string instead.

nornir/plugins/inventory/ansible.py
carlmontanari: Yes? I kinda thought the same, but this was pretty much lifted straight from Ansible source [here](https://github.com/ansible/ansible/blob/b31b5d105a84aa14b17302d2778c7606c78912bf/lib/ansible/utils/vars.py#L95). In the testing I've done so far (using this for some tests at work with a fairly big inventory too) I've not seen any issue.... but that doesn't mean its right!

Replacing copy w/ deep copy on the merge_hash and replace_hash functions works as expected in tests so I can update that if ya want, just let me know!

carlmontanari:
Totally understand on the complexity bits! As I was working on this I was wondering if it would be better to just have this as a second repo that could be installed alongside Nornir and just "pointed" to for inventory purposes (nornsible_inventory lol... another automation chimera!) ... at least that way the complexity would stay out of the main code base and was totally optional for folks. 

All your summary points above are correct. This basically puts the AnsibleInventory plugin at near  (maybe close to full?) parity with Ansible from an inventory perspective, so yeah these are all things in "normal" Ansible functionality. Whether or not we should care is up to you I think!

For me personally, I much prefer to use Nornir over Ansible, however I have always liked the ability to have multiple inventory sources -- dynamic script for say ec2 instances, and yaml file for routers or something like that, then pull it all together into one nice happy inventory.

If we needed to lose any of the functionality here in order to get it rolled into the code base I would definitely want to get rid of the `hash_behavior` stuff first. I don't like that feature in Ansible really, and I don't think it needs to be here -- I just added it since I was just trying to get close to full parity and it was easy enough to lift that from Ansible source and fit it into Nornir!

Let me know what changes you'd like to see and I can make it happen!

dmfigol:
@dbarrosop @carlmontanari 
Carl, thanks for your work on this, monumental effort!
I also have some concern about adding this to the main repo, it seems quite complex. Originally ansible inventory was added as means to simplify transition from Ansible to Nornir.
If we start adding things like this, if/when something does not work, people will come here complaining which we would like to avoid, because none of the maintainers are interested in ansible (correct me if I am wrong). We already have enough issues/questions about some netmiko/napalm behaviors not related to nornir.
My personal opinion: our Ansible inventory support should be limited to most common use-cases (ini, yaml with host_vars/group_vars folders), and be very clear what is supported and what is not in the docs.

> these were pretty much lovingly pulled right out of Ansible code (do we need some kind of attribution or since its just a tiny part its cool? I don't know how that works...)
We can't use that. Ansible uses GPL, which will make Nornir GPL and I doubt we want that.

Note: I didn't review the code deeply yet, mostly glanced through to understand what this PR does
Here is what I propose:
- your minor changes are good but probably should be in a separate PR - they are needed regardless of what we decide about Ansible
- your minor ansible changes (ini extension/exception) sound reasonable
- list of inventories - idk? I tend to say yes (though it is on the edge between "reasonable" and "too much"). If yes, with a single behavior (probably "replaced") clearly documented
- dynamic inventory - I'd say no. If a user is that advanced and can write a Python script for their inventory, what's stopping them from modifying it for Nornir custom inventory?

I also think ideas like this should not go to /dev/null. I think what we need is extensive plugin system, something like flask or django has, where a user can do something like 
`pip install nornir-ansible-inventory` and get this functionality.

carlmontanari:
@dmfigol @dbarrosop Thanks for the feedback! It all sounds reasonable and makes sense to me. This weekend I will make two new PRs so we can close this one out:

1. Minor fixes contained here so we get those sorted out
2. Slimmed down version of this -- basically removing the dynamic inventory but leaving list of inventories (and reswizzling the stolen Ansible bits to not be lifted straight from there so we don't have to worry about licensing stuff)

Figure PR1 is easy, PR2 we can discuss once its simplified (compared to this one) and then make any more tweaks there as needed!

Carl

dbarrosop:
> I think what we need is extensive plugin system, something like flask or django has, where a user can do something like `pip install nornir-ansible-inventory and get this functionality`

Yes, I think that's the way to go too

Thanks @carlmontanari for all the work and patience here and thanks @dmfigol for chiming in.

carlmontanari:
Closing -- replaced by #462 and #463 

Thanks for all the feedback!

PR #462, opened 2020-01-11
cleanup items from #459

carlmontanari:
Replaces part of #459 

Minor fixes discussed in other PR:

- Added .idea to gitignore
- Modified Dockerfile slightly -- I'm not 100% on this, but pretty sure this was just a Poetry syntax change... modified poetry config settings.virtualenvs.create false --> poetry config virtualenvs.create false. I've not really used Poetry but all the tests and such seem to work as desired and I don't get whatever error complaining about invalid setting anymore!
- Modified pyrpoject.toml setting for Black allows-prereleases --> allow-prereleases -- this was another one that just popped up saying it was/is being deported (the "allows" bit)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,5 +103,6 @@ output/
 
 .DS_Store
 .vscode
+.idea
 
 pip-wheel-metadata/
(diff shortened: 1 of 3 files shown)


dmfigol:
@carlmontanari All looks good! Thank you, Carl!

PR #463, opened 2020-01-11
Ansible Inventory - NornirNoValidInventoryError Exception

carlmontanari:
Replaces part of #459

Well I started looking at thinning down the giant PR I made before and I think I'll just end up putting most of that into another library that can be pip installed along-side Nornir as suggested. 

It just didn't seem worth it to have multiple inventory w/out dynamic inventory. Also if I left in the multiple inventory I would need to leave in the merge/replace functionality which is a bit wonky. So net/net I will work on getting that built in another pip-installable thing at some point!

I did want to keep two little enhancements from that effort though:

1- added `ini` to the `VARS_FILENAME_EXTENSIONS` as was in the previous PR (though I dropped `.py` of course)
2- Added the `NornirNoValidInventoryError` which is raised if neither INI or YAML parsers are able to parse a valid inventory. Updated the test to reflect this.

Let me know if this seems reasonable or if I should tweak anything! Thanks for all the feedback and guidance on these PRs!

Carl

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,5 +103,6 @@ output/
 
 .DS_Store
 .vscode
+.idea
 
 pip-wheel-metadata/
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ENV PATH="/root/.poetry/bin:$PATH" \
 RUN apt-get update \
     && apt-get install -yq curl git pandoc \
     && curl -sSL https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py | python \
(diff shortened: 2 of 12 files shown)


carlmontanari:
Ah looks like the build failed because I did not pull in the minor fixes from #462 (if not for other reasons too :)).

Whats the preferred way to sort this -- just add those fixes directly in this commit too or wait till the other PR is merged to dev and rebase (guess it wouldn't matter much if its the same fix, but was just trying to keep them in their own PRs).

dbarrosop:
Awesome, thanks again!

PR #466, opened 2020-01-13
Remove flake8-import-order

### LINKED ISSUE #465: pylama: Import statements are in the wrong order
Symptoms:
```
# ran in venv
-> % pylama nornir
nornir/__init__.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
nornir/init_nornir.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
nornir/core/configuration.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
nornir/core/task.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
nornir/core/connections.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
nornir/core/__init__.py:3:0: Import statements are in the wrong order. 'from nornir.init_nornir import InitNornir' should be before 'import pkg_resources' and in a different group. [import_order]
```
the same happens if you do `poetry run pylama`, `make tests` or `make pylama`.
According to @carlmontanari, for some reason it does not happen in CI.
I tried different combinations of `flake8-import-order` config, including:
```
[pylama]
application-import-names = nornir
```
```
[pylama:import_order]
application-import-names = nornir
```
and some others with no success.
This seems to be working fine in `flake8`:
```
# E501 is line length 79
-> % flake8 nornir --ignore=E501
nornir/core/__init__.py:14:5: F401 'nornir.core.inventory.Host' imported but unused
```
I am giving up trying to solve this and will submit PR which disables the linter with `# noqa` on the offending line.

dmfigol:
Fix #465

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -249,18 +249,6 @@ python-versions = ">=2.7"
 version = "0.3"
 
 [[package]]
-category = "dev"
-description = "Flake8 and pylama plugin that checks the ordering of import statements."
-name = "flake8-import-order"
-optional = false
-python-versions = "*"
-version = "0.18.1"
-
-[package.dependencies]
-pycodestyle = "*"
-setuptools = "*"
-
-[[package]]
 category = "main"
 description = "Clean single-source support for Python 3 and 2"
 name = "future"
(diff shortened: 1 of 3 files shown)


ktbyers:
I personally would completely disable the `import order` check as I just find it annoying, a waste of time.

But this is also fine with me. So  approving to fix this issue.

ktbyers:
+1

dbarrosop:
I agree with Kirk. If the "import order" check is not reliable let's disable it. I don't think we should add this `# noqa` there because it's misleading; there is nothing wrong with that import.

dmfigol:
I reworked PR, so instead of disabling linting on the offending line, I removed `flake8-import-order` based on the consensus in this thread. Ready for review.

dmfigol:
Can we merge this?

PR #467, opened 2020-01-14
Replace multiprocessing.dummy with concurrent.futures

dmfigol:
This pull request replaces `multiprocessing.dummy.Pool` with `concurrent.futures.ThreadPoolExecutor`.
Reasons for change:
- import of `multiprocessing` is misleading, because we are using threads, not multiple processes. We already had a situation when people looking at the source code were surprised
- `multiprocessing.dummy` [docs](https://docs.python.org/3.6/library/multiprocessing.html#module-multiprocessing.dummy) are lacking compared to `concurrent.futures` [ThreadPoolExecutor](https://docs.python.org/3.6/library/concurrent.futures.html#threadpoolexecutor) and [ProcessPoolExecutor](https://docs.python.org/3.6/library/concurrent.futures.html#processpoolexecutor)

I've also done performance testing. I was using this task on 500 hosts with 125 workers:
```
def useless_task(task):
    time.sleep(15)
    return Result(host=task.host, result=task.host.name)
```
Results:
- `concurrent.futures`: 1:00.54 / 1:00.55 / 1:00.54
- `multiprocessing.dummy`: 1:00.71 / 1:00.71 / 1:00.72

I've also tested this implementation with gathering commands using netmiko on 500 routers with ~90 workers, the difference in results was negligible.

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -1,13 +1,16 @@
 import logging
 import logging.config
-from multiprocessing.dummy import Pool
-from typing import List, Optional
+from concurrent.futures import ThreadPoolExecutor
+from typing import List, Optional, TYPE_CHECKING, Dict, Any
 
 from nornir.core.configuration import Config
 from nornir.core.inventory import Inventory
 from nornir.core.processor import Processor, Processors
 from nornir.core.state import GlobalState
 from nornir.core.task import AggregatedResult, Task
+
+if TYPE_CHECKING:
+    from nornir.core.inventory import Host  # noqa: W0611
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ktbyers:
I am generally in favor of this (i.e. unless there is some meaningful negative consequences to it).

This is mainly from the perspective that `concurrent.futures` is the standard way of doing concurrency in Python3 (so it is generally better to use the more common pattern unless there is a good reason not to).

Also I find the `concurrent.futures` interface easier to use and reason about, and there are some nice features there that are not available in the legacy solutions (or at least that would be much harder to implement).

Finally, using the `multiprocessing.dummy` frequently confuses people as to whether we are using threading or not (so this would be clearer in that regard).

dbarrosop:
Thanks for the work, I don't have strong opinions one way or the other, however, before merging anything I'd like to have a benchmarking testsuite:

1. An inventory with 500 dummy devices
2.a. A task like the one in your example.
2.b. A task similar to the one you created but also manipulating some shared data
3. A test profiling CPU/RAM and measuring how long it took.
4. Then run this for 500 devices with different combinations of `num_workers`.

Assuming similar results this looks good to me but I think I'd like to wait until we can reliably measure things.

dbarrosop:
Using the profiler in #471 shows similar results in all the measured metrics so I am good with this change.

PR #469, opened 2020-01-14
Add skip mark to netconf tests

dmfigol:
Netconf tests require docker environment up and running, so outside of that they should be marked with `@skip` marker not to fail the test suite without docker.


dbarrosop:
Thanks!

PR #474, opened 2020-01-26
pinned pydantic

dbarrosop:

PR #472, opened 2020-01-21
napalm_ping.py

tejasmokashi1992:
Add napalm_ping plugin to nornir

PR #475, opened 2020-01-26
prepare 2.4.0 release

dbarrosop:
After #474 and hopefully after #472 

Note: Update changelog after those have been merged


fach:
Would it be possible to sneak in the fix from #481 into 2.4.0? Thanks!

dbarrosop:
Yes, whatever is in develop will make it into 2.4.0 so that should be included. I will need to update the CHANGELOG last minute though to reflect it

PR #477, opened 2020-01-28
Update __init__.py

tejasmokashi1992:
update __init__.py with napalm_ping.py(#472 ) plugin under section tasks/networking


tejasmokashi1992:
#472

PR #479, opened 2020-02-02
Maps all available Netbox device attributes to Host's data dictionary

wvandeun:
With the previous state of the Netbox plugin only certain attributes of a Netbox device were mapped to a key in the data dictionary of a Host. This PR changes that behavior and saves any data returned by the devices API endpoint into the data dictionary of the Hosts object.
This includes the rendered config_context, for which there is a separate PR #456 and other attributes that some people have asked for.
Basically all the attributes that are returned by the `api/dcim/devices` API endpoint are available in the Hosts data dictionary.

I have removed the `use_slugs` argument. It doesn't make a lot of sense to keep this argument with these changes since all the slugs are available in the data dictionary.


dbarrosop:
Nice job, I like PRs that mostly involve red lines :P

This is, unfortunately, a breaking change so we can't incorporate it as is as per semver. I am good with your proposal but it should be gated via an optional argument and the default behavior should be the same as now. Either that or a new netbox inventory plugin flagging the old one as deprecated. I suspect the latter might be preferable, what do you think?

wvandeun:
My preference would be a new plugin or create a new class in the existing plugin, flagging the old one as deprecated.

In any case, we'll have to come up with a proper name for the "new" plugin: my creativity pushes me to NewNetboxInventory or FullNetboxInventory? Any opinions?

dbarrosop:
Let's just call it `NetboxInventoryv2`.

Thanks!

dmfigol:
I will play devil's advocate here. are we sure we want to have another netbox plugin? Users could be very confused when they see two different ones.
I'd prefer an optional argument.

dbarrosop:
I think so, it's going to keep things cleaner and the old one will raise a DeprecationWarning asking users to switch. So the old one won't be maintained and will clearly flagged so it's not used.

dmfigol:
@dbarrosop ok, I concede :) 
1) do we want deprecation warning in the next version or in the next major release?
2) probably it makes sense to have them in separate files

wvandeun:
Created a new class NetboxInventory2 with the new functionality in the same file.
The old class is still available with the same default behavior, but it raises a warning message for deprecation.

clay584:
@wvandeun I made a PR to your fork. If you merge, and then update this PR, some additional features will be in this PR.

wvandeun:
Implemented in NetBoxInventory2 for Nornir 3.0
https://github.com/wvandeun/nornir_netbox

Issue #480, opened 2020-02-03
Should we drop pydantic from our dependencies?


dmfigol:
Related to #473 
Also, to be fair, we were using a private method, so it was expected to break.
Despite this, I still think we should at least consider removing it from our dependencies because:
1. This is the only library we had numerous problems with over the last 1.5 years
2. Pydantic can be used as user-facing library and conflicts between versions we are using and users want to use are imminent. For example, something like [fastAPI](https://github.com/tiangolo/fastapi) web framework uses pydantic too (albeit loosely constrained)

Options I could come up with:
1. Leave it as it is
2. Vendoring - this probably is the easiest option. Take something like pydantic 1.3, vendor and forget about it. It will not collide with user code anymore
3. Something like [dynaconf](https://dynaconf.readthedocs.io) - library to deal with config files. I use it in some of my projects where I have too many input variables/secrets. It does have validators and env prefix.
4. Something like [marshmallow](https://marshmallow.readthedocs.io/en/stable/). Unlike previous project, this is a library specifically for serialization/deserealization, albeit very mature. I don't think it has env var support, since it was not designed for config files
5. Our own implementation

Thoughts?

dbarrosop:
I think short term we should vendor it and longer term replace it with our own.

dmfigol:
I agree.
Pydantic was vendored in #482, so let's close this.

PR #482, opened 2020-02-04
vendor pydantic

dbarrosop:
Relates to #480


dbarrosop:
Done, not sure I like it though. Not sure what the purpose of the leading `_` is and, to make matters worse, not the vendor folder is shown first in the PR

dmfigol:
> Done, not sure I like it though. Not sure what the purpose of the leading `_` is and, to make matters worse, not the vendor folder is shown first in the PR

to signal to users that they shouldn't import it / rely on it

pyproject.toml
```diff
@@ -39,6 +38,9 @@ sphinx-issues = { version = "^1.2", optional = true }
 typing_extensions = "^3.7"
 ncclient = "^0.6.6"
 
+# for pydantic
+dataclasses = {version = "^0.7", python = "3.6"}
```
dmfigol: I think it should be `python = "~3.6"` according to
https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies

dbarrosop: fixed

PR #485, opened 2020-02-11
Capirca task plugin

fach:
This PR adds a Nornir task plugin for utilizing [Capirca](https://github.com/google/capirca) to generate per-host ACLs. Upon success, the rendered ACL is returned as a string in the result field of a `Result`. For an ACL to be rendered, the plugin must be passed a valid capirca platform type and the host platform must be a target in the specified policy file:
```
header {
  comment:: "Sample multitarget loopback filter"
  target:: juniper LOOPBACK
  target:: arista LOOPBACK
  target:: ciscoxr LOOPBACK
}
```
This platform value is different than the top-level platform inventory field, as the existing inventory values (i.e. junos) didn't map 1:1 to capirca platforms, and maintaining a translation map seemed implicitly terrible. Therefore, the capirca platform type must be explicitly passed but can be included as inventory data as below and passed at task runtime:
```
dev4.group_2:
    data:
        capirca:
            platform: juniper
            expiration: 2
            shade_check: true
            optimize: true
```

The application dependency of Capirca v1.122/latest was left off this PR per the contributing guidelines. Please let me know if you want it added here, in a separate PR or if you will handle it. 

Also, if there is value, I'm happy to add this as a Nornir function as well, which would most likely render and write ACLs to a specified directory, much like the current Capirca toolchain.

### DIFF
--- a/nornir/plugins/tasks/text/__init__.py
+++ b/nornir/plugins/tasks/text/__init__.py
@@ -1,4 +1,5 @@
+from .capirca_acl import capirca_acl
 from .template_file import template_file
 from .template_string import template_string
 
-__all__ = ("template_file", "template_string")
+__all__ = ("capirca_acl", "template_file", "template_string")
--- a/nornir/plugins/tasks/text/capirca_acl.py
+++ b/nornir/plugins/tasks/text/capirca_acl.py
@@ -1,0 +1,153 @@
+from capirca.lib import arista
+from capirca.lib import aruba
+from capirca.lib import brocade
+from capirca.lib import cisco
+from capirca.lib import ciscoasa
+from capirca.lib import ciscoxr
+from capirca.lib import ipset
+from capirca.lib import iptables
+from capirca.lib import juniper
+from capirca.lib import junipersrx
+from capirca.lib import naming
+from capirca.lib import nftables
+from capirca.lib import paloaltofw
+from capirca.lib import policy
+from capirca.lib import srxlo
+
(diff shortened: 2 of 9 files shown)


fach:
Checks are failing due to the missing Capirca dependency mentioned above.

ktbyers:
I think I would add the `caprica` dependencies into the `dev` dependencies in pyproject.toml and update the poetry.lock file.

https://github.com/nornir-automation/nornir/blob/develop/pyproject.toml#L44

In other words, make it so the tests/CI-CD can work, but so that end-users are not required to install this dependency.

Others can chime in if they don't like this pattern.

We possibly need a more modular way to manage dependencies for plugins.

fach:
Alright, I did the thing. Travis is happy now.

dmfigol:
@fach thank you for the contribution, Steve! We are right now discussing where the plugins should live #486 and it is very likely that we will end up separating them from the core. Hence at this point, I think it makes sense to hold onto this PR.

fach:
Sounds good. Happy to be a trailblazer with this PR and help get boiler plate going if we want to move plugins outside of the core repo.

dbarrosop:
If you haven't look at #492 and #486 and feel free to provide feedback on the new nornir_napalm repo which I am using to experiment how a plugin repo could look like

ktbyers:
I am going to close this PR as it is no longer applicable to nornir-core (i.e. it should be in its own separate nornir plugin repository).

PR #491, opened 2020-02-15
migrate to github actions

dbarrosop:
Adios travis, hola github actions.

In addition I removed the dependency of running the tests inside containers. I am not a fan of that change but it simplifies the migration. I might look into reverting this in the future but for now this is cleaner and easier.


ktbyers:
This is fine with me.

dbarrosop:
> currently the build is triggered twice on push to a branch involved in PR

Yes, that's by design and that's correct. The jobs may contain different code if your branch is missing some commit from base.

Issue #496, opened 2020-03-10
NetBox add support for tags


egreenspan2:
Would like to propose pulling in tags from NetBox into nornir inventory, should be a simple add to the NB inventory plugin

nornir/plugins/inventory/netbox.py

`host["data"]["tags"] = d["tags"]`

This can make it much easier to deal with certain devices based on the tags from Netbox.

wvandeun:
This PR can be closed. Implemented in NetBoxInventory2 for Nornir3.0.

dbarrosop:
thanks for the heads up

PR #501, opened 2020-03-24
Add netbox inventory plugin tag support #496

egreenspan2:
Get NetBox tags array into Nornir, updated test devices with tags to validate, per #496  (novice developer, be gentle)

### DIFF
--- a/nornir/plugins/inventory/netbox.py
+++ b/nornir/plugins/inventory/netbox.py
@@ -69,6 +69,7 @@ class NBInventory(Inventory):
             host["data"]["serial"] = d["serial"]
             host["data"]["vendor"] = d["device_type"]["manufacturer"]["name"]
             host["data"]["asset_tag"] = d["asset_tag"]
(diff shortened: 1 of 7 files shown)


ktbyers:
@egreenspan2 We are moving towards an external plugin model for plugins like the Netbox plugin. I believe a set of other people were already doing a bunch of work on an updated Netbox plugin. Let me see if I can connect you to them.

ktbyers:
@clay584 you were working on the new Netbox plugin correct?

clay584:
Yes. Myself and @wvandeun

wvandeun:
#479 contains all of the work for tags (and other attributes already).
It will map all of the applied tags for a device into the Nornir host data attribute.

egreenspan2:
> 
> 
> #479 contains all of the work for tags (and other attributes already).
> It will map all of the applied tags for a device into the Nornir host data attribute.

Makes more sense now, will start testing with new plugin.  I'll close this out.

PR #505, opened 2020-03-28
use entrypoints to discover connection plugins

dbarrosop:
This PR uses `entry_points` to discover connection plugins. Currently, an external plugin has to run the following custom code on import to register itself:

```python
from nornir.core.connections import ConnectionPlugin, Connections


CONNECTION_NAME = "napalm"


class Napalm(ConnectionPlugin):
    ...


# Register with nornir
if CONNECTION_NAME not in Connections.available:
    Connections.register(CONNECTION_NAME, Napalm)
```

With this change, the last bits are not needed and an entry_point is used instead:

``` toml
 [tool.poetry.plugins."nornir.plugins.connections"]
"netmiko" = "nornir.plugins.connections.netmiko:Netmiko"
```

We may want to extend this to the inventory plugins so we can avoid requiring the long import path in the configuration. Not sure if there is any benefit to using this for tasks, at best we could present all the tasks under a common namespace but there is no clear benefit for this, consistency/simplicity maybe?

### DIFF
--- a/nornir/init_nornir.py
+++ b/nornir/init_nornir.py
@@ -1,19 +1,20 @@
+import pkg_resources
 import warnings
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, Type
 
 from nornir.core import Nornir
-from nornir.core.connections import Connections
+from nornir.core.connections import Connections, ConnectionPlugin
 from nornir.core.deserializer.configuration import Config
 from nornir.core.state import GlobalState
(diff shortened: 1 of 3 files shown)


dbarrosop:
Pinging @ktbyers @ogenstad and, specially, @dmfigol as this is something we had discussed in the past.

dmfigol:
Despite my lack of Python entrypoints knowledge, it looks great. Learned something new!
Thanks for the change, David!

PR #506, opened 2020-03-29
napalm_traceroute.py

tejasmokashi1992:
Pulgin for napalm_traceroute


ktbyers:
I think just using a grouped task ("custom" task) is sufficient for this i.e. we probably don't need a separate plugin for this (at least in nornir or in nornir_napalm repositories).

Process for grouped task is here (similar to this):

https://github.com/twin-bridges/nornir_course/blob/master/bonus1/collateral/napalm_direct_task/napalm_direct_jnpr.py

tejasmokashi1992:
ohh okk....
No problem will close this PR

ktbyers:
@tejasmokashi1992 Yes, thanks for the contribution...I am just a bit worried about a proliferation of plugins that would have to be maintained.

Regards, Kirk

Issue #508, opened 2020-04-09
improve docs


dbarrosop:
- [ ] adapt docs to nornir 3.0
- [ ] some generic notes about writing plugins; how to pin nornir as per semver and how to add the plugin to the plugin repository
- [ ] document how to write inventory plugins
- [ ] document how to write task plugins
- [ ] document how to write connection plugins #526
- [ ] document how to write processors
- [ ] document how to manage and pin dependencies with poetry and pip

PR #526, opened 2020-04-11
Add howto create a new connection plugin

ubaumann:
Add a howto create new connection plugin with a demo project

#508


ubaumann:
I never used entry points so I was thinking creating the documentation would be a good start for understanding.

I'm not sure if the jupyter notebook the correct choice. I'm open for any feedback. Hopefully I understood the idea correct.

dbarrosop:
Awesome! I think a notebook is the way to go, even if some parts are just written in markdown because they are hard or not very feasible to do in code. I still need to look a bit more in depth to your documentation but something it is worth noting is that you can register your plugin "manually" like this:

```
from nornir.core.connections import Connections

class MyConnectionPlugin:
    ...

Connections.register("my_connection_plugin", MyConnectionPlugin)
```

This method might be more convenient for people using monorepos and don't want/can't package their code properly.

Also, in addition to check with setuptools you can do the following to verify it was registered correctly:

```
Connections.get_plugin("my_connection_plugin")
```

ubaumann:
You are right. The manual registration could be handy for developing and custom implementations. I may focused too much on the entry point registration because it was completely new for me.  Luckily it just the first draft ;)

I will add it and make an example.

It's also the first time I took the time to look finally at poetry (It is awesome) and added it to the how to. Having poetry inside a poetry project is a little messed up so I will move it into markdown cells.

I was now thinking if it would make sense to move the plugin how to a separate menu point.
I'm thinking about a structure like this:
```
doc $ tree -n plugins/
plugins/
├── connection_plugin/
├── connection_plugin.ipynb
├── index.rst
├── inventory_plugin/
├── inventory_plugin.ipynb
├── process_plugin/
├── process_plugin.ipynb
├── setup_project_with_poetry.rst
├── task_plugin/
└── task_plugin.ipynb
```

As I saw you are planning to add the registration for inventory plugins too. Any plans for the other plugins as well?

dbarrosop:
awesome work! :)

PR #528, opened 2020-04-14
Improved whitespace handling in jinja template rendering

indy-independence:
When developing jinja templates for use with Nornir I've run in to two issues that I think we can solve by adjusting the jinja environment for template rendering.

1. If the template includes several blocks that are indented, the output also gets a strange indent. Ex:
```
{% for loop1 %}
  {% if statement1 %}
    {% if statement2 %}
 description {{ description }}
```
All those whitespace before the {%%} blocks seems to be compounded and added before the description line, which makes it's indent look really weird in the output. If we add lstrip_blocks=True those whitespace to the left will be removed.

2. When including several files the removal of trailing whitespace might cause issues. Ex:
```
{% include 'vlan.j2' %}
{% include 'vrf.j2' %}
```
The trailing newline from the file vlan.j2 will be removed causing the last line of vlan.j2 and the first line of vrf.j2 to end up on the same line. This might not be an issue when rendering HTML templates or something but for CLI config it's annoying which is why I suggest we change the defaults.

I don't see any big disadvantage in changing these defaults for the type of config rendering that's done in Nornir, but if there are maybe this should be configurable somehow instead.

### DIFF
--- a/nornir/core/helpers/jinja_helper.py
+++ b/nornir/core/helpers/jinja_helper.py
@@ -11,7 +11,11 @@ def render_from_file(
 ) -> str:
     jinja_filters = jinja_filters or {}
     env = Environment(
-        loader=FileSystemLoader(path), undefined=StrictUndefined, trim_blocks=True
+        loader=FileSystemLoader(path),
+        undefined=StrictUndefined,
+        trim_blocks=True,
+        lstrip_blocks=True,
+        keep_trailing_newline=True,
     )
     env.filters.update(jinja_filters)
     t = env.get_template(template)
--- a/nornir/core/helpers/jinja_helper.py
+++ b/nornir/core/helpers/jinja_helper.py
@@ -7,31 +7,34 @@ def render_from_file(
     path: str,
     template: str,
     jinja_filters: Optional[Dict[str, Any]] = None,
+    jinja_env: Optional[Environment] = None,
     **kwargs: Any
 ) -> str:
     jinja_filters = jinja_filters or {}
-    env = Environment(
-        loader=FileSystemLoader(path),
-        undefined=StrictUndefined,
-        trim_blocks=True,
-        lstrip_blocks=True,
-        keep_trailing_newline=True,
-    )
+    if jinja_env:
+        env = jinja_env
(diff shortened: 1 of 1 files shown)


ktbyers:
I sort of can go either way here on lstrip_blocks=True...you can also pretty easily solve using the standard Jinja2 `{%-` indicator i.e.:

```
{%- for loop1 %}
  {%- if statement1 %}
    {%- if statement2 %}
 description {{ description }}
```

If we do this PR or something similar, we should probably add all three of these as arguments to the function definition (so users could control them if they wanted/needed to).

```
        trim_blocks=True,
        lstrip_blocks=True,
        keep_trailing_newline=True,
```

The argument against doing this would be in certain situations, you might require the Jinja2 get rendered exactly as what you put in (i.e. whitespace doesn't get automatically stripped). But as long as they were function arguments and the user could control them, that would be fine.

dmfigol:
Iam against changing current defaults, because it will break backwards compatibility (e.g. may be you are generating a config file where every space matters and suddenly defaults are changed in the next nornir release) but extra arguments to the task would be a good addition.

ktbyers:
Sounds good to me...i.e. add arguments to support this behavior, but default to being backwards compatible.

dbarrosop:
I actually think it should, instead, accept an environment. That way you get full flexibility to do whatever you want (change the loader, options, even add your own filters). So something like:

```
def render_from_file(
    path: str,
    template: str,
    jinja_filters: Optional[Dict[str, Any]] = None,
    env: Optional[Environemnt] = None,
    **kwargs: Any
) -> str:
   if env is None:
       # we configure the env as we do today
   ...
```

ktbyers:
Yeah, sounds good to me. I think David's solution of passing in the Jinja2 environment object is a better solution.

indy-independence:
Okay yes that seems like a good solution, it will expose some of the underlaying jinja library to the user of Nornir but I think it's better than trying to build something of our own to map these settings. I commited some new changes to the PR now, does it look better?

dbarrosop:
excellent! and sorry this took so long :(

PR #538, opened 2020-05-10
pass ssh_config only if file exists

### LINKED ISSUE #537: Nornir 2.4.0 requires SSH config file for NETCONF connections
Creating simple "shim" tasks for NETCONF fails in Nornir 2.4.0 but works in Nornir 2.3.0. This seems more like a lower level Nornir connection/option handling problem than a NETCONF problem.

Nornir runbook:
```
(pspynet) [centos@devbox nornir_nc_issue]$ cat get_runbook.py
#!/usr/bin/env python

from nornir import InitNornir
from nornir.core.task import Result

def netconf_get_config(task, source="running", **kwargs):
    conn = task.host.get_connection("netconf", task.nornir.config)
    result = conn.get_config(source=source, **kwargs)
    return Result(host=task.host, result=result)

def print_config(task):
    vrf_filter = ("subtree", "<native><vrf></vrf></native>")
    try:
        mresult = task.run(task=netconf_get_config, filter=vrf_filter)
        print(mresult[0].result)
    except Exception as exc:
        print(exc.result.result)

nornir = InitNornir()
nornir.run(task=print_config)
```

Inventory file:
```
(pspynet) [centos@devbox nornir_nc_issue]$ cat hosts.yaml
---
devnet_sbx_iosxe:
  hostname: ios-xe-mgmt-latest.cisco.com
  username: developer
  password: C1sco12345
  platform: ios
  port: 10000  # sandbox uses 10000 for NETCONF
  connection_options:
    netconf:
      extras:
        allow_agent: false
        hostkey_verify: false
        look_for_keys: false
```

Everything works on Nornir 2.3.0 without issue.
```
(pspynet) [centos@devbox nornir_nc_issue]$ pip list | grep nornir
nornir                       2.3.0
(pspynet) [centos@devbox nornir_nc_issue]$ python get_runbook.py
<?xml version="1.0" encoding="UTF-8"?>
<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:563ec7d8-f01a-45b1-b2ca-966d9a5755a1" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0"><data><native xmlns="http://cisco.com/ns/yang/Cisco-IOS-XE-native"><vrf><definition><name>CUST1</name><rd>1:1</rd><route-target><export><asn-ip>1:1</asn-ip></export><import><asn-ip>1:1</asn-ip></import></route-target></definition></vrf></native></data></rpc-reply>
```

Nornir 2.4.0 appears to require an SSH config file and fails.
```
(pspynet) [centos@devbox nornir_nc_issue]$ pip list | grep nornir
nornir                       2.4.0
(pspynet) [centos@devbox nornir_nc_issue]$ python get_runbook.py
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/nornir/core/task.py", line 85, in start
  File "get_runbook.py", line 7, in netconf_get_config
    conn = task.host.get_connection("netconf", task.nornir.config)
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/nornir/core/inventory.py", line 302, in get_connection
    extras=conn.extras,
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/nornir/core/inventory.py", line 352, in open_connection
    configuration=configuration,
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/nornir/plugins/connections/netconf.py", line 118, in open
    connection = manager.connect_ssh(**parameters)
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/ncclient/manager.py", line 143, in connect_ssh
    session.connect(*args, **kwds)
  File "/home/centos/environments/pspynet/lib/python3.7/site-packages/ncclient/transport/ssh.py", line 355, in connect
    with open(os.path.expanduser(ssh_config)) as ssh_config_file_obj:
FileNotFoundError: [Errno 2] No such file or directory: '/home/centos/.ssh/config'
```

To make this issue easy to replicate, you can use the Cisco DevNet sandbox for quick tests using the script and inventory file provided in this issue. If this behavior is expected (ie, that Nornir will now be requiring these SSH config files), I would appreciate a link to the release notes or documentation explaining it. I wasn't able to find anything after searching.

dbarrosop:
fixes #537

### DIFF
--- a/nornir/plugins/connections/netconf.py
+++ b/nornir/plugins/connections/netconf.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Any, Dict, Optional
 
 from ncclient import manager
@@ -108,9 +109,10 @@ class Netconf(ConnectionPlugin):
 
         if "ssh_config" not in extras:
             try:
-                parameters["ssh_config"] = configuration.ssh.config_file  # type: ignore
+                ssh_config_file = Path(configuration.ssh.config_file)  # type: ignore
+                if ssh_config_file.exists():
+                    parameters["ssh_config"] = ssh_config_file
             except AttributeError:
-                parameters["ssh_config"] = None
                 pass
 
         parameters.update(extras)
--- a/tests/inventory_data/netconf_hosts.yaml
+++ b/tests/inventory_data/netconf_hosts.yaml
@@ -11,3 +11,9 @@ netconf1.no_group:
                 hostkey_verify: False
                 look_for_keys: False
                 ssh_config: null
+
+netconf2.no_group:
+    hostname: localhost
+    username: netconf
+    password: netconf
+    port: 65025
--- a/tests/plugins/tasks/networking/test_netconf_capabilities.py
+++ b/tests/plugins/tasks/networking/test_netconf_capabilities.py
@@ -2,6 +2,9 @@ from nornir.plugins.tasks import networking
 
 
 def test_netconf_capabilities(netconf):
+    netconf = netconf.filter(name="netconf1.no_group")
+    assert netconf.inventory.hosts
+
     result = netconf.run(networking.netconf_capabilities)
 
     for _, v in result.items():
--- a/tests/plugins/tasks/networking/test_netconf_connection.py
+++ b/tests/plugins/tasks/networking/test_netconf_connection.py
@@ -1,0 +1,25 @@
+from nornir.core.task import Result, Task
+from nornir.plugins.tasks import networking
+
+
+def get_task(task: Task, path: str = "", filter_type: str = "xpath") -> Result:
+    manager = task.host.get_connection("netconf", task.nornir.config)
+    params = {}
+    if path:
+        params["filter"] = (filter_type, path)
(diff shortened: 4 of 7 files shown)


nornir/plugins/connections/netconf.py
```diff
@@ -108,9 +109,10 @@ def open(
 
         if "ssh_config" not in extras:
             try:
-                parameters["ssh_config"] = configuration.ssh.config_file  # type: ignore
+                ssh_config_file = Path(configuration.ssh.config_file)  # type: ignore
+                if ssh_config_file.exists():
+                    parameters["ssh_config"] = ssh_config_file
```
ogenstad, line 114: I think this can perhaps become a bit confusing if a user has made a typo in the name of the config file, that way the supplied file wouldn't exist and the user wouldn't be notified about that problem. But I'm also fine with this small fix for Nornir 2.0 and then I can change it for 3.0.

PR #539, opened 2020-05-13
Extended NB plugin to extract virtual machines

ashishkurian:
I have extended the NB plugin to extract the virtual machines also from the NB inventory.


dbarrosop:
Bear in mind that the CI is broken, you will need to fix it.

/cc @wvandeun and @clay584

ashishkurian:
@dbarrosop : should I need to fix all the 4 test or I should only concentrate on the Python3.8 failure?

dbarrosop:
All tests should pass but fixing one environment will most likely fix the rest (unless you used some fancy new feature not available in previous versions of python)

ashishkurian:
@dbarrosop Can you help me identify what is going wrong? I see that the error return the virtual machine api url of NB as 
`requests_mock.exceptions.NoMockAddress: No mock address: GET http://localhost:8080/api/virtualization/virtual-machines/?limit=0
`
I also added a new type of mock device and also edited the expected and expected transformation function.

ktbyers:
I am going to close this PR as it looks to be more of PR related to NetBox Plugin (and realistically looks unrecoverable at this point i.e. that the PR would need to be redone to handle all of the conflicts).

PR #541, opened 2020-05-13
regression: pass nornir object to task

### LINKED ISSUE #540: Nornir 3.0 - runners break task plugins on task.nornir.config
This is the current pattern being used on the new nornir_napalm task plugins:

device = task.host.get_connection(CONNECTION_NAME, task.nornir.config)

But `task.nornir` was removed as part of the runner changes

https://github.com/nornir-automation/nornir_napalm/blob/master/nornir_napalm/plugins/tasks/napalm_configure.py#L29

dbarrosop:
fixes #540

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -107,6 +107,7 @@ class Nornir(object):
         """
         task = Task(
             task,
+            self,
             global_dry_run=self.data.dry_run,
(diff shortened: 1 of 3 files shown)


ktbyers:
Testing this now...as I was looking at this a bunch today.

PR #543, opened 2020-05-14
bring back SimpleInventory to nornir3.0.0

dbarrosop:
After discussing it with some of you I have decided to bring back `SimpleInventory` to nornir with a couple of caveats:

1. It is going to be as simple as the name indicates. Features here won't be accepted. The goal of this plugin will be to lower the barrier to entry to nornir.
2. `nornir_utils` (or some other repo) would an alternative version with extra bells and whistles (if needed)

### DIFF
--- a/tests/plugins/inventory/data/netconf_hosts.yaml
+++ b/tests/plugins/inventory/data/netconf_hosts.yaml
@@ -1,0 +1,13 @@
+---
+netconf1.no_group:
+    hostname: localhost
+    username: netconf
+    password: netconf
+    port: 65025
+    connection_options:
+        netconf:
+            extras:
+                allow_agent: False
+                hostkey_verify: False
+                look_for_keys: False
+                ssh_config: null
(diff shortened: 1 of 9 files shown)


tests/plugins/inventory/data/netconf_hosts.yaml
```diff
@@ -0,0 +1,13 @@
...
+    username: netconf
+    password: netconf
+    port: 65025
+    connection_options:
+        netconf:
+            extras:
+                allow_agent: False
+                hostkey_verify: False
+                look_for_keys: False
+                ssh_config: null
```
ogenstad, line 13: This file should be safe to remove from the PR.

dbarrosop:
> SimpleInventory in nornir_utils is renamed to something else

+100 to that. It will probably be renamed to `YAMLInventory` or something of the like. PR will there after this one is merged and released

PR #548, opened 2020-05-18
Default to SimpleInventory if no inventory plugin in specified (nornir3.0)

ktbyers:
This causes a blank InitNornir() call to retain its current behavior (i.e. to default to using SimpleInventory).


dbarrosop:
Thanks for noticing, this should be changed here though:

https://github.com/nornir-automation/nornir/blob/3.0.0/nornir/core/configuration.py#L76

ktbyers:
PR updated.

PR #558, opened 2020-06-09
use the plugin register for connection plugins

dbarrosop:
This reuses the plugin register for connections too

Will require a minor change in the connection plugins:

https://github.com/nornir-automation/nornir_napalm/pull/7


ktbyers:
get_connection is no longer returning the NAPALM/Netmiko object:

```
     32     import ipdb; ipdb.set_trace()
     33     device = task.host.get_connection(CONNECTION_NAME, task.nornir.config)
---> 34     getters_options = getters_options or {}
```

This is from napalm_get.py, but I observed the same thing in nornir_netmiko (after I made the same change removing the ConnectionPlugin inheritance).

Previously it used to return:

```
ipdb> p self.connections['napalm'].connection                                                                    
<napalm.ios.ios.IOSDriver object at 0x7ff950fa4400>
```

Now it returns:

```
ipdb> p self.connections['napalm']                                                                               
<nornir_napalm.plugins.connections.Napalm object at 0x7ff950fa43c8>
```

This  breaks the NAPALM/Netmiko task-plugins as they will use code like this to retrieve the NAPALM/Netmiko connection object:

https://github.com/ktbyers/nornir_netmiko/blob/master/nornir_netmiko/tasks/netmiko_send_command.py#L26

nornir/core/inventory.py
```diff
@@ -405,16 +405,7 @@ def get_connection(self, connection: str, configuration: Config) -> Any:
...
-
-    def get_connection_state(self, connection: str) -> Dict[str, Any]:
-        """
-        For an already established connection return its state.
-        """
-        if connection not in self.connections:
-            raise ConnectionNotOpen(connection)
-
-        return self.connections[connection].state
+        return self.connections[connection]
```
ktbyers, line 408: This is the line that changed (pertaining to returning the NAPALM/Netmiko object (i.e. used to be `self.connections[connection].connection`)

ktbyers:
+1

I tested this against a bunch of my nornir course exercises/reference examples and it all worked fine. I also made a corresponding change in nornir_netmiko and that also worked properly.

PR #563, opened 2020-06-18
Nornir 3.x: Use textfsm with fix for Windows

ktbyers:
I recognize this is development only dependency, but since I am very frequently working on Netmiko (which requires TextFSM) and NAPALM, reverting back to the downrev textfsm 0.4.1 is going to cause me a lot of pain in dependency management.

So I would rather roll forward and include the windows patch here.

That branch I included won't change.


ktbyers:
This PR also fixes the GitHub-actions  breakage.

PR #567, opened 2020-07-06
[For Discussion] Persistent HTTP Connection Plugin

brasswood:
I wrote a very simple persistent HTTP connection plugin that can store cookies and take advantage HTTP keep-alive. It is just a wrapper around a `requests.Session`, with the option to verify certificates or not when using HTTPS.

I think there needs to be some facility for people to handle device-specific login and logout sequences. For example, before my session is `close()`'d, I would like to properly send the 'DELETE' request to log out of my device.

One way to do this is the Ansible way, allowing users to implement callbacks like `login()` and `logout()` in "httpapi" plugins. Then the "connection" plugin calls these callbacks (this confused me for a long time).

Another way could be to create a separate library to handle all of the differences in devices, like netmiko but for HTTP.

Or maybe people could just develop regular nornir connection plugins for their device's API.

What do you think?


ktbyers:
Nornir in version 3.0 is moving to a plugin moduel where plugins are not stored in the core Nornir repository, but are just stored in repositories maintained by individual contributors.

Here is an example plugin that I created for Netmiko:

https://github.com/ktbyers/nornir_netmiko

So this PR probably wouldn't go here, but probably would just be maintained by you in your repository.

brasswood:
@ktbyers ok, how do we write and install plugins?

ktbyers:
Write is copy the pattern employed in something like `nornir-napalm`, `nornir-utils`, or `nornir-netmiko`. Key thing to be aware of is the entry-point which is defined in either pyproject.toml or in setup.py.

Install of plugins would typically be done by `pip` so relevant plugin would be uploaded to pypi and would be pip installed.


Regards, Kirk

dbarrosop:
I think that writing connection plugins is simple enough that implementing callbacks and complex configurations to generic ones brings complexity without solving any real issues but that's just my subjective opinion :)

> I would like to properly send the 'DELETE' request to log out of my device.

I am curious to know which device has an HTTP request that requires that step to log a user out. Certainly not very RESTful :P (not that I think REST is a good model for network operations)

> Write is copy the pattern employed in something like nornir-napalm, nornir-utils, or nornir-netmiko

Just want to clarify that those patterns are optional, you can follow any other pattern if you prefer. The only manadatory bit is registering the connection plugin:

https://nornir.readthedocs.io/en/3.0.0/plugins/index.html

Also, note that this applies to nornir 3 which is in beta at this point but will be released "soon" (when time and summer permits :D)

> Install of plugins would typically be done by pip

Yeah, expanding on that, `pip` is the easiest but:

1. Not required, you can use any other method if you prefer as long as that method allows you to register the plugins and to import them
2. You don't need to upload to pypi to be able to install them, pip supports installing from local folders, remote git repos, etc...

brasswood:
> I think that writing connection plugins is simple enough that implementing callbacks and complex configurations to generic ones brings complexity without solving any real issues but that's just my subjective opinion :)

Agree with you there 100%. The framework should not be too much framework.

ktbyers:
Closing as this is an old plugin submitted against Nornir core so it is no longer applicable here.

PR #573, opened 2020-07-19
Added option to output formatted string instead of printing to stdout

shishir-joshi:
Co-authored-by: Tejas Mokashi <tmokashi1992@gmail.com> @tejasmokashi1992


nornir/plugins/functions/text/__init__.py
```diff
@@ -115,7 +161,8 @@ def print_result(
     vars: List[str] = None,
     failed: bool = False,
     severity_level: int = logging.INFO,
-) -> None:
+    return_output: bool = False,
```
dbarrosop, line 164: instead of using a bool for this and then the global `OUT_BUFFER` I'd suggest the user passes the `StringIO`. Something like:

```
dest: Optional[StringIO] = None
```

nornir/plugins/functions/text/__init__.py
```diff
@@ -41,7 +44,15 @@ def _print_individual_result(
     failed: bool,
     severity_level: int,
     task_group: bool = False,
-) -> None:
+    return_output: bool = False,
```
dbarrosop, line 47: same as below, instead of using a bool just pass here the destination where you want to write the `print`. Difference here is that the parameter should be mandatory.

nornir/plugins/functions/text/__init__.py
```diff
@@ -77,14 +89,25 @@ def _print_result(
     attrs: List[str] = None,
     failed: bool = False,
     severity_level: int = logging.INFO,
-) -> None:
+    return_output: bool = False,
```
dbarrosop, line 92: same as below, instead of using a bool just pass here the destination where you want to write the `print`. Difference here is that the parameter should be mandatory.

nornir/plugins/functions/text/__init__.py
```diff
@@ -115,7 +161,8 @@ def print_result(
     vars: List[str] = None,
     failed: bool = False,
     severity_level: int = logging.INFO,
-) -> None:
+    return_output: bool = False,
+) -> Union[None, str]:
```
dbarrosop, line 165: with my proposal of having the user pass the `StringIO` this doesn't need to change.

dbarrosop:
Nice work, just a couple of comments to avoid using a global variable and to have the user pass the destination instead of "forcing it" (user should be able to pass anything compatible with `print`).

In addition, it'd be nice to add a test for this. There are a few tests already where stdout is captured so it'd be nice to test passing a StringIO as well.

https://github.com/nornir-automation/nornir/blob/develop/tests/plugins/functions/text/test_print_result.py

Once we agree on this we will need to submit this PR as well to this other repo for nornir3 compatibility: https://github.com/nornir-automation/nornir_utils

dbarrosop:
This is excellent, the only thing we are missing now is a test where we pass a `StringIO` to make sure we don't have a regression in the future :)

ktbyers:
This looks like it got stranded and was never completed...so closing it.

PR #574, opened 2020-07-24
fixes issue with urlencoding for filenames/repository in gitlab

### LINKED ISSUE #454: gitlab plugin _get_repository problem

wvandeun:
Fixes an issue with filenames (paths) & namespaced repositories, that contain characters which should be urlencoded.
This PR is basically the same as #455, but it solves some concerns that didn't receive updates.

Fixes #454

### DIFF
--- a/docs/plugins/tasks/version_control/gitlab.ipynb
+++ b/docs/plugins/tasks/version_control/gitlab.ipynb
@@ -90,7 +90,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {
     "autoscroll": false,
     "ein.hycell": false,
@@ -136,7 +136,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 2,
    "metadata": {
     "autoscroll": false,
     "ein.hycell": false,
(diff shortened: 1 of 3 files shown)


k01ek:
I think, this commit can't fix the problem. In some cases search method will return list of repo. The main problem is using search method instead of using URL-encoded path of the project.

wvandeun:
You are right!
Removed the usage of the search API endpoint now.
Still need to work on updating the documentation.

dbarrosop:
feel free to @ me once this PR is ready and its counterpart has been merged in the other repo :)

Thanks!

wvandeun:
@dbarrosop 
Fighting a bit with black here...
For some reason it considers [this line](https://github.com/nornir-automation/nornir/blob/8f63c796b6b0ea5d57c713490f974448dec5cd2a/tests/plugins/tasks/version_control/test_gitlab.py#L92) not properly formatted. Black would change it to a single line statement, which ends up with a line length which is too long.

wvandeun:
@dbarrosop this one should be good now
3.0 plugin in progress https://github.com/wvandeun/nornir_gitlab

PR #576, opened 2020-08-03
print host name when outside of AggregatedResult

brasswood:
This PR adds logic to print the host a result belongs to when that result is being printed from outside of an AggregatedResult.


dbarrosop:
Awesome, due to nornir3 release, would you mind opening this PR here too?

https://github.com/nornir-automation/nornir_utils

dbarrosop:
Merged here: https://github.com/nornir-automation/nornir_utils/pull/8

Thanks!

PR #590, opened 2020-09-22
Add ability to add groups to Inventory Element & Hosts at runtime.

### LINKED ISSUE #430: Assigning host to group during runtime
### Question

Is there any way to programmatically add a host to a particular group in order to retrieve the attribute of that group through the host later?

### Use Case

I am trying to make an automated ASA upgrade application. In order to do so, I need to assign specific models a specific upgrade file. My `groups.yaml` looks like this:

```
ASA5515:
  data:
    required_version: 9.8(4)8
    asa_file: asa984-8-smp-k8.bin
    asdm_file: asdm-openjre-7122.bin
    rest_api_file: asa-restapi-132346-lfbff-k8.SPA
```
The group name is the model string returned by a fact getter. That way I can determine programmatically which upgrade file a target device will use rather than hardcoding model numbers for each device. After I get the ASA model number, I then need to assign it to the appropriate group in order to pull the appropriate image files.

kpeterson-sf:
Update inventory to allow for the addition of groups to an Inventory Element at run time.

Added two methods to the InventoryElement: add() & remove()

Added two methods to Host: add_to_group() and remove_from_group()

Closes #430


nornir/core/inventory.py
```diff
@@ -105,6 +105,18 @@ def __contains__(self, value: object) -> bool:
         else:
             return any([value == g for g in self])
 
+    def add(self, group: "Group") -> None:
+        # only add the group if it doesn't exist
+        if not self.__contains__(group):
+            self.append(group)
+
+    def remove(self, group: "Group") -> None:
```
kpeterson-sf, line 113: this might be confusing.

nornir/core/inventory.py
```diff
@@ -105,6 +105,18 @@ def __contains__(self, value: object) -> bool:
...
+        if self.__contains__(group):
+            # remove overrides the parent class so calling super()
+            super(ParentGroups, self).remove(group)
```
kpeterson-sf, line 116: this might be confusing

dbarrosop: any reason to not use py3-style super; `super().remove(...)`?

nornir/core/inventory.py
```diff
@@ -105,6 +105,18 @@ def __contains__(self, value: object) -> bool:
...
+        else:
+            raise KeyError(group)
```
dbarrosop, line 118: Not sure I like this check, if you try to remove something that doesn’t exist I think it should error out. In addition, in the standard library removing an element from a list raises a ValueError if it doesn’t exist but you are raising a KeyError. I think we should be consistent with the stdlib, in which case this method wouldn’t be needed.

nornir/core/inventory.py
```diff
@@ -105,6 +105,18 @@ def __contains__(self, value: object) -> bool:
         else:
             return any([value == g for g in self])
 
+    def add(self, group: "Group") -> None:
```
dbarrosop, line 108: Would you mind adding a docstring indicating that `add` will not append the Group to the list if the object is already present?

nornir/core/inventory.py
```diff
@@ -150,6 +162,12 @@ def dict(self) -> Dict[str, Any]:
             **super().dict(),
         }
 
+    def add_to_group(self, group: "Group") -> None:
```
dbarrosop, line 165: I think this method is a bit unnecessary. In the original issue it was more interesting because that required also adding a reference but now that’s not needed so:

```
host.groups.add(group)
```

or

```
host.add_to_group(group)
```

do pretty much the same and the first is actually shorter to type :P

nornir/core/inventory.py
```diff
@@ -150,6 +162,12 @@ def dict(self) -> Dict[str, Any]:
...
+        self.groups.add(group)
+
+    def remove_from_group(self, group: "Group") -> None:
```
dbarrosop, line 168: same as comment about `add_to_group`

tests/core/test_inventory.py
```diff
@@ -530,3 +530,50 @@ def test_get_hosts_dict(self, inv):
         assert "group_1" in dev1_groups
         assert dev2_paramiko_opts["username"] == "root"
         assert "dev3.group_2" in hosts_dict
+
+    def test_add_group_to_host_runtime(self, inv):
```
dbarrosop, line 534: excellent. Even if we remove the `add_to_group` and `remove_from_group` methods I’d like to keep this tests as it helps testing the runtime of nornir.

dbarrosop:
Excellent PR, a few things have change in nornir3 making some of the comments in the issue obsolete (sorry about that) but I think the `add` method you proposed and the test you added are excellent ideas. Thanks!

nornir/core/inventory.py
kpeterson-sf: No good reason.  Will update.

nornir/core/inventory.py
kpeterson-sf: Makes sense.  it gets a lot easier without the ref.

nornir/core/inventory.py
kpeterson-sf: noted

tests/core/test_inventory.py
kpeterson-sf: 👍

nornir/core/inventory.py
kpeterson-sf: Good call.

nornir/core/inventory.py
```diff
@@ -105,6 +105,27 @@ def __contains__(self, value: object) -> bool:
...
+            self.append(group)
+
+    def remove(self, group: "Group") -> None:
+        """
+        Remove the parent group.
+
+        :param group: Group object to remove
+        :return: None
+        """
+        super().remove(group)
```
dbarrosop, line 127: I think this means the method isn't really necessary as we get the same behaviour without it, right? Or am I missing something?

dbarrosop:
apologies this is taking so long...

nornir/core/inventory.py
kpeterson-sf: Yeah don't see any reason it needs to stay there.

dbarrosop:
Closed in favour of #624 

Thanks!

Issue #594, opened 2020-09-25
Documentation issues


candlerb:
# Broken example for InitNornir

At https://nornir.readthedocs.io/en/3.0.0/configuration/index.html it says "An example using InitNornir would be:"

However the given example fails to run:

```
# python3
Python 3.6.9 (default, Jul 17 2020, 12:50:27)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from nornir import InitNornir
>>> nr = InitNornir(
...     core={"num_workers": 20},
...     logging={"file": "mylogs", "level": "debug"}
... )
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/usr/local/lib/python3.6/dist-packages/nornir/init_nornir.py", line 57, in InitNornir
    config = Config.from_dict(**kwargs)
  File "/usr/local/lib/python3.6/dist-packages/nornir/core/configuration.py", line 280, in from_dict
    logging=LoggingConfig(**logging or {}),
TypeError: __init__() got an unexpected keyword argument 'file'
>>>
```

I think logging can be corrected to:

```
logging={"log_file": "mylogs", "level": "DEBUG"}
```

but "num_workers" is also rejected and I don't know the replacement for that.

# Installation requirements for print_result

At https://nornir.readthedocs.io/en/3.0.0/tutorial/tasks.html it says `from nornir_utils.plugins.functions import print_result`

```
>>> from nornir_utils.plugins.functions import print_result
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'nornir_utils'
>>> import nornir_utils
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'nornir_utils'
```

It doesn't mention that a separate `pip3 install nornir_utils` is required.  This is not mentioned at https://nornir.readthedocs.io/en/3.0.0/tutorial/install.html either - perhaps that would be a better place for it.

It *does* say later at https://nornir.readthedocs.io/en/3.0.0/tutorial/task_results.html#The-easy-way that print_result "comes with the package `nornir_utils`"

dbarrosop:
Fixed via #598 

Thanks for raising the issue and sorry about that. Look at the linked PR/updated documentation for details.

PR #598, opened 2020-09-28
Fix InitNornir example in docs/configuration/index.rst

ubaumann:
Fix "Broken example for InitNornir" from #594

### DIFF
--- a/docs/howto/advanced_filtering/config.yaml
+++ b/docs/howto/advanced_filtering/config.yaml
@@ -4,4 +4,7 @@ inventory:
     options:
         host_file: "advanced_filtering/inventory/hosts.yaml"
         group_file: "advanced_filtering/inventory/groups.yaml"
-
+runners:
+    plugin: threaded
+    options:
+        num_workers: 20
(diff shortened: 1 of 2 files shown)


docs/howto/advanced_filtering/config.yaml
```diff
@@ -4,4 +4,7 @@ inventory:
     options:
         host_file: "advanced_filtering/inventory/hosts.yaml"
         group_file: "advanced_filtering/inventory/groups.yaml"
-
+runners:
```
ktbyers, line 7: Needs to be `runner` and not `runners`

ktbyers:
Everything else looks good except for the aforementioned `runners` to `runner` change.

docs/howto/advanced_filtering/config.yaml
ubaumann: Thank you for seeing it!

ktbyers:
FYI, the testing failures are due to other issues which I fixed in a separate PR.

dbarrosop (approved):

dbarrosop:
would you mind rebasing so the tests complete successfully? Thanks for the PR!

ubaumann:
Thanks for rebase it and thanks for all your time

PR #599, opened 2020-10-02
Fix documentation for log_file env variable

ogenstad:


ktbyers:
Updated here with fixes to breakage in CI-CD:

https://github.com/nornir-automation/nornir/pull/600

PR #600, opened 2020-10-02
Docs logging fix and Pin Poetry

ktbyers:


ktbyers:
It looks like Poetry 1.1.0 broke format for installing from a specific GitHub repository/branch (which we are doing for textfsm dev dependency).

I also ran into some issue with `make nbval`. I am pretty sure version of `pygments` was breaking things. Some of the other things I pinned might be unnecessary.

dbarrosop:
Thanks!

PR #613, opened 2020-11-21
fix "GitHub Actions: Deprecating set-env and add-path commands"

dbarrosop:
For more info: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -42,7 +42,10 @@ jobs:
           make sphinx
 
   pytest:
-    name: Testing on Python ${{ matrix.python-version }}
+    name: Testing on Python ${{ matrix.python-version }} (${{ matrix.platform}})
+    defaults:
+      run:
+        shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yaml
```diff
@@ -59,28 +59,30 @@ jobs:
...
-          poetry config virtualenvs.path ~/.virtualenvs
+          path: .venv
+          key: venv-${{ runner.os }}-${{ hashFiles('**/poetry.lock') }}
 
       - name: Install Dependencies
         run: poetry install
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
 
       - name: Run pytest
         run: make pytest
```
dbarrosop, line 80: no idea why this doesn't work

dbarrosop:
Merging without approval as the CI is broken. We should revisit the PR though.

PR #623, opened 2020-12-21
improvements to data resolution

### LINKED ISSUE #621: Inheritance Issue
There appears to be an inheritance issue as found by Netigo @ https://nornir.discourse.group/t/inheritance-problem/645

If you recreate the inventory files at https://github.com/nornir-automation/nornir/blob/develop/docs/tutorial/inventory.ipynb

Then try

```
leaf01_bma = nr.inventory.hosts["leaf01.bma"]
print(leaf01_bma["domain"])  
```
The data is supposed to come from the group `global`, but instead comes from defaults.yaml, ie acme.local.  It should be global.local as mentioned in the documentation.

I've attempted a possible patch but not sure how correct it is within the Nornir framework.

In `nornir/core/inventory.py`

Modify the group loop in `_resolve_data` 

    for g in self.groups:
        for k, v in g.data.items():
            if k not in processed:
                processed.append(k)
                result[k] = v

        for pg in g.groups:
            for k, v in pg.data.items():
                if k not in processed:
                    processed.append(k)
                    result[k] = v

Modify the group loop in `__getitem__`

        for g in self.groups:
            try:
                r = g.data[item]
                return r
            except KeyError:
                pass

            for pg in g.groups:
                try:
                    r = pg.data[item]
                    return r
                except KeyError:
                    pass

dbarrosop:
fixes #621 

When resolving data nornir would start climbing up the tree until reach the defaults, however, the desired behaviour was to climb the tree and then back down for the next group before checking reaching the defaults, which should be checked only as a last resort.

For instance, with the tree below:
```
         host
     g1       g2
  g11 g12
        defaults
```

Before the behaviour was `host -> g1 -> g11 -> defaults` and now it is `host -> g1 -> g11 -> g12 -> g2 -> defaults`

In addition, this introduces two new methods:

1. `extended_data`, which returns the data of a given host or group, including its inherited data
2. `extended_groups`, which includes a list of groups the host or group belongs to including its inherited groups

Finally, the `dict` method was resolving the attributes of the hosts and groups, which wasn't supposed to happen, it's been fixed

### DIFF
--- a/docs/tutorial/inventory.ipynb
+++ b/docs/tutorial/inventory.ipynb
@@ -877,7 +877,7 @@
     {
      "data": {
       "text/plain": [
-       "'acme.local'"
+       "'global.local'"
       ]
      },
      "execution_count": 13,
--- a/nornir/core/inventory.py
+++ b/nornir/core/inventory.py
@@ -55,11 +55,11 @@ class BaseAttributes(object):
 
     def dict(self) -> Dict[str, Any]:
         return {
-            "hostname": self.hostname,
-            "port": self.port,
-            "username": self.username,
-            "password": self.password,
-            "platform": self.platform,
+            "hostname": object.__getattribute__(self, "hostname"),
+            "port": object.__getattribute__(self, "port"),
+            "username": object.__getattribute__(self, "username"),
+            "password": object.__getattribute__(self, "password"),
+            "platform": object.__getattribute__(self, "platform"),
         }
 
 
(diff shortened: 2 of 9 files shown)


dbarrosop:
@no-such-anthony would you mind taking a look?

no-such-anthony:
@dbarrosop  Perfect, and the scenarios I had are all returning the correct inherited values.

PR #636, opened 2021-01-29
Create py.typed

Kircheneer:
PEP561 compliancy to enable type checking when Nornir is imported (https://www.python.org/dev/peps/pep-0561/).

This solves an issue where mypy raises "found module but no type hints or library stubs" errors when Nornir is imported to a type-checked module.


ktbyers:
Did you check/verify that this file will be included with Nornir when uploaded to pypi (given Nornir is using Poetry)?

Kircheneer:
> 
> 
> Did you check/verify that this file will be included with Nornir when uploaded to pypi (given Nornir is using Poetry)?

According to the [docs](https://python-poetry.org/docs/pyproject/#include-and-exclude) the include/exclude directives are seeded from the .gitignore file - which does not include/exclude the py.typed file. If I download the .tar.gz from Pypi however it does not include everything thats not excluded from the .gitignore. How is a release conducted generally?

ktbyers:
I think I would test it this way...:

1. Do `poetry build` and generate a wheel file in the `dist` directory.
2. Use `pip` to install that `wheel` file
3. Verify the `py.typed` file is included in the directory installed via pip.

I think that would probably be sufficient for verification.

Kircheneer:
I did the following (using Windows 10 with Git Bash):

```bash
$ poetry build
$ cd /other/directory
$ python -m venv venv
$ source venv/Scripts/activate
$ pip install /nornir/dir/dist/nornir-3.0.0-py3-none-any.whl
$ ls venv/Lib/site-packages/nornir
__init__.py  __pycache__/  core/  init_nornir.py  plugins/  py.typed
```

It looks like the file is indeed included by default.

ktbyers (approved):

ktbyers:
LGTM

PR #640, opened 2021-01-31
Add type-casting to nornire.core.task.Task.__repr__

### LINKED ISSUE #639: Task repr bug
# Summary
[`repr()`](https://docs.python.org/3/library/functions.html#repr) expects a string to be returned.  `nornir.core.task.Task.__repr__` returns the object stored in `self.name` without type-casting it.  This prevents the use of non-string objects for the `name` kwarg of a `nornir.core.task.Task`.

# Example - Break
```
from typing import Any

class Test:
    def __init__(self, test: Any) -> None:
        self.name = test
    def __repr__(self) -> str:
        return self.name

a = Test("this will work")
b = Test({"this": "will not work"})

print(a)

try:
    print(b)
except TypeError:
    print("TypeError")
```

# Example - Fix
```
from typing import Any

class Test:
    def __init__(self, test: Any) -> None:
        self.name = test
    def __repr__(self) -> str:
        return repr(self.name)

a = Test("this will work")
b = Test({"this": "will now work"})

print(a)
print(b)
```

# Test Results
Local run of `poetry run pytest` does not produce a failure.

amitmi704:
Closes #639

### DIFF
--- a/nornir/core/task.py
+++ b/nornir/core/task.py
@@ -73,7 +73,7 @@ class Task(object):
         )
 
     def __repr__(self) -> str:
-        return self.name
+        return str(self.name)
 
     def start(self, host: "Host") -> "MultiResult":
         """


ktbyers:
`self.name` should be a string. I would say your code is broke if it is not a string.

Not seeing why we need to do this? Can you expand on why it is necessarY?

PR #654, opened 2021-03-01
address UTF-8 error on windows

dbarrosop:
Windows throws the following error when trying to open a UTF-8 encoded file without specifying the encoding:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0xac in position 11744: illegal multibyte sequence
```

This PR allows users to specify the encoding of `SimpleInventory`


dbarrosop:
Looks like GH actions is having troubles and the CI isn't running so I am sure if I managed to trigger the error. In any case, the fix should work. I will check later if the issue is triggered in the CI without the fix.

@nmweizi would you mind testing this? thanks for reporting!

dbarrosop:
interestingly the error didn't trigger as described but it certainly was caught:

https://github.com/nornir-automation/nornir/runs/2006936942?check_suite_focus=true

Note the error only shows up with windows

ktbyers:
This looks good to me.


Kirk

nmweizi:
@dbarrosop it is fine.

dbarrosop:
Thanks!

Issue #666, opened 2021-03-31
Nornir state property returning GlobalState class, not object


brandomando:
https://github.com/nornir-automation/nornir/blob/e4f6b8c6258ae2dcfb286098b30652c7a31ecf30/nornir/core/__init__.py#L177

I'm not sure if this is intended behavior but it is a bit confusing to me why there would be a need to return the `GlobalState` class rather than the instantiated object like the `data` attribute, which is an instance of `GlobalState`.

dbarrosop:
`GlobalState` used to be a singleton but that's no longer true so I think this is a leftover. As a matter of fact, `state` is redundant as `data` is what we care about nowadays.

brandomando:
I thought that might be the case. I will submit a PR to remove it.

dbarrosop:
Thanks!

PR #669, opened 2021-03-31
remove state property from Nornir class

brandomando:
Fix for #666. No tests to update, and the code is de-referenced elsewhere in favor of the `data` attribute which is used now.


dbarrosop (approved):

dbarrosop:
Note: even though this violates semver rules I am merging it as this didn't work anyway so if someone is using this it is clearly not in its intended way

PR #694, opened 2021-06-19
update ruamel dependency

itdependsnetworks:


itdependsnetworks:
This is failing due to issue #695 not because of any changes made here.

pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.6"
-"ruamel.yaml" = "^0.16"
+"ruamel.yaml" = ">= 0.16, <0.18"
```
dbarrosop, line 28: Why `<0.18`?

itdependsnetworks: I did not want to be over permissive

itdependsnetworks: Let me elaborate, I did not want to be so permissive not knowing your initial intention

ktbyers:
@itdependsnetworks Can you rebase and update your poetry.lock file (to get David's mypy fix).

Also we should just do:

```
"ruamel.yaml" = "0.16.*"
```

That should allow 0.16 and greater, but disallow 1.x.x.

Alternatively we could just do >=0.16

itdependsnetworks:
I went with `Alternatively we could just do >=0.16`, since semantic versioning technically treats every "minor" version as potentially non-backwards compatible, and poetry respects that.

dbarrosop (approved):

dbarrosop:
Thanks!

PR #697, opened 2021-06-20
check links in documentation aren't broken

dbarrosop:


dteycheney:
This seems reasonable to me.

PR #729, opened 2021-10-14
Fix GitHub actions MacOS arch failure

ktbyers:


dbarrosop:
I have been checking their docs and I think the problem is that they switched `macos-latest` from `macos-10` to `macos-11`.

https://github.com/nornir-automation/nornir/runs/3899700972?check_suite_focus=true#step:1:7

Checking:

https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

and 

https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md

Looks like macos-10 supports py36 while macos-11 doesn't.

So maybe we should change `macos-latest` to `macOS-10.15` instead of excluding py3.6? What do you think?

ktbyers:
@dbarrosop Okay, I think it is good to go now.

dbarrosop:
Thanks!

PR #732, opened 2021-10-19
Update pygments because of vulnerabilities

ubaumann:
Updating pygments because of two vulnerabilities

Minumum version would be 2.7.4 but I guess we can update direclty to 2.10.0?

- https://github.com/advisories/GHSA-9w8r-397f-prfh
- https://github.com/advisories/GHSA-pq64-v7f5-gqh8


dbarrosop:
Thanks for the PR. Yes, I think updating to the latest versionwould be fine, however, we are going to have to re-run and save the notebooks as looks like the rendering is slightly different and tests are failing.

Also, note that these vulnerabilities are not exploitable as we only use pygments to render jupyter notebooks for documentation purposes and we control the code rendered.

PR #780, opened 2022-02-18
Deprecate py3.6

ubaumann:
- remove python 3.6 support
- update dependencies
- add python 3.10


pyproject.toml
```diff
@@ -53,25 +53,25 @@ pytest = "*"
...
-sphinx-issues = "^1.2"
+sphinx-issues = "^3.0"
 
 # for docs only
-## this needs to be fixed, poetry is giving issues due to nornir being beta
-nornir-jinja2 = "0.1.2"
+nornir-jinja2 = "0.2.0"
 nornir-utils = "0.1.2"
 nornir-napalm = "0.1.2"
-# due to https://github.com/google/textfsm/issues/63
```
ubaumann, line 73: @ktbyers This is fixed in version 1.1.2 right?

ktbyers: Yes, that is fixed in >= 1.1.2 (so we should require at least textfsm version 1.1.2 in the dev-dependencies).

ubaumann: Yes I did on line 74
`textfsm = "^1.1.2"`

ubaumann: I should have made the comment on line. 74. I had in mind it shows 2 lines before and after

dbarrosop (approved):

dbarrosop:
Amazing work

PR #802, opened 2022-05-21
Added plugins architecture image to plugins docs

dmulyalin:
Added diagram to docs describing plugins architecture to make it easier for people to comprehend the greatness of Nornir.


dbarrosop:
Thanks, this is pretty awesome. If you don't mind, I'd like to suggest two minor changes:

1. "External database" -> "3d party". Technically it can be a database, a service, a file, whatever.
2. I'd remove the arrow from "transform functions", just leaving the box there. Technically it interacts with nornir's inventory and it doesn't have any knowledge of the inventory plugin

dmulyalin:
Thanks for checking it out and for suggestions, updated the diagram.

dbarrosop:
Awesome, thanks a lot, this is great!

PR #820, opened 2022-09-18
DockerFile Poetry Install Update

h4ndzdatm0ld:
The previous get-poetry.py and install-poetry.py installers are deprecated. 

- Changed env PATH & CURL endpoint


dbarrosop (approved):

dbarrosop:
Thanks!

PR #823, opened 2022-09-27
log number of hosts to run on correctly

txSangyj:


dbarrosop:
Apologies for the long delay, thanks for this!

PR #835, opened 2023-02-09
Upgrade poetry and readthedocs to use dependency groups

ktbyers:


ktbyers:
Items in this PR:
- Commented out codeql GitHub Action as it wasn't working (and wanted to focus on getting core CI working). Do we still need this? Is it worth fixing?
- Commented out publish in GitHub Actions as it is broken (I assume for same reason as nornir_napalm) as they are both failing in the same way. I figured we would fix this on the next publish or when we fix nornir_napalm since it is the same issue.
- Upgraded to poetry 1.3.2
- Simplified pyproject.toml to use new dependency group format which is now supported by readthedocs (poetry install --with docs). Simplified some of the pyproject.toml structure and removed some things that didn't seem necessary any longer.
- Upgraded .readthedocs.yaml to use RTD build V2 and new poetry dependency groups.

ktbyers:
I tested this here and also on my read the docs account here:

https://nornir-ktbyers.readthedocs.io/en/poetry_test/

I also did some testing on an ubuntu machine, but that is probably redundant with the tests that are done on GitHub actions.

ktbyers:
@dbarrosop I think this is good to go.

ktbyers:
Poetry now allows creating dependency groups:

```
[tool.poetry.group.docs.dependencies]
sphinx = "4.5.0"
sphinx_rtd_theme = "1.0.0"
sphinxcontrib-napoleon = "^0.7"
jupyter = "1.0.0"
nbsphinx = "0.8.8"
pygments = "2.11.2"
sphinx-issues = "3.0.1"
```

dbarrosop:
This is pretty amazing! Thanks!

PR #836, opened 2023-02-17
Fix pylama breakage (pep8 reference no longer works)

ktbyers:
It came to my attention on both the NAPALM and Netmiko repositories that pylama . no longer worked (for calling the linter).

Basically 'pycodestyle (pep8)' was no longer executed and it would just silently pass (even when you had linting issues). This update should fix this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,15 +10,15 @@ universal=1
 # E731: do not assign a lambda expression, use a def
 # D10?: missing docstrings
 [pylama]
-linters = mccabe,pep8,pyflakes
-ignore = D203,C901
+linters = mccabe,pycodestyle,pyflakes
+ignore = D203,C901,W503
 skip = .tox/*,.venv/*,nornir/_vendor/*
 
-[pylama:pep8]
+[pylama:pycodestyle]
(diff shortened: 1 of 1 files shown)


ktbyers:
The W503 linting exclusion was causing linting issues (where what black would correct to would actually fail the linter):

https://stackoverflow.com/questions/57074300/what-is-the-recommended-way-to-break-a-long-if-statement-w504-line-break-after?rq=1

It looked like pycodestyle would always fail regardless of the form I used with either a W503 or a W504 error...so I just chose the one that black used (and disabled the other one).

ktbyers:
@dbarrosop Can you disable this in GH settings (at least until someone fixes its failures):

```
CodeQL Expected — Waiting for status to be reported
```

I don't have access to settings for the Nornir repository (or I would disable it).

Thanks.

dbarrosop (approved):

dbarrosop:
> Can you disable this in GH settings (at least until someone fixes its failures):

Done.

> I don't have access to settings for the Nornir repository (or I would disable it).

Made you owner of the full org so you should access to the whole thing now.

PR #864, opened 2023-09-21
add isort to manage/sort imports

nemith:
Add `isort` as a way to cleanup and manage imports in addition to `black`.

Black is great tool and will format the import lines, but doesn't sort them, remove unused imports, or sorts them but it does work well with the `isort` (import sort) tool.

https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort

Black was developed by Łukasz which at facebook and there we always used black along side isort as well. 

See https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort


dbarrosop:
This is fine by me, I am all for delegating as many opinions as possible to intransigent machines so if Kirk is fine with this too we can merge after the conflicts has been resolved.

dbarrosop (approved):

nornir/core/processor.py
```diff
@@ -1,5 +1,7 @@
 from typing import List, Protocol
 
+from typing_extensions import Protocol
```
ktbyers, line 3: @nemith It looks like you need to rebase to get the latest changes that I released last night.

ktbyers:
Replaced by:

https://github.com/nornir-automation/nornir/pull/873

It was easier to just redo the PR than to fix the conflicts.

Issue #861, opened 2023-09-21
mypy_extensions is only declared as a dev dependency but it directly imported


nemith:
We are failing to run some CI tests as mypy_extensions is missing 

```
/usr/local/lib/python3.8/site-packages/nornir/__init__.py:3: in <module>
    from nornir.init_nornir import InitNornir
/usr/local/lib/python3.8/site-packages/nornir/init_nornir.py:3: in <module>
    from nornir.core import Nornir
/usr/local/lib/python3.8/site-packages/nornir/core/__init__.py:6: in <module>
    from nornir.core.inventory import Inventory
/usr/local/lib/python3.8/site-packages/nornir/core/inventory.py:23: in <module>
    from mypy_extensions import Arg, KwArg
E   ModuleNotFoundError: No module named 'mypy_extensions'
```

Looking at  https://github.com/nornir-automation/nornir/blob/main/pyproject.toml#L43 it is only declaring mypy_extentions as a dev dependency but in this case it is not a dev dependency but   

My guess is something else changes in our dep tree that removed mypy_extension but it needs to be explicitly declared as a runtime dependency as it is being directly imported here https://github.com/nornir-automation/nornir/blob/main/nornir/core/inventory.py#L23

nemith:
Seems like using these are deprecated anyway. https://mypy.readthedocs.io/en/stable/additional_features.html?highlight=mypy_extensions#extended-callable-types

ubaumann:
Nornir was just released some hours ago but we missed this line: https://github.com/nornir-automation/nornir/blob/325c8ca479b08662a0f963a90d3328a34ef17f2a/nornir/core/inventory.py#L23

I will take a look when I am home.
In the meantime you can pin the nornir version or add it to your dependencies

ktbyers:
Fixed here:

https://github.com/nornir-automation/nornir/pull/867

New Nornir version 3.4.1 released to pypi.

PR #867, opened 2023-09-22
Version 3.4.1, fix missing dependencies

ktbyers:
Correct issue with broken typing_extensions (which was caused by earlier changes I made to pyproject.toml and missing tests where we should catch this type of issue).


ktbyers:
@nemith I reimplemented this as I wanted to remove some of the clutter (i.e. the isort changes).

I also re-added mypy_extensions as I wanted to keep the changes close to the minimum set needed to fix the issue with the broken release (and looking into Callback Protocols as the typing replacement for the Extended Callback Types was taking too long to research/understand properly).

I am totally open to removing mypy_extensions we should just look to do it as part of the next release.

ktbyers:
I also tested that this imported properly from a poetry built wheel version of 3.4.1 

```
# In blank/new virtual environment
$ pip install nornir-3.4.1-py3-none-any.whl

$ python
Python 3.11.0 (main, Nov 17 2022, 14:54:18) [GCC 7.3.1 20180712 (Red Hat 7.3.1-15)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import nornir
>>> nornir.__version__
'3.4.1'
```

ktbyers:
@dbarrosop @ubaumann 

FYI, I am going to push out a new Nornir release to fix the currently broken release. See issue #https://github.com/nornir-automation/nornir/issues/861

This was caused by the pyproject.toml restructuring/fixes that I did earlier (i.e. imports that were not dev imports got categorized as dev imports) and that this step is not tested currently (i.e. we had no feedback from CI-CD system that it was broken...I will see if I can fix that later).

Thanks to @nemith for his reporting of the issue and doing a bunch of the work on the fix. We should at the removal proposal of mypy_extensions that he recommended in his other PR, but I didn't have time to research that tonight.

nemith:
Yeah these being only used for type checking it should be fairly low risk, but so should reoranging a pyproject.toml file so I understand.  

For next minor releases how do you queue the changes?  Are you ok for me resubmitting another PR to reapply removing mypy_extensions?

ktbyers:
@nemith Yes, definitely go ahead and resubmit the PR to remove `mypy_extensions`.

There is a good chance it will be a while before there is another minor release, but it is better to get this done in the near future when it is fresh in everyone's mind (or else it will probably just get put onto the back burner until the deprecated feature is actually removed).

Issue #868, opened 2023-09-22
Remove mypy_extensions this has been deprecated


ktbyers:

ktbyers:
See parts of PR here for reference:

https://github.com/nornir-automation/nornir/pull/863

PR #870, opened 2023-09-22
remove mypy_extensions dep

nemith:
While looking into #861 I noticed that the way the Callable and with the arguments from `mypy_extension` is deprecated and no longer needed.  This is supported natively in typing in 3.8 or greater with the `Protocol` type.

So this moves to using `Protocol` with a `__call__` stub instead of the `mypy_extensions`.

Fixes: #868


ktbyers (approved):
LGTM

PR #874, opened 2023-10-03
Add isort to dev dependencies

ktbyers:


ktbyers:
I removed "docs" from the directories that isort sorts. Someone else can do that PR if they want, but it is not me :-)

So isort only checks "nornir" and "tests" directories.

Also upgraded black to a newer version (as now is a good time to do it).

ktbyers:
I added "docs" back into isort as the "nbval" issues needed fixed anyways.

ktbyers:
The only file changes that should matter are:

```
Makefile
pyproject.toml
.github/workflows/codeql-analysis.yml      # removal
```

I think everything else is just minor auto-formatting changes made by isort/black or minor testing issues (nbval testing) that arose because of those changes.

dbarrosop (approved):

Issue #880, opened 2023-11-29
Async architecture (inventory)


ktbyers:
This issue is `async inventory` related only, see:

https://github.com/nornir-automation/nornir/issues/881

For additional async related topics and links to separate issues on these topics.

ktbyers:
@dgarros @ogenstad

ktbyers:
Some initial work/thoughts from dbarroso. Note, some of this was from the early days of asyncio so some patterns may or may not be relevant

https://github.com/nornir-automation/nornir/pull/390
https://github.com/nornir-automation/nornir/pull/660

ktbyers:
I think we should start with a discussion regarding Inventory and async and what this means/how it works? 

Currently, inventory is single-threaded sync (it is outside the context of "runners"/before runners). 

So if we have, for example, a NetBox system with 1000 devices and wanted an async inventory plugin how would this operate? I picked NetBox as there is a NetBox Nornir-plugin and I have a working NetBox system to test against (but it could be any remote HTTP API inventory system).

dbarrosop:
> discussion regarding Inventory and async

I think the first question I'd raise here is; is it really necessary? do we actually need an async inventory for any particular reason or is it just to say we support async inventories? Asking because the inventory is a one-shot thing you run when you initialize it, afterwards you mostly work with already loaded data. I'd understand async tasks to incrementally add information (i.e. a sync inventory that loads the hosts/groups and metadata but interfaces' data, BGP information, etc. is loaded afterwards via async tasks) but not sure I understand the use case for an async inventory.

ktbyers:
I would probably generalize the question and ask, "do we need concurrency in inventory loading" (whether async vs threaded). 

In other words, for large inventories being loaded via API what are the load times and what kind of improvements can be made via using a concurrent solution (and how much complexity does that entail).

dbarrosop:
> if an inventory plugin would have an option for a minimal import ... And if more information is required it could be added to a task

100% that, that's why I was suggesting before that what makes more sense is leaving the inventory as it is and rely on "inventory tasks" (which could be async) to load workflow-specific data.

IMO, given there is no clear use-case in mind, I'd suggest tabling this particular issue until something more concrete arises.

P.S: To keep discussions a bit easier to track I'd also suggest to create different issues for each one of your bulletpoints, otherwise this might get messy real quick :P

ktbyers:
[...]

PR #882, opened 2023-11-30
Proposal to support Async inventory 

dgarros:
Related to #880 

This PR is an example of implementation of how Nornir could support both sync and async inventory without much effort and without having to duplicate the code.

The main idea is to have the main code written in Async and wrap these functions with asyncio.run to also support non async execution.


dgarros:
Closing for now, we'll revisit in the near future

PR #888, opened 2023-12-20
WIP: Inventory data plugin

ubaumann:
This pull request is a quick and incomplete implementation of adding an Inventory Data Plugin. The idea is to have a base to discuss this idea and what would be needed to accept such a change.

At the moment, the inventory data is stored in a Python dictionary. This works great. However, I found myself in a situation where I would like to replace the dictionary with objects. More specifically, I would like to use a pydantic model for the data structure. Firstly, to have input validation and, secondly, to be able to add functions to the pedantic objects to reduce logic in templates.

This PR adds a new plugin type called InventoryDataPlugin. The idea is to allow the default dict object to be replaced with another object. For now, I implemented it in a way that does not break any tests, as the new datatype used to store the data needs to provide methods similar to a dictionary. The code and plugin are most likely more complex because of this.

To use pydantic I would try to implement a plugin to use a model similar to this (just "random" pseudocode, untested):
```python
from typing import (
    Any,
    ItemsView,
    KeysView,
    ValuesView,
)

from pydantic import BaseModel, ConfigDict
from pydantic.networks import IPvAnyAddress

class User(BaseModel):
    model_config = ConfigDict(
        strict=True
    )
    name: str
    age: int
    is_active: bool
    ip: IPvAnyAddress

    def __getitem__(self, key: str) -> Any:
        return getattr(self, key)
    
    def get(self, key: str, default: Any = None) -> Any:
        try:
            return self.__getitem__(key)
        except AttributeError:
            return default
    
    def __setitem__(self, key: str, value: Any):
        self.model_validate({**self.__dict__, **{key: value}})
        setattr(self, key, value)
    
    def keys(self) -> KeysView[str]:
        return self.__dict__.keys()
    
    def values(self) -> ValuesView[Any]:
        return self.__dict__.values()
    
    def items(self) -> ItemsView[str, Any]:
        return self.__dict__.items()

```

The model, of course, would be much more complex with nested models. What already works well is to store multiple models in the data dictionary. So, in this case, only the first level would be a dictionary.

### DIFF
--- a/nornir/init_nornir.py
+++ b/nornir/init_nornir.py
@@ -1,3 +1,4 @@
+import inspect
 from typing import Any
 
 from nornir.core import Nornir
@@ -17,7 +18,17 @@ def load_inventory(
 ) -> Inventory:
     InventoryPluginRegister.auto_register()
     inventory_plugin = InventoryPluginRegister.get_plugin(config.inventory.plugin)
-    inv = inventory_plugin(**config.inventory.options).load()
+    inventory_plugin_params = config.inventory.options.copy()
+
+    init_params = inspect.signature(inventory_plugin).parameters
+    if "configuration" in init_params:
+        config_parameter = init_params["configuration"]
+        if config_parameter is not inspect.Parameter.empty and issubclass(
+            config_parameter.annotation, Config
+        ):
+            inventory_plugin_params.update({"configuration": config})
+
+    inv = inventory_plugin(**inventory_plugin_params).load()
 
     if config.inventory.transform_function:
         TransformFunctionRegister.auto_register()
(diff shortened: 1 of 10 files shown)


nornir/init_nornir.py
```diff
@@ -17,7 +18,18 @@ def load_inventory(
 ) -> Inventory:
     InventoryPluginRegister.auto_register()
     inventory_plugin = InventoryPluginRegister.get_plugin(config.inventory.plugin)
-    inv = inventory_plugin(**config.inventory.options).load()
+    inventory_plugin_params = config.inventory.options.copy()
+
+    init_params = inspect.signature(inventory_plugin).parameters
```
ubaumann, line 23: This code is most likely not working yet. The idea would be to detect if the Inventory plugin allows passing down the configuration settings to the Host, Group and Defaults Inventory objects. This would allow us to avoid breaking existing inventory plugins.

ubaumann:
PS: The tests are failing as the package is not installed again in the pipeline because of cashing.

PR #913, opened 2024-06-08
Add property for optional runner

ogenstad:
The `runner` argument to Nornir is defined as being optional but we later use it as if it's always set. Adding a property that returns a `RunnerPlugin` or raises an error, with this we can also remove the `strict_optional = False` bypass in the mypy config.

### DIFF
--- a/nornir/core/__init__.py
+++ b/nornir/core/__init__.py
@@ -47,7 +47,7 @@ class Nornir(object):
         self.inventory = inventory
         self.config = config or Config()
         self.processors = processors or Processors()
-        self.runner = runner
+        self._runner = runner
 
     def __enter__(self) -> "Nornir":
         return self
@@ -65,14 +65,16 @@ class Nornir(object):
         Given a list of Processor objects return a copy of the nornir object with the processors
         assigned to the copy. The orinal object is left unmodified.
         """
-        return Nornir(**{**self.__dict__, **{"processors": Processors(processors)}})
+        return Nornir(
+            **{**self._clone_parameters(), **{"processors": Processors(processors)}}
+        )
 
     def with_runner(self, runner: RunnerPlugin) -> "Nornir":
         """
         Given a runner return a copy of the nornir object with the runner
         assigned to the copy. The orinal object is left unmodified.
         """
-        return Nornir(**{**self.__dict__, **{"runner": runner}})
+        return Nornir(**{**self._clone_parameters(), **{"runner": runner}})
 
     def filter(self, *args: Any, **kwargs: Any) -> "Nornir":
         """
@@ -81,7 +83,7 @@ class Nornir(object):
         Returns:
             :obj:`Nornir`: A new object with same configuration as ``self`` but filtered inventory.
         """
-        b = Nornir(**self.__dict__)
+        b = Nornir(**self._clone_parameters())
         b.inventory = self.inventory.filter(*args, **kwargs)
         return b
 
@@ -143,7 +145,7 @@ class Nornir(object):
         else:
             logger.warning("Task %r has not been run – 0 hosts selected", task.name)
 
-        result = self._runner.run(task, run_on)
+        result = self.runner.run(task, run_on)
 
         raise_on_error = (
             raise_on_error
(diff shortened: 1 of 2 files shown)


dbarrosop:
I wonder if it's better to change the signature instead

ogenstad:
That would be cleaner for sure, I didn't want to change anything that could potentially break someones workflow. I don't know if it would be an issue but potentially someone could do something like this now:

```python
nr = Nornir()
my_nr = nr.with_runner(runner=xyz)
```

My guess is that most people would be using `InitNornir` where the runner is always set but I don't know if this would be considered enough of an issue?

Alternatively load_runner could be moved so that it's called here again if runner isn't set.


```python
self.runner = runner or load_runner(self.config)
```

dbarrosop:
I honestly would consider this a bug. It's also a typing issue so it will only break linters/development, it shouldn't break the runtime, should it?

ogenstad:
It's a typing issue but depending on how it's solved it could be problematic in some use cases.

```python
    def __init__(
        self,
        inventory: Inventory,
        config: Optional[Config] = None,
        data: Optional[GlobalState] = None,
        processors: Optional[Processors] = None,
        runner: Optional[RunnerPlugin] = None,
    ) -> None:
        self.data = data if data is not None else GlobalState()
        self.inventory = inventory
        self.config = config or Config()
        self.processors = processors or Processors()
        self.runner = runner
```

The cleanest way to change this would be to make `runner` required as below, a potential problem here is if anyone today uses Nornir() without a runner and then adds one later (it's possible this never happens but hard to know):

```python
    def __init__(
        self,
        inventory: Inventory,
        runner: RunnerPlugin,       <--- Change
        config: Optional[Config] = None,
        data: Optional[GlobalState] = None,
        processors: Optional[Processors] = None,
    ) -> None:
        self.data = data if data is not None else GlobalState()
        self.inventory = inventory
        self.config = config or Config()
        self.processors = processors or Processors()
        self.runner = runner
```

An alternative would be:

```python
    def __init__(
        self,
        inventory: Inventory,
        config: Optional[Config] = None,
        data: Optional[GlobalState] = None,
        processors: Optional[Processors] = None,
        runner: Optional[RunnerPlugin] = None,
    ) -> None:
        self.data = data if data is not None else GlobalState()
        self.inventory = inventory
        self.config = config or Config()
        self.processors = processors or Processors()
        self.runner = runner or load_runner(self.config)       <--- Change
```

dbarrosop:
got it, I am starting to think the proposal on this PR is probably the best option. I don't like it is a runtime issue but we'd have to change the signature for this to be a "clean" fix as that's probably not ideal...

nornir/core/__init__.py
```diff
@@ -168,6 +169,13 @@ def close_connections_task(task):
 
         self.run(task=close_connections_task, on_good=on_good, on_failed=on_failed)
 
+    @property
+    def _runner(self) -> RunnerPlugin:
```
dbarrosop, line 173: shouldn't we do the opposite? store in `_runner` and have the propery be `runner`?

ogenstad: I tried that first however it breaks some other things, in the below functions we use `self.__dict__` which causes issues with an attributed such as `self._runner` as we'd send in "_runner" as an argument to Nornir instead of "runner".

```python
    def with_processors(self, processors: List[Processor]) -> "Nornir":
        """
        Given a list of Processor objects return a copy of the nornir object with the processors
        assigned to the copy. The orinal object is left unmodified.
        """
        return Nornir(**{**self.__dict__, **{"processors": Processors(processors)}})

    def with_runner(self, runner: RunnerPlugin) -> "Nornir":
        """
        Given a runner return a copy of the nornir object with the runner
        assigned to the copy. The orinal object is left unmodified.
        """
        return Nornir(**{**self.__dict__, **{"runner": runner}})

    def filter(self, *args: Any, **kwargs: Any) -> "Nornir":
        """
        See :py:meth:`nornir.core.inventory.Inventory.filter`

        Returns:
            :obj:`Nornir`: A new object with same configuration as ``self`` but filtered inventory.
        """
        b = Nornir(**self.__dict__)
        b.inventory = self.inventory.filter(*args, **kwargs)
        return b
```

What we could do instead is to replace `self.__dict__` in these cases with a property that returns a dictionary where the keys are hardcoded instead, i.e: inventory, data, config, runner, processors.

dbarrosop: yeah, either that or just a cleaner clone method

dbarrosop (approved):

PR #914, opened 2024-06-08
Fix CI

ogenstad:
The PR follows #909, the problem seemed to be that we switched to `windows-2019` instead of `windows-latest` and a few years ago the windows tests seem to have been disabled all together but started to fail again as we only compared against `windows-latest` we can probably look at reactivating these in a later PR.

Also upgraded executing as version 2.0.0 was yanked. I saw that this was failing in the Windows install log initially and this part was cached for all other jobs, that change would be unrelated to this but probably doesn't hurt either.


ktbyers:
Approving to get CI/CD running again.

We should switch to windows-2022 and also probably switch the caching structure to new form that uses setup-python cache argument (much easier). Should also upgrade some of the GH-actions to their newer versions.

But we can do all those in a newer PR.

ktbyers (approved):

dbarrosop:
thx

PR #918, opened 2024-06-10
Add `setuptools` to dependencies

### LINKED ISSUE #917: `setuptools` is not listed as a dependency
Importing Nornir in a Python 3.12 virtual environment fails because `setuptools` is not installed:

```
    from nornir import InitNornir
.venv/lib/python3.12/site-packages/nornir/__init__.py:1: in <module>
    import pkg_resources
E   ModuleNotFoundError: No module named 'pkg_resources'
```

In Python 3.12, `setuptools` is no longer installed in virtual environments (https://docs.python.org/3/whatsnew/3.12.html):

> [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environment.

Manually installing `setuptools` works, but it should be listed as an explicit dependency of Nornir.

I'd make a PR, but I'm not sure what version constraint to use.

devon-mar:
Fixes #917 

`setuptools` 42.0.0 is the first version that supports Python 3.8, the minimum Python requirement.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1496,6 +1496,16 @@ files = [
     {file = "MarkupSafe-2.1.3-cp311-cp311-musllinux_1_1_x86_64.whl", hash = "sha256:5bbe06f8eeafd38e5d0a4894ffec89378b6c6a625ff57e3028921f8ff59318ac"},
     {file = "MarkupSafe-2.1.3-cp311-cp311-win32.whl", hash = "sha256:dd15ff04ffd7e05ffcb7fe79f1b98041b8ea30ae9234aed2a9168b5797c3effb"},
     {file = "MarkupSafe-2.1.3-cp311-cp311-win_amd64.whl", hash = "sha256:134da1eca9ec0ae528110ccc9e48041e0828d79f24121a1a146161103c76e686"},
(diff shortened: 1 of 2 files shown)


ktbyers:
@devon-mar Can you rebase and update the poetry.lock file? There were some other changes recently to the lock file.

devon-mar:
@ktbyers Done

ktbyers (approved):

ktbyers:
setuptools is no longer guaranteed to be installed in PY3.12 and later:

https://docs.python.org/3/whatsnew/3.12.html

```
[gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools 
in virtual environments created with [venv](https://docs.python.org/3/library/venv.html#module-venv). 
This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by 
default; to access these run pip install setuptools in the 
[activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environment.
```

Issue #919, opened 2024-06-10
Suggestion: Introduce Ruff


ogenstad:
Introduce [Ruff](https://docs.astral.sh/ruff/) and have it replace:

* Black
* isort
* pylama

While Ruff also has checks for type annotations the recommendation is still to keep using mypy.

Thoughts around this @dbarrosop & @ktbyers?

dbarrosop:
I have no opinion, as long as it covers the same functionality (or even more) than the ones being replaced I am good. The formatter is the only thing that I am not sure about. Is it as extremely opinionated as black? I may not always agree with black but I like that I (or anyone else) have no choice other than to abide by it :)

ogenstad:
Ruff will have more functionality, it's pretty much as opinionated as Black (https://docs.astral.sh/ruff/formatter/#black-compatibility), there are some slight changes depending on the codebase there could be a small number of lines that change during the migration.

I think it's nice as it's fewer tools to deal with, it's also faster though that doesn't have much of an impact on this codebase.

ogenstad:
Fixed in #926 and #930

PR #926, opened 2024-06-19
Add linting with ruff

ogenstad:
This PR adds ruff as a linting tool, most of the rules have been activated even though some of the rules might never be relevant for Nornir.

Looking at the activated rules I've also added a number of rules to be ignored for later consideration, a number of these would result in a cleaner code base if fixed.

Another issue is that the change to the complexity rules 

```toml
[tool.ruff.lint.mccabe]
max-complexity = 12
```

This was previously set to 10 with pylama, however the rule was also altogether disabled, so it didn't add anything.

For now I didn't disable the pylama checks, but that can be done in a follow up PR.

Also I didn't activate the formatting rules of ruff to replace black. There would be a number of minor formatting changes required such as:

```diff
 class FilterObj(Protocol):
-    def __call__(self, host: Host, **kwargs: Any) -> bool:
-        ...
+    def __call__(self, host: Host, **kwargs: Any) -> bool: ...


 class Inventory(object):
@@ -607,13 +595,11 @@
         self,
         filter_obj: Optional[FilterObj] = None,
         filter_func: Optional[FilterObj] = None,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> "Inventory":
         filter_func = filter_obj or filter_func
         if filter_func:
-            filtered = Hosts(
-                {n: h for n, h in self.hosts.items() if filter_func(h, **kwargs)}
-            )
+            filtered = Hosts({n: h for n, h in self.hosts.items() if filter_func(h, **kwargs)})
         else:
             filtered = Hosts(
                 {
```

Related to #919


dbarrosop:
LGTM, i'd suggest different PRs to disable pylama and replace black. Thanks!

dbarrosop (approved):

PR #929, opened 2024-06-24
Remove unnecessary `pass` statement (FIX: PIE790)

ogenstad:
These pass statements doesn't do anything as they are after the doc strings


ktbyers:
I know it is bikeshedding, but yuck.

ktbyers (approved):

PR #933, opened 2024-06-27
Fix Sphinx add custom css issue

### LINKED ISSUE #932: Sphinx documentation is broken
The build process for Sphinx is broken since some time back. The problem originates from the fact that we don't even check it in the CI. We have a job to validate `make sphinx` however it doesn't do anything:
https://github.com/nornir-automation/nornir/blob/v3.4.1/Makefile#L25

In #855 Sphinx was upgraded from version 4.5.0 to version 6.2.1 since then the builds at https://readthedocs.org/projects/nornir/builds/ have started to fail but because we didn't validate anything within the CI nobody noticed.

I'm not sure if there are any vulnerabilities in the earlier version or why it was upgraded (if there was a reason). If it doesn't have any issues with security I'd suggest to downgrade and activate the CI job just to have it tracked and then look at what's needed to upgrade. Alternatively check what is required for it to work using the current version.

Also on Read the Docs it looks like it would build for for the `stable` or `develop` branches that doesn't exist, but not for `main`. Could be that I'm just missing something as I don't have any permissions on RTD. @dbarrosop, can you perhaps have a look and see if you can activate the builds for the `main` branch as well?

ubaumann:
Fixes #932

I resolved the issue with the sphinx build and removed the need for the `sphinx` make command, as there is already a `docs` command available.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: Run sphinx
         run: |
           sudo apt-get install pandoc
-          make sphinx
+          make docs
 
   pytest:
     name: Testing on Python ${{ matrix.python-version }} (${{ matrix.platform}})
--- a/Makefile
+++ b/Makefile
@@ -14,12 +14,6 @@ docker:
 pytest:
 	poetry run pytest --cov=nornir --cov-report=term-missing -vs ${ARGS}
 
-.PHONY: sphinx
-sphinx:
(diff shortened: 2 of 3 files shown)


ogenstad:
LGTM, added some comments around the warnings here: https://github.com/nornir-automation/nornir/issues/932#issuecomment-2195572378

I do think it would be good if we could get the builds to start to run on the `main` branch too, it doesn't have to be done before this is merged but would be nice.

ktbyers (approved):

ktbyers:
Okay, it looks like we will need @dbarrosop to fire off a readthedocs build to make sure this is working now.

PR #958, opened 2024-07-01
Remove unnecessary assignment before `return` statement

ogenstad:


ktbyers (approved):

ktbyers:
Ruff is convincing me not to use it in other projects (LOL).

ogenstad:
> Ruff is convincing me not to use it in other projects (LOL).

Well it mostly reimplements rules from other linters, though it has a few rules of its own too. I think most of the rules make sense and adds clarity. As with other linkers it's always easier if it's in place when you start coding instead of going back and adding it later.

ktbyers:
Yeah, it doesn't really matter (as with almost all linter things--it just matters that it is done/decided automatically).

PR #961, opened 2024-07-09
Add Python 3.12 to test matrix

ogenstad:


ktbyers:
@ogenstad I would probably try rolling nornir-napalm to 0.5.0 (since your error seems related to outdated ncclient which is possibly being caused by older nornir-napalm):

https://github.com/nornir-automation/nornir/blob/main/pyproject.toml#L46

And it should be updated anyways.

ogenstad:
@ktbyers, thanks. Looks like it got a bit further but hit the next problem. Might be that we just need to upgrade nbval, I'll try to return to it a bit later.

ktbyers (approved):

ktbyers:
LGTM

PR #971, opened 2024-08-24
Replace pkg_resources with importlib.metadata

ubaumann:
`pkg_resources` is deprecated in Python 3.12 (https://setuptools.pypa.io/en/latest/pkg_resources.html), and `importlib.metadata` is available since Python 3.8 (https://docs.python.org/3.12/library/importlib.metadata.html)

### DIFF
--- a/nornir/__init__.py
+++ b/nornir/__init__.py
@@ -1,7 +1,7 @@
-import pkg_resources
+from importlib.metadata import version
 
 from nornir.init_nornir import InitNornir
 
-__version__ = pkg_resources.get_distribution("nornir").version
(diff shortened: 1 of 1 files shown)


ogenstad:
Is setuptools (https://github.com/nornir-automation/nornir/blob/main/pyproject.toml#L34) still required after this?

ktbyers:
@ogenstad I am pretty sure we still need `setuptools` and as of Python 3.12 we have to explicitly include it as a dependency.

ktbyers (approved):
LGTM

PR #973, opened 2024-09-20
Update type hints within ./tests

ogenstad:
This PR corrects most of the type hints within ./tests. There are a few remaining for which I added more specific overrides to ignore those rules.


ktbyers:
Why?

dbarrosop (approved):

Issue #975, opened 2024-11-11
license metadata on pypi.org


elveshoern32:
nornir is open source software licensed under the Apache Licence 2.0, isn't it?
At least I find no other statement within nornir's project page at github.com .

However, on pypi.org it looks different:
Beginning with release 3.2.0 there is (additionally to the Apache License) a reference to some 'Other/Proprietary License'.
Is this change of license just a bug in the metadata or did some real change occur?

This issue hits us because OSS gets scanned in our organization periodically, and every time nornir rings a bell and leads to manual checks.
If the mentioning of a proprietary license is just a bug, we would like to see it fixed.

```
> ls -lad  pypi/src/nornir-3.?.?
drwxr-x--- 3 ya17378 yao20 4096 Nov 11 17:09 pypi/src/nornir-3.1.1
drwxr-x--- 3 ya17378 yao20 4096 Nov 11 17:09 pypi/src/nornir-3.2.0
drwxr-x--- 3 ya17378 yao20 4096 Nov 11 17:09 pypi/src/nornir-3.3.0
drwxr-x--- 3 ya17378 yao20 4096 Nov 11 17:09 pypi/src/nornir-3.4.0
drwxr-x--- 3 ya17378 yao20 4096 Nov 11 17:09 pypi/src/nornir-3.4.1
> grep Propriet -r pypi/src/nornir-3.?.?
pypi/src/nornir-3.2.0/PKG-INFO:Classifier: License :: Other/Proprietary License
pypi/src/nornir-3.3.0/PKG-INFO:Classifier: License :: Other/Proprietary License
pypi/src/nornir-3.4.0/PKG-INFO:Classifier: License :: Other/Proprietary License
pypi/src/nornir-3.4.1/PKG-INFO:Classifier: License :: Other/Proprietary License
> ls -lad  pypi/whl/nornir-3.?.?
drwxr-x--- 4 ya17378 yao20 4096 Nov 11 17:37 pypi/whl/nornir-3.1.1
drwxr-x--- 4 ya17378 yao20 4096 Nov 11 17:38 pypi/whl/nornir-3.2.0
drwxr-x--- 4 ya17378 yao20 4096 Nov 11 17:38 pypi/whl/nornir-3.3.0
drwxr-x--- 4 ya17378 yao20 4096 Nov 11 17:39 pypi/whl/nornir-3.4.0
[...]

PR #976, opened 2024-11-11
Update Apache-2.0 in pyproject.toml

dgarros:
I think this should fix the issue in Pypi reported by #975 

https://python-poetry.org/docs/pyproject/#license


ktbyers:
@dgarros Can you also switch the classifier to:

License :: OSI Approved :: Apache Software License 2.0 (Apache-2.0)

In pyproject.toml

See:

https://github.com/pypa/trove-classifiers/issues/17

dgarros:
> @dgarros Can you also switch the classifier to:
> 
> License :: OSI Approved :: Apache Software License 2.0 (Apache-2.0)

Done

ktbyers (approved):

ktbyers:
FYI, the second part of this didn't work i.e. pypi rejected it:

```
WARNING  Error during upload. Retry with the --verbose option for more details.                                                     
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/                                                            
         'License :: OSI Approved :: Apache Software License 2.0 (Apache-2.0)' is not a valid classifier. See                       
         https://packaging.python.org/specifications/core-metadata for more information.
```

Also the classifier never accepted the change that was in the referenced issue (see):

https://github.com/pypa/trove-classifiers/blob/main/src/trove_classifiers/__init__.py#L283

So I am going to revert this part of the PR.

ktbyers:
I checked both the METADATA file and the PKG-INFO file and the Properietary License referfence was gone (after reverting the classifier part of this.

So adding the hyphen was sufficient to fix the original issue.

ktbyers:
@dbarrosop @ogenstad Should be a quick review/approval.

PR #979, opened 2024-12-16
Remove PY3.8 support; Add PY3.13 support

ktbyers:
Update GH actions to reflect this.

Switch to `types-setuptools` since pypi removed the previously used types-pkg-resources.


ktbyers:
Closer, but now breaking at a nbval test (on PY3.13).

ktbyers:
Note, for some reason the docs/tutorial/processors.ipynb test was failing on PY3.13 due to Hosts being in a different order. I assumed this in some way pertained to a threading change i.e. that insertion order was changing.

I just added a `sort_keys=True` on the json.dumps() call to ensure that dictionary order of the hosts was the same (on the output).

ktbyers:
@dbarrosop @ogenstad I think this should be good to go, if one of you want to review/approve.

Note, I will plan on doing a release once this is approved.

Note also, I pinned to a test version of PyEZ for Nornir testing purposes. This test version of PyEZ fixes the underlying PY3.13 breaking issue. This is only in Nornir dev dependencies.

This does imply that anyone using Nornir-napalm and PY3.13 will still run into breakage due to PYEZ, but I don't think we should block all other Nornir users while waiting for PyEZ to fix their underlying issue.

nornir/core/configuration.py
```diff
@@ -16,7 +16,7 @@
 T = TypeVar("T")
 
 
-class Parameter:
+class Parameter(Generic[T]):
```
ktbyers, line 19: This looked like the recommended way to enforce type consistency i.e. that a Parameter that was a 'str' was consistently required to be a 'str').

There were some mypy issues until I did this.

nornir/core/configuration.py
```diff
@@ -47,14 +47,18 @@ def resolve(self, value: Optional[T]) -> T:
...
+
         return v
 
 
 class SSHConfig:
     __slots__ = ("config_file",)
 
     class Parameters:
-        config_file = Parameter(default=DEFAULT_SSH_CONFIG, envvar="NORNIR_SSH_CONFIG_FILE")
+        config_file = Parameter[str](default=DEFAULT_SSH_CONFIG, envvar="NORNIR_SSH_CONFIG_FILE")
```
ktbyers, line 61: This was a secondary requirement of using `Parameter(Generic[T])`

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["poetry==1.3.2"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
```
ktbyers, line 3: New Poetry build format.

pyproject.toml
```diff
@@ -41,10 +41,15 @@ pytest-cov = "*"
 requests-mock = "*"
 mypy = "^1.5.1"
 types-Jinja2 = "^2.11.9"
-types-pkg-resources = "^0.1.3"
+types-setuptools= ">=75.5.0.20241116"
```
ktbyers, line 44: `types-pkg-resources` was completely removed from pypi--all versions. Was fairly annoyed by that.

dbarrosop (approved):

dbarrosop:
Great job, thanks!

> This does imply that anyone using Nornir-napalm and PY3.13 will still run into breakage due to PYEZ, but I don't think we should block all other Nornir users while waiting for PyEZ to fix their underlying issue.

Agreed, that shouldn't block nornir core.

PR #982, opened 2025-01-05
Fix license classifier

ktbyers:
This change is needed for the Nornir 3.5.0 release (as pypi is rejected the classifier that we specified in #976)


ktbyers:
This PR reverts the classifier part of this PR (as it was incorrect):

https://github.com/nornir-automation/nornir/pull/976

ktbyers:
Additional notes:

Pypi rejected uploaded of Nornir 3.5.0 with:

```
WARNING  Error during upload. Retry with the --verbose option for more details.                                                     
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/                                                            
         'License :: OSI Approved :: Apache Software License 2.0 (Apache-2.0)' is not a valid classifier. See                       
         https://packaging.python.org/specifications/core-metadata for more information.
```

Also the classifier never accepted the change that was in the referenced issue (see):

https://github.com/pypa/trove-classifiers/blob/main/src/trove_classifiers/__init__.py#L283

ktbyers:
I checked both the METADATA file and the PKG-INFO file and the Properietary License referfence was gone (after reverting the classifier part of this).

So the original issue #975 should still be fixed.

ogenstad:
So we think that the initial issue might have been resolved by this change:

```diff
-license = "Apache 2.0"
+license = "Apache-2.0"
```

But that other metadata change caused some other violation?

dbarrosop (approved):

ktbyers:
@ogenstad Yes, the hyphen fixed the original METADATA issue (I verified this).

The classifier change was wrong and broke the pypi upload.

Issue #994, opened 2025-03-26
Tweak to MultiResult to allow pickling for multiprocessing runner


amylieb:
Hello!

I've been using Nornir to pull data from my network and have been experimenting with ways to improve performance.

I've found that my code is more CPU bound than I/O bound because it does a bunch of normalization (it's basically a bunch of custom napalm getters that feed a postgres database).

It was trivial to create my own Multiprocessing runner based on the `ThreadedRunner`, but to get it to work I had to tweak the `MultiResult` class because `__getattr__` is overridden in that class and it breaks the pickling.

Will open a PR with this tweak so you can see what I mean - it's not a particularly elegant solution, but is it something you'd guys consider adding or fixing in some other way? I do get that adding support for Multiprocessing in this way it's own potential can of worms.

dbarrosop:
Yeah, don't have strong opinions about this. Very first iteration of Nonir, back when it was brigade, actually used processes instead of threads but performance dropped when scaling io intensive tasks so I ended up changing it. I am sure there is some discussion somewhere about it as I recall discussing this (maybe hidden behind slack's paywall)

PR #995, opened 2025-03-26
fix to allow pickling Result objects

amylieb:
Ref https://github.com/nornir-automation/nornir/issues/994

### DIFF
--- a/nornir/core/task.py
+++ b/nornir/core/task.py
@@ -254,7 +254,7 @@ class MultiResult(List[Result]):
         # without this pickling breaks
         if name in ["__getstate__", "__setstate__"]:
             return super().__getattr__(name)
-        
+
         return getattr(self[0], name)
 
     def __repr__(self) -> str:
--- a/nornir/core/task.py
+++ b/nornir/core/task.py
@@ -253,7 +253,7 @@ class MultiResult(List[Result]):
 
         # without this pickling breaks
         if name in ["__getstate__", "__setstate__"]:
-            return super().__getattr__(name)
+            return super().__getattribute__(name)
 
         return getattr(self[0], name)
 
--- a/nornir/core/task.py
+++ b/nornir/core/task.py
@@ -251,7 +251,7 @@ class MultiResult(List[Result]):
 
     def __getattr__(self, name: str) -> Any:
 
-        # tweak for multiproccessor runner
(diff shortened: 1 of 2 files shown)


nornir/core/task.py
```diff
@@ -250,6 +250,11 @@ def __init__(self, name: str) -> None:
         self.name = name
 
     def __getattr__(self, name: str) -> Any:
+
+        # tweak for multiproccessor runner
```
dbarrosop, line 254:
```suggestion
        # without this pickling breaks
```

dbarrosop:
I'd suggest adding a test pickling and object that breaks without this snippet in case someone sees this code and thinks "why the hell?" and removes it

amylieb:
> I'd suggest adding a test pickling and object that breaks without this snippet in case someone sees this code and thinks "why the hell?" and removes it

Sure - added a basic test for this. It felt silly to use an assertion statement when a failure will naturally raise an exception, but I can add one if that's best practice (Sorry - am relatively new to contributing, so I'm unsure as to what's normal).

dbarrosop:
I don't think an assert is needed as you pointed out but not propagating the error is probably nicer (see comment).

amylieb:
> I don't think an assert is needed as you pointed out but not propagating the error is probably nicer (see comment).

Makes sense to me. Thanks for considering this update!

Side note - the other pickling issue I'm running into is when `Task.start` catches an exception that contains a non-pickleable object, which then gets saved and returned as a `Result`. Junos pyez is the specific culprit - I would never expect an exception to return a complex object like [they are doing](https://github.com/Juniper/py-junos-eznc/blob/master/lib/jnpr/junos/device.py#L1412)). It's easy to work around this with try/excepts within a task function, but if you guys ever consider adding a Multiprocessing runner to the core plugins you may want to cast exceptions to strings or something before saving them to a `Result`.

dbarrosop:
looks like the linter isn't happy

amylieb:
That mypy error is interesting. Technically it's right in that `List` does not have a `__getattr__` class (which I never would have thought to check). But the code works that way anyway? I've changed the call to `__getattribute__` and it seems to behave in the same way.

dbarrosop (approved):

dbarrosop:
Thanks!

Issue #869, opened 2023-09-22
Need to add tests into ci-cd to ensure wheel built from pyproject.toml file is importable


ktbyers:
Nornir Release 3.4.0 -- this breakage wasn't detected until after the fact.

dbarrosop:
what happened?

dbarrosop:
oh, got it. Saw the other PR with the explanation.

Reference #867

PR #1049, opened 2026-07-02
chore: migrate project tooling from Poetry to uv

### LINKED ISSUE #983: GH Actions Cache is broken
Tests work when cache is invalidated via updating the poetry.lock file. Breaks on second CI-CD run (when cache is used).

Switch to new form of cacheing (see Netmiko repository).

Should probably update Poetry version at the same time.

### LINKED ISSUE #1048: Migrate project tooling from Poetry to uv
## Need

The project currently uses Poetry for dependency management, packaging, and its
development workflow. We want to move to [uv](https://docs.astral.sh/uv/) as the
project's package/dependency manager.

This matters because uv offers substantially faster dependency resolution and
installs, a single tool for environment + dependency + build management, and has
become the de-facto standard tooling in the Python ecosystem. Faster, simpler
tooling lowers friction for contributors and speeds up CI.

The migration should be complete rather than partial — every place that assumes
Poetry today needs to move over, otherwise contributors hit a mix of two tools.

## Context

Poetry is currently referenced across several surfaces, all of which are in
scope for the switch:

- **Packaging / project config** — `pyproject.toml`, `poetry.lock`
- **Docker build** — `Dockerfile`
- **CI** — `.github/workflows/main.yaml`
- **Dev workflow** — `Makefile`
- **Docs build** — `.readthedocs.yml`
- **Documentation** — `README.md`, `CONTRIBUTING.rst`, `docs/plugins/index.rst`

The supported Python versions and the set of dependencies/extras should stay the
same through the migration; this is a tooling change, not a dependency bump.

## References

- uv documentation: https://docs.astral.sh/uv/
- Related tooling issue: #869 (wheel built from `pyproject.toml` is importable)

dgarros:
## Summary

Moves the project's package/dependency management from Poetry to uv, giving contributors one tool for environments, dependencies, and builds, and making dependency installs in CI and Docker substantially faster. The migration is complete — no surface still assumes Poetry — while supported Python versions (3.10–3.14) and all effective dependency constraints stay exactly the same.

Closes #1048
Closes #983 — the broken manual `.venv` cache is gone; setup-uv caches uv's wheel store keyed on `uv.lock` and every run builds a fresh venv.

## Key Changes

- Contributors now run `uv sync` / `uv run …` (or the unchanged `make` targets) instead of installing and configuring Poetry; a bare `uv sync` installs the same dev+docs groups a bare `poetry install` did.
- Packaging is now standards-based: PEP 621 `[project]` metadata, PEP 735 dependency groups, and `[project.entry-points]` for nornir's runner/inventory plugin discovery. The wheel built by `uv build` was verified to be importable in a clean environment with both entry-point groups intact (relates to #869).
- CI installs dependencies via `astral-sh/setup-uv` with built-in lockfile-keyed caching, replacing three setup steps per job; the Windows cached-venv hang workaround is no longer needed.
- Docker and ReadTheDocs builds use `uv sync --locked`, preserving the dependency layer cache in the Dockerfile.
- Locked versions are carried over from `poetry.lock` for packages whose newer releases broke checks (mypy, pytest, pytest-cov, pluggy, decorator, setuptools) — this PR is a tooling change, not a dependency bump.
- Ruff now correctly infers target Python 3.10 from `requires-python` (it previously defaulted to py38 because there was no `[project]` table); the newly-firing pyupgrade rules (UP006/UP007/UP035/UP036) were added to the existing reactivate-later ignore block instead of churning source files.
- Fixes `make tests` depending on a nonexistent `sphinx` target (now `docs`), which previously made `make docker-tests` fail immediately.

## Documentation Updates

- `README.md` and `CONTRIBUTING.rst` now describe the uv workflow.
- `docs/plugins/index.rst` recommends the standards-based `[project.entry-points]` form first and keeps the Poetry form as an alternative, since plugin authors' own projects may use either tool.

## Test Plan

- `make ruff`, `make mypy`, `make pytest` (117 passed), `make nbval` (117 passed), and `make docs` all pass locally via uv.
- `make docker && make docker-tests` runs the full suite green inside the rebuilt container.
- `uv build` wheel inspected (entry points, metadata, no tests packaged) and installed into a fresh venv: `InitNornir` imports and both plugin entry-point groups resolve.
- CI on this PR exercises the new workflow across the full 3.10–3.14 × Linux/macOS/Windows matrix; ReadTheDocs is the one surface only verifiable by its own build.


Dockerfile
```diff
@@ -1,29 +1,30 @@
...
     && apt-get install -yq curl git pandoc make \
-    && curl -sSL https://install.python-poetry.org  | python3 - \
-    && poetry config virtualenvs.create false
-
-COPY pyproject.toml .
-COPY poetry.lock .
-
-# Dependencies change more often, so we break RUN to cache the previous layer
-RUN poetry install --no-interaction
+    && rm -rf /var/lib/apt/lists/*
```
ogenstad, line 15: I'm not sure that the Dockerfile is really used now, but I think we can also get rid of `curl` when we don't need it to install poetry.

ogenstad (approved):
LGTM, small comment when building a package is that we get a warning:

```bash
❯ uv build
Building source distribution (uv build backend)...
warning: Found license classifier `License :: OSI Approved :: Apache Software License`. License classifiers are ambiguous and deprecated per PEP 639; projects should use `project.license` and `project.license-files` instead.
Building wheel from source distribution (uv build backend)...
Successfully built dist/nornir-3.5.0.tar.gz
Successfully built dist/nornir-3.5.0-py3-none-any.whl
```

Also with the changes to the CI we can also close #983.

dgarros:
Thanks for the review @ogenstad — all three points addressed in 80aa2c6:

- Dropped the deprecated `License :: OSI Approved` classifier; `uv build` is warning-free now. Also declared `license-files = ["LICENSE"]` while in there — Poetry used to bundle LICENSE into the artifacts implicitly, and `uv_build` needs it declared, so the wheel/sdist ship the license file again.
- Removed `curl` from the Dockerfile (verified nothing else in the image needs it; `make docker` and the in-container test suite still pass).
- Added `Closes #983` to the description — the manual `.venv` cache that caused those second-run failures is gone entirely with this migration.

PR #1055, opened 2026-07-03
fix(simple_inventory): handle empty inventory files gracefully

### LINKED ISSUE #1054: Empty hosts.yaml gives cryptic AttributeError instead of clear error message
## Describe the bug

When `hosts.yaml` is empty, Nornir raises a cryptic `AttributeError: 'NoneType' object has no attribute 'items'` instead of a clear error message explaining the real problem.

The root cause is that `yaml.safe_load()` returns `None` for an empty YAML file. This `None` value is then passed downstream where `.items()` is called on it, which fails.

A new user has no way to know what `NoneType` refers to here without digging into Nornir's source code.

## To reproduce

1. Create an empty `hosts.yaml` file:

```
touch inventory/hosts.yaml
```

2. Run `InitNornir(config_file=config.yaml)`

3. See the error:

```
AttributeError: 'NoneType' object has no attribute 'items'
```

## Expected behavior

A clear error message such as:

```
ValueError: hosts.yaml is empty. Please define at least one host or use an empty mapping: {}
```

or simply handling the `None` case gracefully:

```
hosts_dict = yaml.safe_load(f) or {}
```

which would treat an empty file as an empty inventory.

The same issue applies to `groups.yaml` as well.

## Environment

- Nornir version: latest (tested with current release)
- Python version: 3.12
- OS: Linux

DharmaBytesX:
Closes #1054

When hosts.yaml (or groups.yaml, defaults.yaml) is empty, ruamel.yaml.YAML.load() returns None. This caused an AttributeError when iterating over the result with .items().

This fix ensures that None is coalesced to an empty dict, consistent with how the other inventory files were already handled. A warning is logged when this happens so the user is aware.

Changes:
- Default hosts_dict to {} if yaml.load() returns None
- Add warning logs for empty inventory files
- Add test for completely empty hosts.yaml

All existing tests pass.

### DIFF
--- a/nornir/plugins/inventory/simple.py
+++ b/nornir/plugins/inventory/simple.py
@@ -97,8 +97,9 @@ class SimpleInventory:
                 defaults_dict = yml.load(f) or {}
             if defaults_dict == {}:
                 logger.warning(
-                    f"'{self.defaults_file}' is empty or contains only comments, "
-                    "defaulting to empty defaults."
+                    "'%s' is empty or contains only comments, "
+                    "defaulting to empty defaults.",
+                    self.defaults_file,
                 )
             defaults = _get_defaults(defaults_dict)
         else:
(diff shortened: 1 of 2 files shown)


DharmaBytesX:
Je viens de fix, j'ai remplacé les f-strings. Le linter devrait fonctionner maintenant.

dgarros (approved):

PR #1057, opened 2026-07-03
fix: raise a readable error when Task.run() is called outside a nested task

### LINKED ISSUE #1051: Task.run() guard raises a garbled tuple as the exception message
## What happens

The guard at the top of `Task.run()` builds its error message with trailing commas inside the parentheses, so `msg` is a **tuple of two strings** rather than a single string ([nornir/core/task.py:150-155](https://github.com/nornir-automation/nornir/blob/main/nornir/core/task.py#L150-L155)):

```python
if not self.host:
    msg = (
        "You have to call this after setting host and nornir attributes. ",
        "You probably called this from outside a nested task",
    )
    raise Exception(msg)
```

When the guard fires, the user sees the tuple's repr:

```
Exception: ('You have to call this after setting host and nornir attributes. ', 'You probably called this from outside a nested task')
```

A related wrinkle in the same guard: `self.host` is only assigned in `Task.start()`, so the most likely misuse this message is written for — calling `task.run(...)` from outside a nested task — never actually reaches it. It fails earlier with `AttributeError: 'Task' object has no attribute 'host'`. The tuple message only surfaces when `host` exists but is falsy.

## What should happen

A single readable message:

```
Exception: You have to call this after setting host and nornir attributes. You probably called this from outside a nested task
```

— ideally raised for the "called from outside a nested task" case too, instead of the bare `AttributeError`.

## Steps to reproduce

```python
from nornir.core.processor import Processors
from nornir.core.task import Task

def sub(task):
    pass

t = Task(sub, nornir=None, global_dry_run=False, processors=Processors())
t.run(sub)        # AttributeError: 'Task' object has no attribute 'host'

t.host = None
t.run(sub)        # Exception: ('You have to call this after setting host...', 'You probably...')
```

## Context

This is also the only bare `raise Exception(...)` in this module — ruff's TRY002 ("create your own exception") and TRY003 are currently suppressed in `pyproject.toml`. Fixing this spot is a natural opportunity to use a dedicated exception class from `nornir.core.exceptions` instead.

dgarros:
## Summary
When `Task.run()` is misused — called before the task has been assigned a host, i.e. from outside a running nested task — users now get a single readable error message pointing at the actual mistake, instead of a garbled tuple repr or an unrelated `AttributeError`.

## Key Changes
- Users misusing `task.run(...)` now see `You have to call this after setting host and nornir attributes. You probably called this from outside a nested task` — previously the guard raised the repr of a two-string tuple (trailing commas made `msg` a tuple).
- The most common misuse path (no `host` attribute set at all, since `Task.start()` was never called) now reaches the guard and gets the same readable error, instead of leaking `AttributeError: 'Task' object has no attribute 'host'`.
- The error is a dedicated `NornirTaskNotStartedError` in `nornir.core.exceptions` (replacing the module's only bare `raise Exception(...)`), picklable so runner plugins that serialize `Result.exception` keep working. Backward compatible: it subclasses `Exception`, so existing handlers still catch it.
- Public typing of `Task.host` is deliberately unchanged (`Host`, not `Optional[Host]`), so downstream plugin code doing `task.host.get_connection(...)` stays mypy-clean.

## Related Context
Fixes #1051.

## Test Plan
- New regression test `tests/core/test_tasks.py::Test::test_run_outside_nested_task` covers both reproduction paths from the issue (missing `host` attribute, and `host = None`) and asserts the exact message.
- `uv run pytest` (118 passed), `uv run mypy nornir tests`, `uv run ruff check` / `ruff format --check` all clean.


nornir/core/task.py
```diff
@@ -147,12 +151,8 @@ def grouped_tasks(task):
 
         This method will ensure the subtask is run only for the host in the current thread.
         """
-        if not self.host:
-            msg = (
-                "You have to call this after setting host and nornir attributes. ",
-                "You probably called this from outside a nested task",
-            )
-            raise Exception(msg)
+        if not getattr(self, "host", None):
```
ogenstad, line 154: This feels a bit off to me. I think the real problem is that self.host is only defined on line 90/94 and we don't have any default assignment in `__init__()` looks like ruff doesn't have support for this pylint rule:

https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/attribute-defined-outside-init.html

I think a cleaner approach would be to set self.host to a `None` value in __init__(), or possibly set self._host and have a property for self.host that always returns a host object.

PR #1061, opened 2026-07-12
refactor: modernize type annotations to PEP 585/604 syntax (#1058)

### LINKED ISSUE #1058: refactor: modernize type annotations to PEP 585/604 syntax
## Need

The codebase uses legacy typing style throughout: `Dict`/`List`/`Optional`/`Union` from `typing` (79 `Optional[...]` uses alone), `KeysView`/`ValuesView` imported from `typing` rather than `collections.abc`, and quoted string forward-references like `"Nornir"`. No module uses `from __future__ import annotations`.

This is inconsistent with the minimum supported Python (3.9+) and makes signatures noisier than they need to be. The project has already committed to this cleanup: `pyproject.toml` suppresses `UP006`, `UP007`, `UP035`, and `UP045` under a comment stating they "should be removed once the code has been updated ... so that we can reactivate them one by one."

## Context

- The relevant suppressions live in the `[tool.ruff.lint] ignore` list in `pyproject.toml` (UP006, UP007, UP035, UP045), flagged as temporary.
- The bulk is mechanically addressable via ruff's autofix; adding `from __future__ import annotations` is what lets the quoted forward-refs be unquoted.
- Scope is style-only — no behavioural change intended.

## References

- Related: #985 (minimum supported Python version)

dgarros:
## Summary
Modernizes the codebase's type annotations to PEP 585/604 syntax and
reactivates the ruff rules that enforce it. This is a style-only cleanup
with no behavioural change — it removes the temporary UP006/UP007/UP035/UP045
suppressions the project had committed to lifting "one by one".

Closes #1058.

## Key Changes
- Reactivated ruff `UP006`, `UP007`, `UP035`, `UP045` by removing them from
  the temporary `[tool.ruff.lint] ignore` list.
- `Dict`/`List`/`Type`/`Set` → `dict`/`list`/`type`/`set` (PEP 585);
  `Optional[X]`/`Union[X, Y]` → `X | None` / `X | Y` (PEP 604).
- `KeysView`/`ValuesView`/`ItemsView`/`Iterator` now import from
  `collections.abc` instead of `typing`.
- Added `from __future__ import annotations` across the package so quoted
  forward-references (including `TYPE_CHECKING`-only ones like `"Nornir"`)
  are unquoted.
- Relocated type-only standard-library imports into `TYPE_CHECKING` blocks
  (ruff `TC003`), a direct side-effect of the lazy annotations.

## Scope boundaries
Two rules that only surface as by-products of the above are suppressed with
explanatory comments, as reactivating them would exceed a style-only change:
- `FURB189` — several public classes deliberately subclass `dict`/`list`
  (`Hosts`, `Groups`, `Processors`, `MultiResult`, `AggregatedResult`);
  switching to `UserDict`/`UserList` would change behaviour.
- `TC001` — moving first-party imports into `TYPE_CHECKING` blocks is an
  import-restructuring refactor with public re-export risk, left for a
  follow-up.

## Test Plan
- `make ruff` — passes (with the four rules re-enabled)
- `make mypy` — no issues
- `make pytest` — 118 passed


ogenstad (approved):
LGTM

PR #1072, opened 2026-08-16
ci: test wheel importability

### LINKED ISSUE #869: Need to add tests into ci-cd to ensure wheel built from pyproject.toml file is importable

TheCodeOfDave:
## Summary

- build the project wheel in CI with `uv build`
- install the wheel into a fresh virtual environment
- import `nornir` and `InitNornir` outside the source checkout
- verify that the imported module comes from the isolated environment

This adds regression coverage for packaging problems that can remain hidden when tests run with development dependencies or import directly from the repository.

## Verification

- GitHub Actions passed on the fork: 17/17 jobs
- the new `Build and import wheel` job passed on Ubuntu with Python 3.14
- a local negative control with an undeclared runtime import failed the same test as expected
- `uv run pytest -q`: 118 passed
- `uv run ruff check .`: passed
- `uv run ruff format --check .`: passed
- `uv run mypy nornir tests`: passed
- actionlint: passed

Fixes #869

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -41,28 +41,8 @@ jobs:
           python-version: "3.14"
           enable-cache: true
 
-      - name: Build wheel
-        run: uv build
-
-      - name: Install and import wheel
-        run: |
-          uv venv /tmp/nornir-wheel-test
-          uv pip install --python /tmp/nornir-wheel-test/bin/python dist/*.whl
-          cd /tmp
-          /tmp/nornir-wheel-test/bin/python -I - <<'PY'
-          import pathlib
-          import sys
-
-          import nornir
-          from nornir import InitNornir
-
-          module_path = pathlib.Path(nornir.__file__).resolve()
-          environment = pathlib.Path(sys.prefix).resolve()
-          if environment not in module_path.parents:
-              raise SystemExit(f"Imported nornir from outside {environment}: {module_path}")
-          if InitNornir.__name__ != "InitNornir":
-              raise SystemExit("InitNornir is not importable from the built wheel")
-          PY
+      - name: Build and import wheel
+        run: make wheel
 
   pytest:
     name: Testing on Python ${{ matrix.python-version }} (${{ matrix.platform}})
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -30,6 +30,40 @@ jobs:
...
+        uses: astral-sh/setup-uv@v8.2.0
+        with:
+          python-version: "3.14"
+          enable-cache: true
+
+      - name: Build wheel
+        run: uv build
+
+      - name: Install and import wheel
+        run: |
```
ogenstad: Thanks for the contribution! One change though is that these inline code snippets are hard to test outside of the CI run. It would be better with a Python script that lived in the repo and we instead added an entry to the Makefile, then ran that makefile entry here.

ogenstad (changes requested):

.github/workflows/main.yaml
TheCodeOfDave: Thanks you are right that way is much better.  I moved the isolated wheel verification into tests/wheel_importability.py, added a make wheel target, and changed the workflow to call that target. The check now runs locally on Windows and through the Make target on Linux/Python 3.14. I also added focused tests for wheel discovery and POSIX/Windows virtual-environment layouts. Please let me know if that is what you had in  mind.

Makefile
```diff
@@ -8,6 +8,11 @@ PYTHON:=3.10
 docker:
 	docker build --build-arg PYTHON=$(PYTHON) -t $(NAME):latest -f Dockerfile .
 
+.PHONY: wheel
+wheel:
```
ogenstad, line 12: Can you add a `rm -rf dist` here so that it would first remove any existing wheels if you ran this locally?

ogenstad (commented):
LGTM, just added a small comment about one extra thing.

Makefile
TheCodeOfDave: Added `rm -rf dist` before `uv build` in the `wheel` target in a93cbfa. I verified it by planting a stale wheel and running `make wheel` under Linux/Python 3.14: the stale wheel was removed, the isolated import passed, and exactly one wheel remained. The full pytest suite still passes (124 tests).

ogenstad (approved):
Thank you for the updates!

PR #1083, opened 2026-08-27
Expose InitNornir configuration parameters explicitly

### LINKED ISSUE #1079: InitNornir hides supported keyword argument with **kwargs
Currently the InitNornir function (https://github.com/nornir-automation/nornir/blob/v3.6.0/nornir/init_nornir.py#L43-L77) takes a **kwarg and passes it as is:

```python
def InitNornir(
    config_file: str = "",
    dry_run: bool = False,
    **kwargs: Any,
) -> Nornir:
    """
    Arguments:
        config_file(str): Path to the configuration file (optional)
        dry_run(bool): Whether to simulate changes or not
        configure_logging: Whether to configure logging or not. This argument is being
            deprecated. Please use logging.enabled parameter in the configuration
            instead.
        **kwargs: Extra information to pass to the
            :obj:`nornir.core.configuration.Config` object

    Returns:
        :obj:`nornir.core.Nornir`: fully instantiated and configured
    """
    ConnectionPluginRegister.auto_register()

    if config_file:
        config = Config.from_file(config_file, **kwargs)
    else:
        config = Config.from_dict(**kwargs)

    data = GlobalState(dry_run=dry_run)

    config.logging.configure()

    return Nornir(
        inventory=load_inventory(config),
        runner=load_runner(config),
        config=config,
        data=data,
    )
```

These kwargs then land in the Config class (https://github.com/nornir-automation/nornir/blob/v3.6.0/nornir/core/configuration.py#L260-L289)

```python
class Config:

    @classmethod
    def from_dict(
        cls,
        inventory: dict[str, Any] | None = None,
        ssh: dict[str, Any] | None = None,
        logging: dict[str, Any] | None = None,
        core: dict[str, Any] | None = None,
        runner: dict[str, Any] | None = None,
        user_defined: dict[str, Any] | None = None,
    ) -> Config:
        return cls(
            inventory=InventoryConfig(**inventory or {}),
            ssh=SSHConfig(**ssh or {}),
            logging=LoggingConfig(**logging or {}),
            core=CoreConfig(**core or {}),
            runner=RunnerConfig(**runner or {}),
            user_defined=user_defined or {},
        )

    @classmethod
    def from_file(
        cls,
        config_file: str,
        inventory: dict[str, Any] | None = None,
        ssh: dict[str, Any] | None = None,
        logging: dict[str, Any] | None = None,
        core: dict[str, Any] | None = None,
        runner: dict[str, Any] | None = None,
        user_defined: dict[str, Any] | None = None,
    ) -> Config:
```

The use of `**kwargs` makes it seem like the parameters could be dynamic or that we don't know which parameters might be available. However, looking at the next step we do have this information already. I.e. we know that the only supported parameters would be:
* inventory
* ssh
* logging
* core
* runner
* user_defined

As such we could replace the `**kwargs` in this place as it could catch typos earlier and allow for a linter to warn end users if they used incorrect arguments.

leojuni:
Closes #1079

Replaces `InitNornir()`'s `**kwargs` with the six configuration sections already accepted by `Config`. This makes the supported arguments visible to users, IDEs and type checkers while preserving existing keyword-based calls.

I also updated the docstring and added tests for the public signature and configuration forwarding, both with and without a config file.

Tests: `make tests`

### DIFF
--- a/nornir/init_nornir.py
+++ b/nornir/init_nornir.py
@@ -43,18 +43,25 @@ def load_runner(
 def InitNornir(
     config_file: str = "",
     dry_run: bool = False,
-    **kwargs: Any,
+    *,
+    inventory: dict[str, Any] | None = None,
+    ssh: dict[str, Any] | None = None,
+    logging: dict[str, Any] | None = None,
+    core: dict[str, Any] | None = None,
+    runner: dict[str, Any] | None = None,
+    user_defined: dict[str, Any] | None = None,
 ) -> Nornir:
     """Instantiate and configure a Nornir object.
 
     Arguments:
         config_file(str): Path to the configuration file (optional)
         dry_run(bool): Whether to simulate changes or not
-        configure_logging: Whether to configure logging or not. This argument is being
-            deprecated. Please use logging.enabled parameter in the configuration
-            instead.
-        **kwargs: Extra information to pass to the
-            :obj:`nornir.core.configuration.Config` object
+        inventory(dict): Inventory configuration section
+        ssh(dict): SSH configuration section
+        logging(dict): Logging configuration section
+        core(dict): Core configuration section
+        runner(dict): Runner configuration section
+        user_defined(dict): User-defined configuration section
 
     Returns:
         :obj:`nornir.core.Nornir`: fully instantiated and configured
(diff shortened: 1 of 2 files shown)


ogenstad (approved):
Looks good to me, thanks!

PR #1094, opened 2026-09-02
chore: set up the repository for AI coding agents

### LINKED ISSUE #1084: Setting up the repository for AI coding agents
> This proposal is part of the Nornir mid-term roadmap. The overview, the themes and the timeline are in #1091.

## Why

People are already using AI coding agents on this repository. The problem is not capability, it is consistency: an agent starts every session with no knowledge of the project, and it guesses at everything that is not written down. Most of what an agent needs to know here is real but unwritten:

- Notebook output under `docs/` is verified in CI, so it must be re-executed, not edited by hand.
- The ruff ignore list in `pyproject.toml` is a debt list to shrink, not a permission to add more entries.
- The plugin contracts are `typing.Protocol`, so changing a signature breaks every third-party plugin at once, with no inheritance error to catch it.
- `docs/api/` is generated by `docs/build_api.sh` and committed, so adding a module changes tracked files.
- Windows is a first-class CI target, not best-effort.

An agent that does not know these things produces work a reviewer has to catch. Writing them down once moves that burden from every review to a single file. This issue documents the change. It has three parts.

## 1. Repository structure

Agent-facing content lives in one place, and each tool gets a thin view of it:

```
AGENTS.md          portable instructions, the source of truth for agents
CLAUDE.md          thin router: imports AGENTS.md, holds Claude-only overrides
.agents/           vendor-neutral content, authored once
  skills/          packaged procedures agents can invoke
  commands/        thin entrypoints
  rules/           behavioural rules
.claude/           adapter: skills/commands/rules are views of .agents/
.specify/          spec-kit workflow engine
```

If each tool kept its own copy of the same instructions, the copies would drift apart, and the drift would stay invisible until an agent acts on a stale copy. Everything is authored once under `.agents/`, and each tool points at it, so adding or removing a tool is a wiring change, not a rewrite.

`.agents/` is not a Nornir invention. It is an emerging cross-tool convention, and several agents already read `.agents/skills` natively. `.claude/` is an adapter, not a second home: only files truly specific to Claude live there (settings, hooks, subagents), and the shared directories are views of `.agents/`, so there is no copy that can fall out of sync.

## 2. Spec-kit

Spec-kit is a workflow engine for planning work before writing it. It installs as `.specify/` and drives four steps: specify (what and why), plan (how), tasks (an ordered breakdown), and implement. Each step produces a markdown file under `specs/`. It is plain files and prompts: no service, no runtime dependency, and nothing runs unless someone invokes it.

It also carries `memory/constitution.md`, the project's standing principles that each planning step is checked against: Python-first / no DSL, thin core with plugins out of tree, strict typing, the full support matrix, executable docs, and public API stability. These come from `CONTRIBUTING.rst` and existing practice; they are not new.

The value is mostly for larger changes, where the useful conversation is about the plan and not the diff. Small fixes do not need it.

## 3. `AGENTS.md`

`AGENTS.md` is the file an agent reads first. It is deliberately short and points to deeper material instead of duplicating it: what the core owns and what belongs in a plugin, where things live, which make target is the authoritative quality gate, and the conventions listed at the top of this issue.

`CLAUDE.md` exists only because Claude Code does not read `AGENTS.md` natively yet. It is a single line that imports `AGENTS.md`, plus any Claude-only overrides.

This is a context-building exercise, not a policy change. Almost nothing in it is new: it is the knowledge that lives in maintainers' heads and gets re-explained in every review, written down in a form both people and agents can read. New human contributors get the same information for free.

## What this does not change

- No changes to runtime code, dependencies, or published artifacts. The wheel and the sdist still contain only `nornir/` plus metadata.
- Nobody has to use an agent, or spec-kit, to contribute.
- No new CI checks, and no PR is gated on any of this.

For contributors who do not use AI tooling, the practical effect is a couple of extra markdown files at the root of the repository.

dgarros:
## Summary

AI coding agents are already being used on this repository, and every session starts from zero: the agent guesses at everything that is real but unwritten — that notebook output is CI-verified, that `docs/api/` is generated and committed, that the ruff ignore list is a debt ledger, that the `typing.Protocol` plugin contracts make any signature change silently breaking. This PR writes that knowledge down once, in a portable layout, so agents and new human contributors start from the project's actual rules instead of guesses.

## Key Changes

- Agents get a single entry point: `AGENTS.md` records the project's unwritten conventions (executable notebooks, generated API docs, the ruff debt ledger, Protocol-based plugin contracts, Windows as a first-class target) and points to deeper material. `CLAUDE.md` is a thin router that imports it.
- Agent-facing content (skills, commands, rules) is authored once under `.agents/`; `.claude/` is an adapter whose shared directories are symlink views, so no per-tool copy can drift.
- Spec-kit installs as `.specify/`, providing a specify → plan → tasks → implement workflow for larger changes — plain files and prompts, no service, no runtime dependency, nothing runs unless invoked.
- The project constitution is ratified at v1.0.0 (`.specify/memory/constitution.md`): six principles drawn from `CONTRIBUTING.rst` and existing practice, which spec-kit planning steps are checked against.

Nothing in the runtime package changes — the wheel and sdist still contain only `nornir/` plus metadata. No new CI checks, and nobody has to use an agent or spec-kit to contribute.

## Deliberate policy change (beyond #1084)

This PR intentionally carries one change that is more than documentation: the dependency guidelines in `CONTRIBUTING.rst` no longer require a dedicated dependency-only PR. Recorded maintainer approval on the PR that makes the change replaces it, and a feature or fix PR may carry the dependency change its own code needs (limited to what the code requires, with the `uv.lock` update, called out for review). Bulk refreshes still land on their own. This inverts the previous rule and is flagged here for explicit review rather than folded into the constitution ratification.

## Related Context

Closes #1084 (part of the mid-term roadmap, #1091).

## Test Plan

- No runtime code is touched; `make tests` is unaffected.
- Full CI is already green on this branch (run 33496889503, after rebase onto latest `main`).


.agents/skills/.gitkeep
ogenstad, line 1: I don't think we need this file.

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+
+Every behavioural change MUST arrive with tests. A bug fix MUST include a test that fails before
+the fix and passes after it.
+
+`make pytest` MUST pass on every supported Python version (3.10 through 3.14) and every supported
+platform (Linux, macOS, Windows). Windows is a first-class target, not best-effort: code MUST NOT
+assume POSIX path separators, unlimited path lengths, or shell semantics.
+
+Work that genuinely cannot be exercised on a platform MUST state that explicitly in the pull
+request rather than being silently skipped.
```
ogenstad: I think we might even want such work to be relegated to some other repo where it would live as a posix only plugin or similar.

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+`make tests` is the authoritative gate. It runs `ruff`, `mypy`, `nbval`, `pytest`, and `docs`, and
+all five MUST pass before a pull request merges. `make docker-tests` reproduces the CI Linux
+environment locally for changes that behave differently outside it.
+
+Dependency policy:
+
+- Runtime dependencies pin to the major version when the upstream project follows semver, and to an
+  exact version when it does not.
+- Development dependencies are unpinned (`*`), except `ruff`, which is pinned exactly so that
+  formatting and lint results are reproducible across contributors and CI.
```
ogenstad: I don't think there's a problem with pinned dev dependencies when it's for first level dev dependencies. Also nothing downstream breaks if we pin the dev dependencies in pyproject.toml and I think that it can be handy to have a quick reference for it in pyproject.toml. However dependencies that Nornir (or any of the Nornir plugins depend on that we ship should be as loose as possible to allow for broad installations.

.specify/memory/constitution.md
```diff
@@ -0,0 +1,218 @@
...
+- Adding, removing, or changing a dependency requires **maintainer approval**, recorded in the pull
+  request that makes the change. Approval is the gate — a dedicated issue or a separate
+  dependency-only pull request is NOT required, and development MUST NOT be blocked waiting for one.
+  A feature or fix PR MAY carry the dependency change its own code needs, provided the change is
+  limited to what that code actually requires and is called out explicitly for review.
```
ogenstad, line 145: I don't know if we want to separate dev and prod dependencies here, as I think we'd want a very good reason to add any additional non standard library dependencies to the core.

AGENTS.md
```diff
@@ -0,0 +1,147 @@
...
+- `.specify/memory/constitution.md` — the six ratified principles plus the quality-gate and
+  workflow sections. This is the authority; where any other document conflicts with it, it wins
+  and the other one is to be corrected.
+- `CONTRIBUTING.rst` — contributor-facing workflow (environment setup, dependency policy, tests).
+  Also rendered as the docs' contributing page.
+- `CHANGELOG.rst` — user-visible history; every user-visible change adds an entry here.
+
+One known drift, so you do not "fix" the wrong side of it: both the constitution and
+`CONTRIBUTING.rst` state that development dependencies are unpinned (`*`) with `ruff` alone pinned
+exactly, but `pyproject.toml` currently uses ranges (`ruff>=0.15.20,<0.16`, `mypy>=1.5.1,<2`) and
```
ogenstad: As I commented on for the constitution I don't think this is an issue. What I'd say here though is that I don't think we should mention any versions in the markdown files as they tend to get stale. Instead we should just use pyproject.toml or uv.lock as the reference for that.

ogenstad (approved):
Added some minor comments but looks good to me.

.specify/memory/constitution.md
dgarros: Good point, and it aligns well with Principle II. I updated Principle IV to say that platform specific functionality should not live in the core and should be a plugin out of tree instead. The requirement to state it explicitly in the PR now only covers the rare case where a core change really can't be exercised on a platform.

.specify/memory/constitution.md
dgarros: Makes sense, I updated the policy to match the pins instead of the opposite. The policy now says: runtime dependencies (`[project.dependencies]`) stay as loose as possible to allow broad installations, and development dependencies can be pinned or constrained to ranges since nothing downstream is affected. I also removed the special case for ruff. Constitution bumped to v1.1.0 and `CONTRIBUTING.rst` updated to match.

.specify/memory/constitution.md
dgarros: I reworked this section. The bullets do not present dev and prod as symmetric rules anymore. The runtime bullet is about installability and points to Principle II ("the core MUST NOT take a new runtime dependency to serve a single plugin's use case"), so the bar for adding a new core dependency is defined in one place and referenced from here.

AGENTS.md
dgarros: Agreed. Since your other comments resolve the policy question, this paragraph was not describing a drift anymore. I replaced it with a one line summary of the policy, without version numbers. I also added an explicit rule in the constitution that prose documents must not mention dependency versions, `pyproject.toml` and `uv.lock` are the reference.

